PR #1, opened 2021-04-08
feat: initial daemon implementation

andreabadesso:


src/utils.ts
```diff
@@ -0,0 +1,456 @@
...
+    yield {
+      type: 'reorg',
+      success: false,
+      message: 'Our best block was voided, we should reorg.',
+    };
+
+    return;
+  }
+
+  if (ourBestBlock.height > meta.height) {
```
pedroferreira1, line 370: I didn't understand when this situation may happen. The txId in the full node has a height that is bigger than the same txId in the wallet service?

src/utils.ts
```diff
@@ -0,0 +1,456 @@
...
+      success: true,
+      blockId: preparedBlock.tx_id,
+      height: preparedBlock.height,
+      transactions: uniqueTxs.map((tx: FullTx) => {
+        return tx.txId;
+      }),
+    };
+  }
+
+  yield {
```
pedroferreira1, line 452: If we have a transaction failure we still yield success: true here, right? Should we change that?

pedroferreira1 (changes requested):

src/index.ts
```diff
@@ -0,0 +1,47 @@
...
+    console.log('Transitioned to state: ', state.value);
+  }
+});
+
+const handleMessage = (message: any) => {
+  switch(message.type) {
+    case 'dashboard:metrics':
+      break;
+    case 'network:new_tx_accepted':
+      if (!message.is_block) return;
```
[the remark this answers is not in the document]
andreabadesso: Done in 211417e61068d7261413f2c06aa07da0b764e130

src/index.ts
```diff
@@ -0,0 +1,47 @@
...
+      if (message.is_voided) return;
+      if (message.type === 'network:new_tx_accepted') {
+        machine.send({ type: 'NEW_BLOCK' });
+      }
+      break;
+    case 'state_update':
+      if (message.state === Connection.CONNECTED) {
```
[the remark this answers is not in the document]
andreabadesso: Done in e8daa1c76746b742fc40f2225e94125c475c6cb8

src/machine.ts
```diff
@@ -0,0 +1,121 @@
+import {
+  Machine,
+  assign,
+  send,
+} from 'xstate';
+import { syncToLatestBlock } from './utils';
+import {
+  SyncSchema,
+  SyncContext,
+  // SyncEvent,
```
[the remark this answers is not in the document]
andreabadesso: Done in 3d3cbed54a2f6b41e486a78355fa2220ea108459

Also commented out the `SyncEvent` type as it is not being used yet

src/machine.ts
```diff
@@ -0,0 +1,121 @@
...
+  return () => {
+    console.log('Stopping the iterator.');
+    iterator.return('finished');
+
+    return;
+  };
+};
+
+// TODO: We need to type the Event
+export const SyncMachine = Machine<SyncContext, SyncSchema, any>({
```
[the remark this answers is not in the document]
andreabadesso: Done in 749d96829a4eeea9e33e8d368b0763d369d4f731

src/types.ts
```diff
@@ -0,0 +1,108 @@
...
+    reorg: {};
+  }
+}
+
+export interface SyncContext {
+  hasMoreBlocks: boolean;
+  error?: {};
+}
+
+export type SyncEvent =
```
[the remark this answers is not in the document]
andreabadesso: It is not.. Yet

We are currently using `any` for the events because I was having an issue with typescript types on them

I left a TODO to debug it further because it was taking too much time

Since the event is only locally generated, I don't see it as a big risk for now

Commented it out in 3d3cbed54a2f6b41e486a78355fa2220ea108459

src/utils.ts
```diff
@@ -0,0 +1,456 @@
...
+import axios from 'axios';
+import dotenv from 'dotenv';
+
+dotenv.config();
+
+AWS.config.update({
+  region: 'us-east-1',
+});
+
+// const FULLNODE_URL = process.env.FULLNODE_URL || 'ws://localhost:8080/v1a/ws/';
```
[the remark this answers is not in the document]
andreabadesso: Done in 5184025a5bff3c3e3f8cc6d184236c7adde0f913

src/utils.ts
```diff
@@ -0,0 +1,456 @@
...
+const DEFAULT_SERVER = process.env.DEFAULT_SERVER || 'https://node1.foxtrot.testnet.hathor.network/v1a/';
+const TOKEN_INDEX_MASK = 0b01111111;
```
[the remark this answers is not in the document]
andreabadesso: Done in 5184025a5bff3c3e3f8cc6d184236c7adde0f913

src/utils.ts
```diff
@@ -0,0 +1,456 @@
...
+
+let globalCache = {};
+
+export const lambdaCall = (fnName: string, payload: any): Promise<any> => new Promise((resolve, reject) => {
+  const lambda = new AWS.Lambda({
+    apiVersion: '2015-03-31',
+    endpoint: process.env.STAGE === 'local'
+      ? 'http://localhost:3002'
```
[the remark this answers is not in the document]
andreabadesso: Yes. It is the `wallet-service` running locally. Maybe I should change it to an env variable?

src/utils.ts
```diff
@@ -0,0 +1,456 @@
...
+      } catch(e) {
+        console.log('Erroed parsing response body: ', data.Payload);
+
+        return reject(e.message);
+      }
+    }
+  });
+});
+
+export const sendTx = async (tx): Promise<ApiResponse> => {
```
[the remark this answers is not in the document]
andreabadesso: Done in cc6b15212a652ae3854ad70ee14911aca1c785ef

src/utils.ts
andreabadesso: Done in cf3a62329aef8293923a9cb10497f2934dc347cc

src/utils.ts
```diff
@@ -0,0 +1,456 @@
...
+      return typedOutput;
+    }),
+    parents: responseBlock.parents,
+    height: responseBlock.height as number,
+  };
+
+  return block;
+};
+
+export const recursivelyDownloadTx = async (blockId: string, txIds: string[] = [], data: FullTx[] = []): Promise<FullTx[]> => {
```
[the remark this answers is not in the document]
andreabadesso: Added docstrings to the methods on `src/utils.ts` on f054877c39dd28c326ab96161e26b5a082eba8df

src/utils.ts
andreabadesso: We are downloading the best block from the full node and comparing it to our own. If a reorg happened, I believe there is a chance that the full node's best block height is lower than ours

Not sure if this is a real situation tho, this is one of the things I want to test using the full node simulator when it is available @msbrogli

src/utils.ts
```diff
@@ -0,0 +1,456 @@
...
+};
+
+export const downloadTx = async (txId: string, noCache: boolean = false) => {
+  if (!noCache && globalCache[txId]) {
+    return globalCache[txId];
+  }
+
+  const response = await axios.get(`${DEFAULT_SERVER}transaction?id=${txId}`);
+
+  if (Object.keys(globalCache).length > 200) {
```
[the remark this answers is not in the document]
andreabadesso: I clean the cache because transactions are big and they are being kept in memory

I updated the cache to use an LRU cache, added tests and set the `TX_CACHE_SIZE` as a env variable

Done in a168187ca0a72a9e4079f7968d5607fe712e1557

src/utils.ts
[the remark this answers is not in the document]
andreabadesso: Added docstrings to the methods in src/utils in f054877c39dd28c326ab96161e26b5a082eba8df

src/utils.ts
```diff
@@ -0,0 +1,456 @@
...
+
+export const getWalletServiceBestBlock = async (): Promise<Block> => {
+  const response = await lambdaCall('getLatestBlock', {});
+  const bestBlock: Block = response.block;
+
+  return bestBlock;
+};
+
+export const getFullNodeBestBlock = async (): Promise<Block> => {
+  const response = await axios.get(`${DEFAULT_SERVER}transaction?type=block&count=1`);
```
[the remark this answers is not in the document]
andreabadesso: I tried using this api on my local fullnode running on `block_by_height` branch and got the same error  described in https://github.com/HathorNetwork/sec-hathor-core/issues/57

Will wait for that issue to be resolved before refactoring it. What do you think?

src/machine.ts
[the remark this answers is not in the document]
pedroferreira1: Why did you comment the event type? Why haven't you deleted it?

src/utils.ts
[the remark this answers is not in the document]
pedroferreira1: I agree. Env variable or a configuration file.

src/utils.ts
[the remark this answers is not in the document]
pedroferreira1: I agree. At least leave a TODO FIXME to change the API, otherwise we will keep this forever and eventually this API will be deprecated

src/utils.ts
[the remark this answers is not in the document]
pedroferreira1: You still have no parameter type in `sentTx` method

src/utils.ts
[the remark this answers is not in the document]
andreabadesso: Done in 9354570ddbc21248a05a485ffff81f32d574180d

src/utils.ts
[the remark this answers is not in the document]
andreabadesso: Done in fd11312f9d35863e00a3e8a55f82c6ea12bab755

src/utils.ts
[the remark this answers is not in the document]
andreabadesso: You are 100% correct, fixed in 8bc87903e1ae1b2146423af56c3cba7dbabf21ce

src/utils.ts
```diff
@@ -0,0 +1,456 @@
...
+    return data;
+  }
+
+  const txId = txIds.pop();
+  const txData = await downloadTx(txId);
+  const { tx, meta } = txData;
+
+  const parsedTx: FullTx = parseTx(tx);
+
+  if (parsedTx.parents.length > 2) {
```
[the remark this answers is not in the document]
pedroferreira1: We discussed outside this thread and we can leave like this for now.

pedroferreira1 (approved):

README.md
```diff
@@ -0,0 +1,37 @@
+# Hathor Wallet Service -- Sync Daemon
+
+## State Machine
+
+The state machine diagram can be visualized at https://gist.github.com/andreabadesso/7299c0ed0ce189bc121a06dce1e11638
```
msbrogli, line 5: Nice to have this gist but I think we should copy the latest image here (or render automatically if it is possible).

README.md
```diff
@@ -0,0 +1,37 @@
+# Hathor Wallet Service -- Sync Daemon
+
```
msbrogli, line 2: It would be nice to have simple commands to install, run a local environment, build, update staging, and deploy. If it gets too big, we can move to separate files.

src/index.ts
```diff
@@ -0,0 +1,57 @@
...
+
+import { interpret } from 'xstate';
+import { SyncMachine } from './machine';
+import { Connection } from '@hathor/wallet-lib';
+
+const machine = interpret(SyncMachine).start();
+
+machine.onTransition(state => {
+  if (state.changed) {
+    console.log('Transitioned to state: ', state.value);
```
msbrogli, line 16: Maybe we should use `winston` for logging? See an example here: https://github.com/HathorNetwork/hathor-wallet-headless/blob/master/src/logger.js

src/machine.ts
```diff
@@ -0,0 +1,122 @@
...
+  SyncSchema,
+  SyncContext,
+} from './types';
+
+export const syncHandler = (_context, _event) => (callback, onReceive) => {
+  console.log('Sync handler instantiated');
+  const iterator = syncToLatestBlock();
+  const asyncCall: () => void = async () => {
+    for (;;) {
+      const block = await iterator.next();
```
msbrogli, line 17: What's the type?

src/machine.ts
```diff
@@ -0,0 +1,122 @@
...
+
+      const { value, done } = block;
```
msbrogli, line 19: What's the types of value and done?

src/machine.ts
```diff
@@ -0,0 +1,122 @@
...
+      if (done) {
+        console.log('Done!', value)
+        break;
+      }
+
+      if (value && !value.success) {
+        console.log('Erroed!', value.message);
```
msbrogli, line 26: Maybe `console.error(...)`?

src/machine.ts
```diff
@@ -0,0 +1,122 @@
...
+        callback('ERROR');
+        return;
+      }
+
+      if (value.type === 'reorg') {
+        console.log(value.message);
```
msbrogli, line 32: Shouldn't we prefix with `REORG`?

src/machine.ts
```diff
@@ -0,0 +1,122 @@
...
+        callback('REORG');
+        return;
+      }
+
+      if (value.type === 'finished') {
+        console.log('FINISHED!');
+        callback('DONE');
+      }
+
+      console.log(value);
```
msbrogli, line 42: I feel we will log the same message multiple times. Should we put it in an else block?

src/machine.ts
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    return;
+  };
+
+
+  onReceive((e) => {
```
msbrogli, line 49: Missing docstring and the type of `e`.

src/types.ts
```diff
@@ -0,0 +1,158 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+export interface Block {
```
msbrogli, line 8: Are these types duplicate from other repo?

tsconfig.json
```diff
@@ -0,0 +1,27 @@
+{
+  "include": ["src", "types"],
+  "compilerOptions": {
+    "module": "esnext",
+    "lib": ["dom", "esnext"],
+    "importHelpers": true,
+    "declaration": true,
+    "sourceMap": true,
+    "rootDir": "./src",
+    "strict": false,
```
msbrogli, line 10: Maybe we should be as strict as possible. What do you think?

src/utils.ts
```diff
@@ -0,0 +1,564 @@
...
+  PreparedOutput,
+} from './types';
+import AWS from 'aws-sdk';
+import axios from 'axios';
+import dotenv from 'dotenv';
+import { constants } from '@hathor/wallet-lib';
+
+dotenv.config();
+
+AWS.config.update({
```
msbrogli, line 29: I feel we should move all API methods to an `api.js` file. We can split it into `api/lambda.js` and `api/fullnode.js` if it makes sense.

src/utils.ts
```diff
@@ -0,0 +1,564 @@
...
+  }
+
+  const txId = txIds.pop();
+  const txData = await downloadTx(txId);
+  const { tx, meta } = txData;
+  const parsedTx: FullTx = parseTx(tx);
+
+  if (parsedTx.parents.length > 2) {
+    // We downloaded a block, we should ignore it
+    return recursivelyDownloadTx(blockId, txIds, data);
```
msbrogli, line 228: Why do we call it again? I feel we could be diving too deep into the recursion using this strategy. I was expecting to simply return when we want to skip something.

Maybe we should have an outside loop calling this method for all `txIds`.

src/utils.ts
```diff
@@ -0,0 +1,564 @@
...
+
+      if (input.tokenData === 0) {
+        return {
+          ...baseInput,
+          token: '00',
+        };
+      }
+
+      /* eslint-disable */
+      const { uid } = tx.tokens[(input.tokenData & TOKEN_INDEX_MASK) - 1]; // eslint-disable-line no-bitwise
```
msbrogli, line 274: Don't we have it in hathor-wallet-lib? We should at least move to a separate method.

src/utils.ts
```diff
@@ -0,0 +1,564 @@
...
+
+      if (output.tokenData === 0) {
+        return {
+          ...baseOutput,
+          token: '00',
+        };
+      }
+
+      /* eslint-disable */
+      const { uid } = tx.tokens[(output.tokenData & TOKEN_INDEX_MASK) - 1]; // eslint-disable-line no-bitwise
```
msbrogli, line 300: Same here.

src/utils.ts
```diff
@@ -0,0 +1,564 @@
...
+
+  return prepared;
+};
+
+/**
+ * Types a tx that was received from the full_node
+ *
+ * @param tx - The transaction object as received by the full_node
+ */
+export const parseTx = (tx: any): FullTx => {
```
msbrogli, line 318: I feel we should move it to a `utils.js` file.

src/utils.ts
```diff
@@ -0,0 +1,564 @@
...
+  const fullNodeBestBlock: Block = await getFullNodeBestBlock();
+
+  // Check if our best block is still in the fullnode's chain
+  const ourBestBlockInFullNode = await getBlockByTxId(ourBestBlock.txId, true);
+
+  if (!ourBestBlockInFullNode.success) {
+    yield {
+      type: 'error',
+      success: false,
+      message: 'Could not validate our best block',
```
msbrogli, line 409: Suggestion: "Best block not found in the full-node. Reorg?!"

src/utils.ts
msbrogli: It should never happen. Anyway, we should assert `outBestBlock.height <= meta.height` (or just verify and raise an exception as we are doing here).

src/utils.ts
```diff
@@ -0,0 +1,564 @@
...
+    yield {
+      type: 'reorg',
+      success: false,
+      message: 'Our height is higher than the wallet-service\'s height, we should reorg.',
+    };
+
+    return;
+  }
+
+  console.log('Best block is valid.');
```
msbrogli, line 440: We could log `Downloading {meta.height - outBestBlock.height} blocks...`.

src/utils.ts
```diff
@@ -0,0 +1,564 @@
...
+        message: `Failure on block ${preparedBlock.tx_id}`,
+      };
+
+      success = false;
+
+      break;
+    }
+
+    // Exclude duplicates:
+    const uniqueTxs: FullTx[] = txs.reduce((acc: FullTx[], tx: FullTx): FullTx[] => {
```
msbrogli, line 476: O(n^2)? I feel we can do it using a dictionary in O(n). It can get really slow if a block has too many transactions.

src/utils.ts
```diff
@@ -0,0 +1,564 @@
...
+
+  yield {
+    success,
+    type: 'finished',
+  };
+}
+
+export class LRU {
+  max: number;
+  cache: Map<string, any>;
```
msbrogli, line 526: Just leave a note that `Map` remembers the original insertion order.

msbrogli (changes requested):

README.md
andreabadesso: Documented how to run locally, how to build the docker image and how to run using docker

The commit is: bff6a3fc384a9f5630b5a68ce2c9d7ce13cac787

Not sure if I should explain more in depth how to run the fullnode and the wallet-service. What do you think?

README.md
andreabadesso: Updated the URL to render automatically from my gist, here is the URL I added to the description:
https://xstate.js.org/viz/?gist=7299c0ed0ce189bc121a06dce1e11638

Commit hash is cb1a5b36aaa63d354997cb646a0f5ff72b8b49ce

msbrogli: Great to have the most updated link. I also think we should have a copy in the repository (in case the link stops working).

README.md
msbrogli: It seems good enough to me. Thanks!

src/index.ts
andreabadesso: I agree, done in: d13009045fbf876a0edbb2c7a9cc24abe6132a7d

The difference between my implementation and the Headless's is that I removed the config file entirely, making the log level the only configuration available (on an env variable).

I think it is better to leave the responsibility of managing logs to the deploy mechanism

src/machine.ts
andreabadesso: Done in d13009045fbf876a0edbb2c7a9cc24abe6132a7d

src/machine.ts
andreabadesso: Done in d13009045fbf876a0edbb2c7a9cc24abe6132a7d

src/machine.ts
andreabadesso: Done in d13009045fbf876a0edbb2c7a9cc24abe6132a7d

src/types.ts
andreabadesso: There are some shared types with the Wallet Service, yes

But we use all types defined in this file on the sync-daemon, why?

src/utils.ts
andreabadesso: I agree.

Done in c5e1e4b4b2b7be18707e3e1c61f18e5bd758fe49

src/utils.ts
andreabadesso: Done in c5e1e4b4b2b7be18707e3e1c61f18e5bd758fe49

src/utils.ts
andreabadesso: Done in c0ce52c8fcdf93e7f1fe4603658b5af0980379a8

src/utils.ts
andreabadesso: Done in c0ce52c8fcdf93e7f1fe4603658b5af0980379a8

src/utils.ts
andreabadesso: Done in e3ae097005dab38dbe371f3f7865ba3c6dd1e710

src/utils.ts
andreabadesso: Done in e3ae097005dab38dbe371f3f7865ba3c6dd1e710

src/utils.ts
andreabadesso: I agree. I left this to be refactored later and forgot about it, sorry

Done in f654dbe6c0f0dce363b8348c33be758f05df87b5

src/utils.ts
andreabadesso: Done in 1b3435e83e095882bd247ff524c68a6fbc2c6612

src/machine.ts
andreabadesso: Done in 908acaf8781cd29e8ac0735cf83a02f3b6f77cc1

src/machine.ts
andreabadesso: Done in 908acaf8781cd29e8ac0735cf83a02f3b6f77cc1

`value` is a `StatusEvent`
`done` is a boolean, indicating if the generator has reached its end

src/machine.ts
andreabadesso: Done in 6d804938dc8037d93db33c734dc9c13de2c981fc

tsconfig.json
andreabadesso: Done in 07b9c1e5f82729e0ffa9b2c7a695498527d6013d

I had to set "strictFunctionTypes" and "noImplicitAny" to false

`strictFunctionTypes`: Conflict with `xstate` internal types
`noImplicitAny`: I'm using any on the received objects from the fullnode websocket, I could type the raw input, but we already have a `prepareTx` method to manually type the data (https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon/blob/07b9c1e5f82729e0ffa9b2c7a695498527d6013d/src/utils.ts#L84)

src/utils.ts
andreabadesso: I was thinking about an "impossible" situation where the fullnode API could answer with a block on `parents[1]` or `parents[2]` for some reason

In this case, we want to continue download the transaction list on `txIds`

By default, we never call this method with blocks, so we would never dive too deep

src/machine.ts
msbrogli: Thanks!

src/machine.ts
msbrogli: Thanks!

src/types.ts
msbrogli: I felt I've already seen those. I guess there's no easy way to share these types between the projects, right?

tsconfig.json
msbrogli: Can't we skip these verification specifically at these points (instead of disabling them in the whole project)?

src/utils.ts
msbrogli: I really liked it. Thanks!

src/utils.ts
msbrogli: That's impossible by design. Blocks must have their parent block as first element in the parent list.

Hum, I'm still a little concerned because the DAG can get really deep sometimes. Maybe add a protection that will fail and report a problem or move to a DFS implementation using a stack (instead of a recursion).

src/utils.ts
msbrogli: Great. Thanks!

src/types.ts
andreabadesso: Maybe we could create a library to contain types common on Hathor typescript projects, I can create an todo for that on our backlog and an issue here to use it when it is ready. What do you think?

tsconfig.json
andreabadesso: I agree.

Done in 3b7f2ea05186cac6680b215e34e6cf5e7ded3ae7 and 135191c30b26579d2cc043823875553b106e40e1

src/index.ts
msbrogli: Ok. We can easily change it in the future.

src/types.ts
msbrogli: Shouldn't we move these types to the `hathor-wallet-lib` for now? I'm assuming that these types are pretty universal and they are useful in several different places.

tsconfig.json
msbrogli: That's great! Thanks!

msbrogli (approved):

src/types.ts
andreabadesso: I agree, it would be nice to have all common types centralized somewhere and I think the wallet-lib is a nice place to put them for now

src/utils.ts
andreabadesso: Created an issue for this refactor here at https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon/issues/6

PR #4, opened 2021-04-26
Added dockerfile to the project

andreabadesso:


pedroferreira1 (approved):

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+COPY package.json /app/
+RUN cd /app && npm install --global --unsafe-perm tsdx@0.14.1
+RUN npm install --production
+
+COPY . /app/
+
+RUN cd /app && npm run build
+
+FROM node:14-alpine3.13 AS builder
```
msbrogli, line 17: Why are we using the builder? Just to get the `tsdx` out of the final image?

Dockerfile
```diff
@@ -0,0 +1,22 @@
+# Copyright 2020 Hathor Labs
+# This software is provided ‘as-is’, without any express or implied
+# warranty. In no event will the authors be held liable for any damages
+# arising from the use of this software.
+# This software cannot be redistributed unless explicitly agreed in writing with the authors.
+
+FROM node:14 AS builder
+
+COPY package.json /app/
+RUN cd /app && npm install --global --unsafe-perm tsdx@0.14.1
```
msbrogli, line 10: Shouldn't we add the `tsdx` to the package.json?

msbrogli (approved):

Dockerfile
andreabadesso: I am using a multi-stage Dockerfile because the alpine image won't run `tsdx` so we build on a larger base image (~300MB) and run on the alpine base image (~100MB)

Dockerfile
andreabadesso: We already have it on `devDependencies`, I agree with you and removed the global tsdx from the dockerfile on b4fa93ab4ac3b7dd6eaf75d48df6a47d9879c08d

Dockerfile
msbrogli: Great! Thanks!

Dockerfile
msbrogli: Great.

msbrogli (approved):

PR #5, opened 2021-05-17
feat: ignoring genesis transactions before sending them to the wallet…

andreabadesso:
We are already doing this on the wallet service, but the wallet service will throw an error, causing the daemon to stop sending blocks.

With this PR, the daemon will not even send the transaction if it is a genesis tx

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -39,6 +39,20 @@ import logger from './logger';
 
 dotenv.config();
 
+export const IGNORE_TXS = {
+  mainnet: [
+    '000006cb93385b8b87a545a1cbb6197e6caff600c12cc12fc54250d39c8088fc',
+    '0002d4d2a15def7604688e1878ab681142a7b155cbe52a6b4e031250ae96db0a',
+    '0002ad8d1519daaddc8e1a37b14aac0b045129c01832281fb1c02d873c7abbf9',
+  ],
+  testnet: [
+    '0000033139d08176d1051fb3a272c3610457f0c7f686afbe0afe3d37f966db85',
+    '00e161a6b0bee1781ea9300680913fb76fd0fac4acab527cd9626cc1514abdc9',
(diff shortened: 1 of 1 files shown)


msbrogli (approved):

src/utils.ts
```diff
@@ -39,6 +39,20 @@ import logger from './logger';
 
 dotenv.config();
 
+export const IGNORE_TXS = {
```
msbrogli, line 42: I don't like having these hashes here. Maybe move to the constants file?

pedroferreira1: Do we have the wallet-lib installed here? I think we should have the genesis hashes there and use it in the wallets. We already have this in the wallet-service, explorer and daemon

src/utils.ts
```diff
@@ -315,8 +338,12 @@ export async function* syncToLatestBlock(): AsyncGenerator<StatusEvent> {
       for (const key of Object.keys(uniqueTxs)) {
       const preparedTx: PreparedTx = prepareTx(uniqueTxs[key]);
 
+
       try {
-        const sendTxResponse: ApiResponse = await sendTx(preparedTx);
+        const sendTxResponse: ApiResponse = await sendTx({
+          ...preparedTx,
+          height: preparedBlock.height,
```
pedroferreira1, line 345: Why did you add this key?

pedroferreira1 (approved):

src/utils.ts
andreabadesso: We don't, but I agree that would be a good idea.

I'll merge and open an issue to refactor it when it's ready

src/utils.ts
andreabadesso: This was added by mistake on this branch, it is part of some changes I am doing to the sync daemon to handle reorgs

Removed it from last commit and force pushed it.

PR #7, opened 2021-06-01
feat: handle reorg on the daemon

andreabadesso:


src/utils.ts
```diff
@@ -97,9 +99,14 @@ export const recursivelyDownloadTx = async (blockId: string, txIds: string[] = [
     return recursivelyDownloadTx(blockId, txIds, data);
   }
 
-  const newParents = parsedTx.parents.filter((parent) => txIds.indexOf(parent) < 0 && parent !== txId);
+  // check if we have already downloaded the parents
```
pedroferreira1, line 102: Can you add another comment here explaining each condition? It was a long time since I've seen this code and I was trying to understand why did you add `!data.has(parent)`.

For what I understand you check `txIds` for the txs that are in the array to search but were not downloaded yet, `parent` if the tx is a parent of itself and added the `!data.has(parent)` in order to prevent downloading a tx that was already downloaded and is in the data map, right?

src/machine.ts
```diff
@@ -116,7 +119,25 @@ export const SyncMachine = Machine<SyncContext, SyncSchema>({
...
+
+          return;
+        },
+        onDone: {
+          target: 'idle',
+        },
+        onError: {
+          target: 'failure',
+        },
+      }
```
pedroferreira1, line 140: This comment is not related to the changes in this PR, they are good for me. I was just re-reading the code where you set the 'reorg' type for the machine because it was already handled before.

I was just thinking about this line here: https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon/blob/master/src/utils.ts#L260

I think we should try to consider this a reorg if `fullNodeBestBlock.height > ourBestBlock.height` but maybe we could just continue sending error to the machine and we handle this manually (we could just give more information to the error handler, e.g. `fullNodeBestBlock` and `ourBestBlock` heights)

What do you think?

pedroferreira1 (approved):

src/utils.ts
andreabadesso: Added the comments on 1ac06f4b1532df1cb3c1ac3cbca100cd91390bde

Exactly as you described. The data is serving as a "seen" list for the BFS, so we don't add transactions that we already downloaded to the queue

src/machine.ts
andreabadesso: I think we can just transition to `reorg` and try to handle it on the wallet service

Done in b1871dcece4ca1b0e6b056ebcec6c0c58a7740b8

src/api/lambda.ts
```diff
@@ -53,14 +53,20 @@ export const lambdaCall = (fnName: string, payload: any): Promise<any> => new Pr
...
             return reject(e.message);
           }
         }
       });
 });
 
+export const invokeReorg = async (): Promise<ApiResponse> => {
+  const response = await lambdaCall('onHandleReorgRequest', {});
+
+  return response;
```
r4mmer, line 67: Will we need the response for anything? If not, shouldn't we just return the promise `lambdaCall(...)`?

src/utils.ts
```diff
@@ -63,16 +63,18 @@ const TX_CACHE_SIZE: number = parseInt(process.env.TX_CACHE_SIZE as string) || 2
...
+export const recursivelyDownloadTx = async (blockId: string, txIds: string[] = [], data = new Map<string, FullTx>()): Promise<Map<string, FullTx>> => {
   if (txIds.length === 0) {
     return data;
   }
 
   const txId: string = txIds.pop() as string;
-  const network = process.env.NETWORK || 'mainnet';
+  const network: string = process.env.NETWORK || 'mainnet';
+
+  if (IGNORE_TXS.has(network)) {
```
r4mmer, line 74: Should we allow an unknown network? And maybe the txs on `IGNORE_TXS` should be in the wallet-lib constants.

r4mmer (approved):

src/api/lambda.ts
andreabadesso: I agree. Done in f5bc6f4184d2ad6a86c8cd31f5d91ae927a7652a

src/utils.ts
andreabadesso: I don't think we should allow an unknown network. What do you think, @pedroferreira1 ?

I agree that the txs should be in the wallet-lib, I will create an issue there.

pedroferreira1: In the lib we already raise exception if we try to use a network different than 'mainnet' or 'testnet', so I think it's fine to handle only them.

About the txs in the lib I agree and I believe we've created an issue about that (or should've created) because we've already discussed this

src/api/lambda.ts
```diff
@@ -40,7 +40,7 @@ export const lambdaCall = (fnName: string, payload: any): Promise<any> => new Pr
 
       lambda.invoke(params, (err, data) => {
         if (err) {
-          logger.error('err', err);
+          logger.error(err);
```
msbrogli, line 43: Should we add a prefix? Or does it automatically add information about where the error was logged from?

msbrogli (approved):

src/api/lambda.ts
andreabadesso: Yes we should, it does not add information automatically.

Done in 73bd6bf3c8dbab8b7a5598b6ea1cb8abd1c558c1

PR #11, opened 2021-06-14
feat: sync mempool

r4mmer:
Creating another state machine for the mempool sync to run along the block sync state machine.

### DIFF
--- a/src/api/fullnode.ts
+++ b/src/api/fullnode.ts
@@ -30,6 +30,7 @@ export const downloadTx = async (txId: string, noCache: boolean = false) => {
     return globalCache.get(txId);
   }
 
+  console.log(`Gonna download ${txId} from ${DEFAULT_SERVER}`);
   const response = await axios.get(`${DEFAULT_SERVER}transaction?id=${txId}`);
 
   if (!noCache) {
--- a/src/api/fullnode.ts
+++ b/src/api/fullnode.ts
@@ -40,6 +40,18 @@ export const downloadTx = async (txId: string, noCache: boolean = false) => {
 };
 
 /**
+ * Returns a list of transactions on the mempool
+ *
+ */
+export const downloadMempool = async () => {
+  // Maybe api could return a list of tx ids (to make this call lighter in case mempool gets too big)
+  // with a list of tx ids the daemon should use the list to download each tx by id
+  // Use cache to prevent download of known tx if it's still in the mempool on other runs
+  const response = await axios.get(`${DEFAULT_SERVER}mempool`);
+  return response.data;
+};
+
+/**
  * Returns a `FullBlock` downloaded from the full_node
  *
  * @param height - The block's height
--- a/src/api/fullnode.ts
+++ b/src/api/fullnode.ts
@@ -16,6 +16,7 @@ import {
 } from '../types';
 import axios from 'axios';
 import { globalCache } from '../utils';
+import logger from '../logger';
 
 const DEFAULT_SERVER = process.env.DEFAULT_SERVER || 'https://node1.foxtrot.testnet.hathor.network/v1a/';
 
@@ -30,7 +31,6 @@ export const downloadTx = async (txId: string, noCache: boolean = false) => {
     return globalCache.get(txId);
   }
 
-  console.log(`Gonna download ${txId} from ${DEFAULT_SERVER}`);
   const response = await axios.get(`${DEFAULT_SERVER}transaction?id=${txId}`);
 
   if (!noCache) {
@@ -50,6 +50,7 @@ export const downloadMempool = async () => {
   const data = response.data;
 
   if (!data.success) {
+    logger.error(data);
     throw new Error('Mempool API failure');
   }
   return data;
--- a/src/api/fullnode.ts
+++ b/src/api/fullnode.ts
@@ -45,11 +45,14 @@ export const downloadTx = async (txId: string, noCache: boolean = false) => {
  *
  */
 export const downloadMempool = async () => {
-  // Maybe api could return a list of tx ids (to make this call lighter in case mempool gets too big)
-  // with a list of tx ids the daemon should use the list to download each tx by id
-  // Use cache to prevent download of known tx if it's still in the mempool on other runs
   const response = await axios.get(`${DEFAULT_SERVER}mempool`);
-  return response.data;
+
+  const data = response.data;
+
+  if (!data.success) {
+    throw new Error('Mempool API failure');
+  }
+  return data;
 };
 
 /**
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,7 +6,7 @@
  */
 
 import { interpret } from 'xstate';
-import { SyncMachine } from './machine';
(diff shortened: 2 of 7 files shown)


src/machine.ts
```diff
@@ -78,6 +81,58 @@ export const syncHandler = () => (callback, onReceive) => {
...
+      }
+
+      if (value && !value.success) {
+        logger.error(value.message);
+        // if there is any error calling the lambda, wait 30 seconds and retry
+        if (value.type == 'wait') {
+          callback('WAIT');
+          return;
+        }
+        callback('ERROR');
```
andreabadesso, line 106: I like the retry mechanism! Maybe we should add that to the block sync as well 

For what types of errors will we transition to the `failure` state? Maybe add a comment explaining

src/utils.ts
```diff
@@ -242,6 +262,69 @@ export const parseTx = (tx: RawTx): FullTx => {
...
+      };
+      return;
+    }
+
+    const preparedTx: PreparedTx = prepareTx(tx);
+
+    try {
+      const sendTxResponse: ApiResponse = await sendTx(preparedTx);
+      if (!sendTxResponse.success) {
+        logger.debug(sendTxResponse);
```
andreabadesso, line 303: Maybe use `logger.error` instead of `logger.debug`?

src/utils.ts
```diff
@@ -242,6 +262,69 @@ export const parseTx = (tx: RawTx): FullTx => {
...
+  if (!mempoolResp.success) {
+    yield {
+      success,
+      type: 'wait',
+      message: 'Could not download from mempool api',
+    };
+    return;
+  }
+
+  for (let i = 0; i < mempoolResp.transactions.length; i++) {
```
andreabadesso, line 286: Maybe we should keep track of the transactions we have already sent on a FIFO queue to prevent unnecessary lambda calls

I know that we currently have very few transactions on the `mempool` and even fewer of those are voided, but maybe the list will increase if someone is "voiding" transactions on purpose

andreabadesso (changes requested):

src/utils.ts
r4mmer: Maybe use `LRU` like `globalCache`?

src/utils.ts
r4mmer: Done.

src/machine.ts
r4mmer: `failure` will be entered when the fullnode mempool api fails, any other error could be mitigated by waiting for the next block (which should take up to 30s).

src/api/fullnode.ts
```diff
@@ -39,6 +40,21 @@ export const downloadTx = async (txId: string, noCache: boolean = false) => {
...
+ * Returns a list of transactions on the mempool
+ *
+ */
+export const downloadMempool = async () => {
+  const response = await axios.get(`${DEFAULT_SERVER}mempool`);
+
+  const data = response.data;
+
+  if (!data.success) {
+    throw new Error('Mempool API failure');
```
andreabadesso, line 53: Maybe log `data` so we can see the response on the logs

src/index.ts
```diff
@@ -13,12 +13,8 @@ import { Connection } from '@hathor/wallet-lib';
...
 
 // @ts-ignore
-const machine = interpret(SyncMachine).start();
-
-machine.onTransition(state => {
-  if (state.changed) {
-    logger.debug('Transitioned to state: ', state.value);
-  }
+const machine = interpret(SyncMachine).onTransition(state => {
+  console.log(`Sync on state: ${state.value}`);
```
andreabadesso, line 17: Use `logger`

src/machine.ts
```diff
@@ -78,6 +79,65 @@ export const syncHandler = () => (callback, onReceive) => {
...
+
+    return;
+  };
+
+  onReceive((e: HandlerEvent) => {
+    switch(e.type) {
+      case 'START':
+        asyncCall();
+        break
+      case 'NEW_BLOCK':
```
andreabadesso, line 127: _Semi-related to line number:_

I think we should some tests to check if the machine is really pausing the mempool sync and transitioning to block sync if a `NEW_BLOCK` event is received and add a comment here explaining that we are pausing the mempool sync because the blocks sync has priority so we have more chance to have all transaction dependencies

andreabadesso (changes requested):

src/machine.ts
```diff
@@ -86,13 +146,46 @@ export const SyncMachine = Machine<SyncContext, SyncSchema>({
   initial: 'idle',
   context: {
     hasMoreBlocks: false,
+    hasMempoolUpdate: false,
   },
   states: {
     idle: {
       always: [
         { target: 'syncing', cond: 'hasMoreBlocks' },
+        { target: 'mempoolsync', cond: 'hasMempoolUpdate' },
```
andreabadesso, line 155: Add a comment explaining that if both conditions are true, it will target the first one (syncing) as per https://xstate.js.org/docs/guides/guards.html#multiple-guards

src/utils.ts
```diff
@@ -242,6 +262,69 @@ export const parseTx = (tx: RawTx): FullTx => {
...
+
+  yield {
+    success: true,
+    type: 'finished',
+  };
+}
+
 /**
  * Syncs to the latest block
  *
```
andreabadesso, line 330: _Unrelated to line number:_

Update the gist on README with the new machine schema, you can do that by copying the code (and removing/mocking imports) and pasting to https://xstate.js.org/viz/, logging in with github and saving the gist

src/api/fullnode.ts
r4mmer: Done.

src/index.ts
r4mmer: Done.

src/machine.ts
r4mmer: I tested this behavior on another simpler machine, it stops and  transitions to the block sync state.

src/machine.ts
r4mmer: Done.

src/utils.ts
r4mmer: Done.

andreabadesso (approved):

src/index.ts
```diff
@@ -13,11 +13,9 @@ import { Connection } from '@hathor/wallet-lib';
 import logger from './logger';
 
 // @ts-ignore
-const machine = interpret(SyncMachine).start();
-
-machine.onTransition(state => {
-  if (state.changed && state.value) {
-    logger.debug(`Transitioned to state: ${state.value}`);
+const machine = interpret(SyncMachine).onTransition(state => {
```
r4mmer, line 16: Merge conflict: I kept my decoupling of the `machine.start()` method. This can be useful later if we have the creating of the instances (machines, handlers, etc.) separated from the init of the service.

src/utils.ts
r4mmer: Resolved not to include a cache on this PR.

PR #13, opened 2021-07-06
chore: pre-deploy adjustments

luislhl:
### What was done
- Fixed Dockerfile build
- Changed name of some env vars for clarity

### DIFF
--- a/scripts/build-and-push-docker.sh
+++ b/scripts/build-and-push-docker.sh
@@ -1,0 +1,17 @@
+
+if [ -z "$AWS_ACCOUNT_ID" ]; then
+    echo "Please export a AWS_ACCOUNT_ID env var before running this";
+    exit 1;
+fi
+
+if [ -z "$DOCKER_IMAGE_TAG" ]; then
+    commit=`git rev-parse HEAD`;
+    timestamp=`date +%s`;
+    export DOCKER_IMAGE_TAG="dev-$commit-$timestamp";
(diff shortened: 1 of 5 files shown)


andreabadesso (approved):
lgtm

scripts/build-and-push-docker.sh
```diff
@@ -0,0 +1,17 @@
...
+fi
+
+if [ -z "$DOCKER_IMAGE_TAG" ]; then
+    commit=`git rev-parse HEAD`;
+    timestamp=`date +%s`;
+    export DOCKER_IMAGE_TAG="dev-$commit-$timestamp";
+fi;
+
+echo $DOCKER_IMAGE_TAG;
+aws ecr get-login-password --region eu-central-1 | docker login --username AWS --password-stdin $AWS_ACCOUNT_ID.dkr.ecr.eu-central-1.amazonaws.com;
```
msbrogli, line 14: Should we exit in case one of these commands fail? We can use something like this:

```bash
{ cmd } || { echo "error"; exit -1; }}
```

msbrogli (approved):

scripts/build-and-push-docker.sh
luislhl: Good point.

I've added this to the script, to assure that:
```
set -e
set -o pipefail
```

Do you think it's enough?

PR #15, opened 2021-07-13
fix: ignoring NFT transaction outputs with undecoded scripts

andreabadesso:
NFT transactions were causing the daemon to fail on `sendTransaction` as we are not able to decode the tx_output script

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -11,8 +11,8 @@ export interface Block {
 }
 
 export interface DecodedScript {
-  type: string;
-  address: string;
+  type?: string;
+  address?: string;
   timelock?: number | undefined | null;
   value?: number | undefined | null;
   tokenData?: number | undefined | null;
@@ -222,12 +222,15 @@ export interface RawDecodedInput {
   token_data: number;
 }
 
+/* Everything is optional because scripts that were not able to
+ * be decoded will be returned as {}
+ */
 export interface RawDecodedOutput {
-  type: string;
-  address: string;
+  type?: string;
+  address?: string;
   timelock?: number | null;
-  value: number;
-  token_data: number;
+  value?: number;
+  token_data?: number;
 }
 
 export interface RawInput {
@@ -276,6 +279,7 @@ export interface Meta {
   accumulated_weight: number;
   score: number;
   height: number;
+  validation?: string;
   first_block?: string | null;
 }
 
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -138,38 +138,29 @@ export const recursivelyDownloadTx = async (blockId: string, txIds: string[] = [
       !data.has(parent)
   });
 
-  let newData;
-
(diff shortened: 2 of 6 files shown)


src/types.ts
```diff
@@ -276,6 +279,7 @@ export interface Meta {
   accumulated_weight: number;
   score: number;
   height: number;
+  validation?: string;
```
pedroferreira1, line 282: Why do you need this?

src/utils.ts
```diff
@@ -137,9 +138,30 @@ export const recursivelyDownloadTx = async (blockId: string, txIds: string[] = [
...
+ * @param tx - `FullTx` to remove the invalid outputs
+ */
+export const cleanInvalidOutputs = (tx: FullTx) => ({
+  ...tx,
+  // Filter outputs that we can't handle (script was unable to be decoded)
+  outputs: tx.outputs.filter((output) => {
+    const validDecoded = !isNil(get(output, 'decoded.type'));
+
+    if (!validDecoded) {
+      logger.warn(`Ignoring tx output from tx ${tx.txId} as script couldn't be decoded.`);
```
pedroferreira1, line 158: Important to log the output index as well.

pedroferreira1 (approved):

src/types.ts
andreabadesso: We don't, but it is part of the `meta` object that comes from the full-node:

https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon/pull/15/files#diff-1a258b987746216dc81c746069ca5944c0f1ec848aa095a8a9f109a976dc2343R366

luislhl (approved):

src/utils.ts
andreabadesso: Fixed in 65d9a5a613fe35fc73c0c1fee05f83f74c30b324

src/types.ts
```diff
@@ -222,12 +222,15 @@ export interface RawDecodedInput {
   token_data: number;
 }
 
+/* Everything is optional because scripts that were not able to
+ * be decoded will be returned as {}
+ */
 export interface RawDecodedOutput {
```
r4mmer, line 228: Same here

src/types.ts
```diff
@@ -11,8 +11,8 @@ export interface Block {
 }
 
 export interface DecodedScript {
```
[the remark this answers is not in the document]
andreabadesso: Since this will need a refactor everywhere that uses `DecodedScript`, I will create an issue to refactor this and merge this PR as we discussed in Slack

src/types.ts
andreabadesso: Same as above

src/types.ts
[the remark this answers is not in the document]
andreabadesso: https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon/issues/16

PR #17, opened 2021-07-21
feat: websocket conn error log

andreabadesso:
# Motivation

The `Connection` class from `hathor-wallet-lib` did not log anything on connection failures from the WebSocket, this PR uses the new `connection_error` event from the lib to log the connection failures so we can detect them when debugging connection issues

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -47,14 +47,21 @@ const handleMessage = (message: any) => {
       if (message.state === Connection.CONNECTED) {
         machine.send({ type: 'NEW_BLOCK' });
       }
+      if (message.state === Connection.CONNECTING) {
+        logger.error(`Websocket is attempting to connect to ${process.env.DEFAULT_SERVER}`);
+      }
+      if (message.state === Connection.CLOSED) {
+        logger.error('Websocket connection was closed.');
+      }
       break;
   }
 };
 
-machine.start();
-
 const DEFAULT_SERVER = process.env.DEFAULT_SERVER;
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -47,14 +47,21 @@ const handleMessage = (message: any) => {
       if (message.state === Connection.CONNECTED) {
         machine.send({ type: 'NEW_BLOCK' });
       }
+      if (message.state === Connection.CONNECTING) {
+        logger.error(`Websocket is attempting to connect to ${process.env.DEFAULT_SERVER}`);
```
pedroferreira1, line 51: This is not always an error, the connection process always start with connecting before being connected. Not sure if this can be bad for your debug but it's semantically wrong to this all CONNECTING messages as errors.

src/index.ts
```diff
@@ -47,14 +47,21 @@ const handleMessage = (message: any) => {
...
+      }
+      if (message.state === Connection.CLOSED) {
+        logger.error('Websocket connection was closed.');
```
pedroferreira1, line 54: Here I think I agree with the error. We expect the websocket connection to be always opened with the full node

pedroferreira1 (approved):

src/index.ts
andreabadesso: I agree! Fixed in 39c81a146d0d8e69c1cafa09e4cd5cdbeb34d882

luislhl (approved):

PR #19, opened 2021-08-31
ci: configure automated deployment

luislhl:
### Acceptance Criteria
- We should build and push a Docker image to our registry, when commits/tags are created.
- We should tag the images according to the branch/tag:
    - `dev` branch: `dev-$commit-$timestamp` pattern
    - `master` branch: `master-$commit-$timestamp` pattern
    - git tags: the tag is also used as the Docker tag.
- We should send a message to Slack when new images are pushed

### Additional Notes
The failing tests are caused by an old CI pipeline that we had in this repo, not by the one created in this PR. It didn't fail before because Github Actions was disabled in this repo.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -2,7 +2,7 @@ name: deploy
 
 on:
   push:
-    branches: [dev, master]
+    branches: [dev, master, ci/deploy]
     tags: ['v*.*.*']
 
 env:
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           node-version: ${{ matrix.node }}
 
       - name: Install deps and build (with cache)
-        uses: bahmutov/npm-install@v1
+        uses: bahmutov/npm-install@c67aaab58a864ea2873950cde9c1c9379f9f711a
```
pedroferreira1, line 23: What's this change?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,57 @@
...
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Push Dev Image
+        if: github.ref == 'refs/heads/dev'
+        run: |
+          make build-and-push-docker
```
pedroferreira1, line 23: Here you don't need to export the `DOCKER_IMAGE_TAG`?

pedroferreira1 (approved):

.github/workflows/main.yml
luislhl: Since this is a third-party action, it's safer to use the commit hash instead of a tag.

https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions

.github/workflows/deploy.yml
luislhl: When it's not exported, the script creates a default for us with the `dev-` prefix

andreabadesso (approved):

PR #34, opened 2022-03-21
refactor: sending alerts to sqs instead of logging

andreabadesso:
### Acceptance Criteria
- Alerts should be send directly to the alerts SQS instead of just logging them

### Security Checklist


.eslintrc.yml
```diff
@@ -0,0 +1,84 @@
+parser: "@typescript-eslint/parser"
```
andreabadesso, line 1: *Comment unrelated to line number*

This is the `eslintrc` we are using on the Wallet Service

luislhl:
I"ve talked to Abadesso on Slack about this.

Maybe we should make this through our log-manager solution, because it has mechanisms intended to avoid spam in our channels and in our alert queue.

But to do this, we would need to log in JSON here

r4mmer (approved):
Many uses (with `Severity.INFO`) seem like they're not actually an alert so I think another name (e.g. `sendLog`) would be better for the `addAlert`.

luislhl:
> I've talked to Abadesso on Slack about this.
> 
> Maybe we should use our log-manager solution for this, because it has mechanisms intended to avoid spam in our channels and in our alert queue.
> 
> But to do this, we would need to log in JSON here

After further discussion, this will be done in a new issue, and we will proceed with this PR

src/api/lambda.ts
```diff
@@ -67,6 +67,43 @@ export const lambdaCall = (fnName: string, payload: any): Promise<any> =>
...
+    severity,
+    metadata,
+    application: process.env.APPLICATION_NAME,
+  };
+
+  const sqs = new AWS.SQS({ apiVersion: '2015-03-31' });
+
+  const params = {
+    MessageBody: JSON.stringify(preparedMessage),
+    QueueUrl: process.env.ALERT_QUEUE_URL as string,
```
luislhl, line 90: Just remember we will need to add this in our infra code as well

src/api/lambda.ts
```diff
@@ -67,6 +67,43 @@ export const lambdaCall = (fnName: string, payload: any): Promise<any> =>
...
+      None: {
+        DataType: 'String',
+        StringValue: '--',
+      },
+    }
+  };
+
+  sqs.sendMessage(params, (err) => {
+    if (err) {
+      logger.error('[ALERT] Erroed while sending message to the alert sqs queue');
```
luislhl, line 101: This means we shouldn't disable the old alerting mechanism yet, right?

luislhl (approved):

src/api/lambda.ts
andreabadesso: That is correct, a failure to create the alert should be an alert

PR #46, opened 2022-05-20
refactor: recursivelyDownloadTx now downloads the input tree as well

andreabadesso:
Fixes https://github.com/HathorNetwork/hathor-wallet-service/issues/227

### Acceptance Criteria
- We should get not only the transactions that have the block we are downloading as first_block but also the input tree for that transaction, ignoring transactions that were confirmed by previous blocks

### Security Checklist

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -99,6 +99,7 @@ export const downloadTxFromId = async (
  */
 export const recursivelyDownloadTx = async (
   blockId: string,
+  blockHeight: number,
   txIds: string[] = [],
   data = new Map<string, FullTx>()
 ): Promise<Map<string, FullTx>> => {
@@ -114,7 +115,7 @@ export const recursivelyDownloadTx = async (
 
     if (networkTxs.includes(txId)) {
       // Skip
-      return recursivelyDownloadTx(blockId, txIds, data);
+      return recursivelyDownloadTx(blockId, blockHeight, txIds, data);
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -124,24 +129,32 @@ export const recursivelyDownloadTx = async (
...
     // We downloaded a block, we should ignore it
-    return recursivelyDownloadTx(blockId, txIds, data);
+    return recursivelyDownloadTx(blockId, blockHeight, txIds, data);
   }
 
-  // If the first_block from the downloaded tx is not from the block we are searching, we should ignore it.
-  if (meta.first_block !== blockId) {
-    return recursivelyDownloadTx(blockId, txIds, data);
+  // Transaction is already confirmed
+  if (meta.first_block) {
```
pedroferreira1, line 136: Maybe we can have here `if (meta.first_block && meta.first_block !== blockId)`?

In that case, we prevent doing this extra request to get the full block from the full node again.

pedroferreira1 (approved):

src/utils.ts
andreabadesso: Done in 94ca49b33d78c9ec6d4c11251c6cfcaec94016ba

We wouldn't go to the fullnode there tho since we have a local cache and the block was already downloaded but it's nice to avoid it

src/utils.ts
```diff
@@ -124,24 +129,32 @@ export const recursivelyDownloadTx = async (
...
+    }
   }
 
-  // check if we have already downloaded the parents
-  const newParents = parsedTx.parents.filter(parent => {
+  const inputList = parsedTx.inputs.map((input) => input.txId);
+  const txList = [...parsedTx.parents, ...inputList];
+
+  // check if we have already downloaded any of the transactions of the new list
+  const newTxIds = txList.filter(transaction => {
```
msbrogli, line 150: This algorithm is O(n^2). We can easily create a set in O(n) and then run this algorithm in O(n) since the search on the set will run in O(1). I think we should add a note explaining that it gets slower and slower as the DFS dives into the DAG. I also think we should create an issue to get this algorithm fixed soon.

src/utils.ts
```diff
@@ -124,24 +129,32 @@ export const recursivelyDownloadTx = async (
...
-  // If the first_block from the downloaded tx is not from the block we are searching, we should ignore it.
-  if (meta.first_block !== blockId) {
-    return recursivelyDownloadTx(blockId, txIds, data);
+  // Transaction was already confirmed by a different block in the past
+  if (meta.first_block && meta.first_block !== blockId) {
+    const firstBlockResponse = await downloadTx(meta.first_block);
+
+    // If the transaction was confirmed by an older block, ignore it as it was
+    // already sent to the wallet-service
+    if (firstBlockResponse.tx.height < blockHeight) {
```
msbrogli, line 141: Should we add an assert when `firstBlockResponse.tx.height === blockHeight`?

msbrogli (approved):

src/utils.ts
andreabadesso: I think we can just refactor `txIds` as a Set so we don't have to create it on every iteration, I will do it in another PR

Added the comment you suggested on c290b6ee588542ecdf3c52ebe75c34c1099bd00b

src/utils.ts
andreabadesso: Why? Is there any chance the fullnode would respond a different block here?

PR #84, opened 2023-01-09
fix: handling throwing lambda call

andreabadesso:
This is related to the error described in https://github.com/HathorNetwork/on-call-incidents/issues/87

### Acceptance Criteria
- If the lambda call throws instead of gracefully returning an error, we should catch it and handle it properly by sending the alert to alert-manager and transitioning to the `failure` state.

### Security Checklist

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -588,13 +588,18 @@ export async function* syncToLatestBlock(): AsyncGenerator<StatusEvent> {
       }
     }
 
-    // We will send the block only after all transactions were sent
-    // to be sure that all downloads were succesfull since there is no
-    // ROLLBACK yet on the wallet-service.
-    const sendBlockResponse: ApiResponse = await sendTx(preparedBlock);
+    try {
+      // We will send the block only after all transactions were sent
+      // to be sure that all downloads were succesfull since there is no
+      // ROLLBACK yet on the wallet-service.
+      const sendBlockResponse: ApiResponse = await sendTx(preparedBlock);
 
-    if (!sendBlockResponse.success) {
-      logger.debug(sendBlockResponse);
+      if (!sendBlockResponse.success) {
+        logger.debug(sendBlockResponse);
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -588,13 +588,18 @@ export async function* syncToLatestBlock(): AsyncGenerator<StatusEvent> {
...
+      // to be sure that all downloads were succesfull since there is no
+      // ROLLBACK yet on the wallet-service.
+      const sendBlockResponse: ApiResponse = await sendTx(preparedBlock);
+
+      if (!sendBlockResponse.success) {
+        logger.debug(sendBlockResponse);
+        throw new Error('Erroed sending block to the wallet-service.');
+      }
+    } catch(e) {
+      logger.error(e);
```
alexruzenhack, line 602: I see no purpose to throw an error to be captured in its own error handling because the error constructor will use memory for the stack trace without necessity. It would be nice to work with a return object:

```
const sendTxHandler: TxResult = async (preparedBlock) => {
  try {
    await response = await sendTx(preparedBlock);
    if (!response.success) {
      logger.debug(response);
    } 
    return { success: response.success };
  }
  catch(e) {
    logger.error(e);
    return { success: false }
  }
}

const { success } = sendTxHandler(preparedBlock);
if (!success) {
  addAlert...
}
```

Being the exception handler a function `sendTxHandler`, it turns the code more readable and more efficient in relating to the memory.

alexruzenhack (approved):
I just suggest in the comment a way to give the same behavior but with an increase in readability.

luislhl (approved):

src/utils.ts
andreabadesso: I like the idea! Thanks

Refactor made in 6fce84723454b38da4ddd052ed3ed9ad88778029

alexruzenhack (approved):

PR #89, opened 2023-09-25
feat: realiable integrations [part 1 - machine]

andreabadesso:
### Acceptance Criteria

- [ ] We should include the initial project structure, including typescript builder, packages and a Dockerfile
- [ ] We should include the machine, with mocked services

### Security Checklist


.envrc
```diff
@@ -0,0 +1,7 @@
+if [[ $(type -t use_flake) != function ]]; then
+  echo "ERROR: use_flake function missing."
+  echo "Please update direnv to v2.30.0 or later."
+  exit 1
+fi
+
```
r4mmer, line 6: Since we are configuring a new tool for the project, should we add a description on the README?

PR #90, opened 2023-09-25
feat: reliable integrations [part 2]

andreabadesso:
### Acceptance Criteria
- We should include the services and all dependencies (db, types and utils methods)
- We should include the websocket actor

### Security Checklist


.github/workflows/main.yml
```diff
@@ -9,8 +9,8 @@ jobs:
     runs-on: ${{ matrix.os }}
```
r4mmer, line 9: I tihnk the actions checkout v2 and setup-node v1 are close to being disabled, maybe we should update to the latest versions

src/actors/WebSocketActor.ts
```diff
@@ -0,0 +1,71 @@
...
+import { Event } from '../machines/types';
+// @ts-ignore
+import { get } from 'lodash';
+import logger from '../logger';
+
+const WS_URL = process.env.WS_URL;
+if (!WS_URL) {
+  logger.error('WS_URL is not defined.');
+  process.exit(1);
+}
```
r4mmer: Maybe we should have a config module, so that we can know which configurations are needed for the service.

src/actors/index.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+export { default as WebSocketActor } from './WebSocketActor';
+export * from './WebSocketActor';
```
r4mmer: Why do we need this line?

src/guards/index.ts
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { Context, Event } from "../machines/types";
+import { TxCache } from "../machines";
+import { hashTxData } from "../utils";
+
+export const metadataIgnore = (_context: Context, event: Event) => {
```
r4mmer, line 17: It would be nice to have docstrings explainging why, when and how each method is used.

src/machines/SyncMachine.ts
```diff
@@ -0,0 +1,298 @@
...
+      },
+    },
+    [SYNC_MACHINE_STATES.CONNECTING]: {
+      entry: assign({
+        socket: () => spawn(WebSocketActor),
+      }),
+      on: {
+        WEBSOCKET_EVENT: [{
+          cond: 'websocketDisconnected',
+          target: 'RECONNECTING',
```
r4mmer: Should we use the `SYNC_MACHINE_STATES.RECONNECTING`?
This way we not only know the string value but that the target is part of a known and mapped state from `SYNC_MACHINE_STATES`.

src/machines/SyncMachine.ts
```diff
@@ -0,0 +1,298 @@
...
+  validateNetwork: 'validateNetwork',
+  handlingUnhandledEvent: 'handlingUnhandledEvent',
+  handlingMetadataChanged: 'handlingMetadataChanged',
+  handlingVertexAccepted: 'handlingVertexAccepted',
+  handlingVoidedTx: 'handlingVoidedTx',
+  handlingUnvoidedTx: 'handlingUnvoidedTx',
+  handlingFirstBlock: 'handlingFirstBlock',
+};
+
+const SyncMachine = Machine<Context, any, Event>({
```
r4mmer, line 78: _unrelated to line_

Maybe we should take this PR as an opportunity to upgrade xstate to v5 since v4 is not maintained anymore.
There is a guide for migration from v4 to v5 [here](https://stately.ai/docs/migration).

src/machines/SyncMachine.ts
```diff
@@ -0,0 +1,298 @@
...
+            onError: `#${SYNC_MACHINE_STATES.ERROR}`,
+          },
+        },
+        [CONNECTED_STATES.handlingUnvoidedTx]: {
+          id: CONNECTED_STATES.handlingUnvoidedTx,
+          invoke: {
+            src: 'handleUnvoidedTx',
+            data: (_context: Context, event: Event) => event,
+            onDone: {
+              target: `#${CONNECTED_STATES.handlingVertexAccepted}`,
```
r4mmer, line 227: Why do we add the `#` on the state? couldn't find anything about this on the docs

src/machines/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as SyncMachine } from './SyncMachine';
+export * from './SyncMachine';
```
r4mmer, line 2: Why the second export?

src/types.ts
```diff
@@ -0,0 +1,556 @@
...
+  // eslint-disable-next-line camelcase
+  token_data: number;
+  script: string;
+  token: string;
+  decoded: DecodedOutput | null;
+}
+
+export interface StringMap<T> {
+  [x: string]: T;
+}
```
r4mmer: Not really requesting a change, but is there any difference between this and:
```ts
export interface StringMap<T> = Record<string, T>;
```

I'm just wondering because map and plain object have some differences that can cause some confusion.

.github/workflows/main.yml
andreabadesso: Yes, I guess we can do that on the monorepo PR, what do you think?

src/guards/index.ts
andreabadesso: Added in ee4c59bab5771bba6ddc4b4228eda6b000463c73

src/machines/SyncMachine.ts
andreabadesso: Yes we should, added in d45cf4da54c90a29606f5c02af73bc1e1978d8a4

src/machines/SyncMachine.ts
andreabadesso: I don't think we should for now, v5 is in beta and the migration should be pretty straight forward as I already updated some deprecated methods

src/machines/SyncMachine.ts
andreabadesso: It's an [identified state node](https://xstate.js.org/docs/guides/ids.html#identifying-state-nodes), this way it's easier to transition

src/actors/WebSocketActor.ts
andreabadesso: I agree, done in 7c8bdfac125663b9068778863dab417ea206dc48

src/actors/index.ts
andreabadesso: We don't, removed in 1c080c98f4bba927f70afe836735a7afc8680f52

src/machines/index.ts
andreabadesso: So we can do something like:

```typescript
import { SyncMachine, CONNECTED_STATES, SYNC_MACHINE_STATES } from '../machines/SyncMachine';
```

src/config.ts
```diff
@@ -0,0 +1,66 @@
...
+};
+
+// The service name to go with the logs
+export const SERVICE_NAME = process.env.SERVICE_NAME ?? 'wallet-service-daemon';
+// The default log level
+export const CONSOLE_LEVEL = process.env.CONSOLE_LEVEL ?? 'debug';
+// Number of transactions to cache in the LRU in-memory cache
+export const TX_CACHE_SIZE = parseInt(process.env.TX_CACHE_SIZE ?? '10000', 10);
+// Number of blocks before unlocking a block utxo
+export const BLOCK_REWARD_LOCK = parseInt(process.env.BLOCK_REWARD_LOCK ?? '10', 10);
```
glevco, line 36: Shouldn't this come from the full node?

src/machines/types.ts
```diff
@@ -0,0 +1,67 @@
...
+        height: number;
+      };
+    }
+  }
+}
+
+export interface Context {
+  socket: ActorRef<any, any> | null;
+  retryAttempt: number;
+  event: unknown;
```
glevco: Why `unknown`?

src/machines/types.ts
```diff
@@ -0,0 +1,67 @@
...
+  | { type: 'CONNECTED' }
+  | { type: 'DISCONNECTED' };
+
+export type MetadataDecidedEvent = {
+  type: 'TX_VOIDED' | 'TX_UNVOIDED' | 'TX_NEW' | 'TX_FIRST_BLOCK' | 'IGNORE';
+  originalEvent: FullNodeEvent;
+}
+
+export type WebSocketSendEvent = {
+  // TODO: Well-defined types here
```
glevco: Should we do this in this PR?

src/utils.ts
```diff
@@ -0,0 +1,400 @@
...
+  clear(): void {
+    this.cache = new Map();
+  }
+}
+
+export const isAuthority = (tokenData: number): boolean => (
+  (tokenData & constants.TOKEN_AUTHORITY_MASK) > 0    // eslint-disable-line no-bitwise
+);
+
+export const prepareOutputs = (outputs: EventTxOutput[], tokens: string[]): TxOutputWithIndex[] => {
```
glevco: Maybe add a docstring for this? There are also others missing, but this one is more complex

src/utils.ts
```diff
@@ -0,0 +1,400 @@
...
+ * @returns A map of addresses and its token balances
+ */
+export const getAddressBalanceMap = (
+  inputs: TxInput[],
+  outputs: TxOutput[],
+): StringMap<TokenBalanceMap> => {
+  const addressBalanceMap = {};
+
+  for (const input of inputs) {
+    const address = input.decoded?.address;
```
glevco: We throw `if !output.decoded` below, shouldn't we do the same here?

src/utils.ts
```diff
@@ -0,0 +1,400 @@
...
+ * @param mysql - Database connection
+ * @param now - Current timestamp
+ */
+export const unlockTimelockedUtxos = async (mysql: MysqlConnection, now: number): Promise<void> => {
+  const utxos: DbTxOutput[] = await getExpiredTimelocksUtxos(mysql, now);
+
+  await unlockUtxos(mysql, utxos, true);
+};
+
+export const prepareInputs = (inputs: EventTxInput[], tokens: string[]): TxInput[] => {
```
glevco: Also docstring

src/utils.ts
glevco, line 1: It seems that some of the functions in this file are more general than others, and also that some of them are related. Maybe break this file into smaller ones, where it makes sense?

src/types.ts
glevco, line 1: Should we group related types in specific files? For example one for DB models, one for WebSocket models, etc

src/types.ts
```diff
@@ -0,0 +1,556 @@
...
+   */
+  static merge(b1: Balance, b2: Balance): Balance {
+    let lockExpires = null;
+    if (b1.lockExpires === null) {
+      lockExpires = b2.lockExpires;
+    } else if (b2.lockExpires === null) {
+      lockExpires = b1.lockExpires;
+    } else {
+      // @ts-ignore
+      lockExpires = Math.min(b1.lockExpires, b2.lockExpires);
```
glevco: Why min? Wouldn't part of the balance only expire later?

src/types.ts
```diff
@@ -0,0 +1,556 @@
...
+  }
+
+  /**
+   * Create a TokenBalanceMap from a TxOutput.
+   *
+   * @param output - The transaction output
+   * @returns The TokenBalanceMap object
+   */
+  static fromTxOutput(output: TxOutput): TokenBalanceMap {
+    // TODO check if output.decoded exists, else return null
```
glevco: Is this going to be done in this PR?

src/machines/SyncMachine.ts
```diff
@@ -0,0 +1,304 @@
...
+    initialEventId: null,
+    txCache: new LRU(TX_CACHE_SIZE),
+  },
+  states: {
+    [SYNC_MACHINE_STATES.INITIALIZING]: {
+      invoke: {
+        src: 'fetchInitialState',
+        onDone: {
+          actions: ['storeInitialState'],
+          target: 'CONNECTING',
```
glevco: Same as https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon/pull/90#discussion_r1364119163

src/machines/SyncMachine.ts
```diff
@@ -0,0 +1,304 @@
...
+    [SYNC_MACHINE_STATES.CONNECTING]: {
+      entry: assign({
+        socket: () => spawn(WebSocketActor),
+      }),
+      on: {
+        WEBSOCKET_EVENT: [{
+          cond: 'websocketDisconnected',
+          target: 'RECONNECTING',
+        }, {
+          target: 'CONNECTED',
```
glevco: Same as https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon/pull/90#discussion_r1364119163

src/machines/SyncMachine.ts
```diff
@@ -0,0 +1,304 @@
...
+        }],
+      },
+    },
+    [SYNC_MACHINE_STATES.RECONNECTING]: {
+      onEntry: ['clearSocket', 'increaseRetry'],
+      after: {
+        BACKOFF_DELAYED_RECONNECT: 'CONNECTING',
```
glevco: Same as https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon/pull/90#discussion_r1364119163

src/machines/SyncMachine.ts
```diff
@@ -0,0 +1,304 @@
...
+          },
+        },
+        [CONNECTED_STATES.handlingFirstBlock]: {
+          id: CONNECTED_STATES.handlingFirstBlock,
+          invoke: {
+            src: 'handleTxFirstBlock',
+            data: (_context: Context, event: Event) => event,
+            onDone: {
+              target: 'idle',
+              actions: ['storeEvent', 'sendAck', 'updateCache'],
```
glevco, line 241: It looks like we have some repeated patterns in the state machine definition, for example, multiple states with `actions: ['storeEvent', 'sendAck', 'updateCache']`.

Thinking of possible futures changes, it could be easy to change one of them and forget the others. Would it make sense to somehow group those repeated patterns? If they're all semantically related, maybe they could share a variable like `FINISH_HANDLING_EVENT_ACTIONS = ['storeEvent', 'sendAck', 'updateCache']`, or maybe even a single action that calls the other three. This is just an example, I'm not sure if this would be beneficial for readability/maintainability, but I guess it's worth raising the question.

src/actors/WebSocketActor.ts
```diff
@@ -0,0 +1,68 @@
...
+  receive((event: Event) => {
+    if (event.type !== 'WEBSOCKET_SEND_EVENT') {
+      logger.warn('Message that is not websocket_send_event reached the websocket actor');
+
+      return;
+    }
+
+    if (!socket) {
+      logger.error('Received event but no socket yet');
+    }
```
glevco: Should we return here?

src/actions/index.ts
```diff
@@ -0,0 +1,120 @@
...
+  initialEventId: (_context: Context, event: Event) => {
+    // @ts-ignore
+    return event.data.lastEventId;
+  },
+});
+
+export const unwrapEvent = assign({
+  event: (_context: Context, event: Event) => {
+    if (event.type !== 'METADATA_DECIDED') {
+      return;
```
glevco: Shouldn't it be `return event`?

src/actions/index.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    return {
+      type: 'WEBSOCKET_SEND_EVENT',
+      event: {
+        message: JSON.stringify({
+          type: 'START_STREAM',
+          window_size: 1,
+          // @ts-ignore
+          last_ack_event_id: lastAckEventId,
+        }),
```
glevco: Should we create a model for WS requests? So it's easier to update in case the schema changes in `hathor-core`

src/actions/index.ts
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    const eventId = get(event, 'event.event.id', -1);
+    const contextEventId = get(context, 'event.id', -1);
+
+    if (eventId === -1) return;
+
+    if (contextEventId > -1) {
+      // @ts-ignore
+      if (event.event.event.id < context.event.event.id) {
```
glevco: Is it possible to improve those repeated `event` names? Maybe specify what they are, or where they come from, like in `updateCache` you used `fullNodeEvent`. Maybe change this name at the context level?

src/actions/index.ts
glevco, line 1: Add docstrings for each action?

src/guards/index.ts
```diff
@@ -0,0 +1,183 @@
...
+
+/*
+ * This guard is used during the `handlingMetadataChanged` to check if
+ * the result was a TX_FIRST_BLOCK event, which means that we should insert
+ * the height of this transaction to the database
+ */
+export const metadataFirstBlock = (_context: Context, event: Event) => {
+  if (event.type !== 'METADATA_DECIDED') {
+    return false;
+  }
```
glevco: Instead of `false`, wouldn't it be an error to call this guard when the type is not `METADATA_DECIDED`? The same question for other guards in this file.

src/guards/index.ts
glevco, line 1: Should we create constants for the strings in this file?

src/services/index.ts
```diff
@@ -0,0 +1,448 @@
...
+  getLastSyncedEvent,
+  getTxOutputsFromTx,
+  markUtxosAsVoided,
+  cleanupVoidedTx,
+} from '../db';
+import getConfig from '../config';
+import logger from '../logger';
+
+export const metadataDiff = async (_context: Context, event: Event) => {
+  const mysql = await getDbConnection();
```
glevco, line 70: Is this creating a new connection? If it's expensive, should this be persisted in the context so it can be reused, or something like this?

src/services/index.ts
```diff
@@ -0,0 +1,448 @@
...
+
+    if (first_block
+       && first_block.length
+       && first_block.length > 0) {
+      if (!dbTx.height) {
+        return {
+          type: 'TX_FIRST_BLOCK',
+          originalEvent: event,
+        };
+      }
```
glevco: What if a transaction gets a first block in the same event it gets unvoided?

src/services/index.ts
```diff
@@ -0,0 +1,448 @@
...
+
+    const dbTx: DbTransaction | null = await getTransactionById(mysql, hash);
+
+    // Db is already on the database, ignore...
+    if (dbTx) {
+      logger.debug(`Transaction ${hash} already on the database, ignoring..`);
+      mysql.destroy();
+
+      return;
+    }
```
glevco: If we call `handleVertexAccepted` with an existing tx, shouldn't it be an error?

src/services/index.ts
```diff
@@ -0,0 +1,448 @@
...
+
+    if (version === hathorLib.constants.CREATE_TOKEN_TX_VERSION
+       && token_name
+       && token_symbol) {
+      if (!token_name || !token_symbol) {
+        logger.error('Processed a token creation event but it did not come with token name and symbol');
+        process.exit(1);
+      }
+      await storeTokenInformation(mysql, hash, token_name, token_symbol);
+    }
```
glevco: Why do we `process.exit(1)` instead of `throw new Error()` like in the block height check above?

glevco: Also, shouldn't we remove the `&& token_name && token_symbol` from the first `if`, so it can fall in the second?

src/services/index.ts
glevco, line 1: I noticed that there's a lot of logic related to locks in this project, both time locks and block height locks. Maybe a future improvement for the reliable integration in the full node would be to send some events related to unlocking, where possible, so this logic could be removed from the client's responsibility. What do you think?

src/services/index.ts
```diff
@@ -0,0 +1,448 @@
...
+      timestamp,
+      version,
+      weight,
+    );
+
+    // Add utxos
+    await addUtxos(mysql, hash, txOutputs, heightlock);
+    await updateTxOutputSpentBy(mysql, txInputs, hash);
+
+    // Handle genesis parent txs:
```
glevco: Why genesis?

src/services/index.ts
```diff
@@ -0,0 +1,448 @@
...
+        // might need to generate new addresses to keep maxGap
+        await addNewAddresses(mysql, walletId, newAddresses, lastUsedAddressIndex);
+        // update existing addresses' walletId and index
+      }
+      // update wallet_balance and wallet_tx_history tables
+      const walletBalanceMap: StringMap<TokenBalanceMap> = getWalletBalanceMap(addressWalletMap, addressBalanceMap);
+      await updateWalletTablesWithTx(mysql, hash, timestamp, walletBalanceMap);
+    }
+
+    // TODO: Send message on SQS  for real-time update
```
glevco, line 290: Should this be done in this PR?

src/services/index.ts
```diff
@@ -0,0 +1,448 @@
...
+
+      if (!dbTxOutput) {
+        throw new Error('Transaction output different from database output!');
+      }
+
+      return {
+        ...output,
+        locked: dbTxOutput.locked,
+      };
+    });
```
glevco: If we are locking the outputs, shouldn't it be `locked: true` in the returned value?

src/services/index.ts
```diff
@@ -0,0 +1,448 @@
...
+      timestamp,
+      version,
+      weight,
+    } = fullNodeEvent.event.data;
+
+    let height: number | null = metadata.height;
+
+    if (!metadata.first_block) {
+      height = null;
+    }
```
glevco: Shouldn't we throw an error if `handleTxFirstBlock` is called and there's no `first_block`?

src/services/index.ts
```diff
@@ -0,0 +1,448 @@
...
+  const lastEvent = await getLastSyncedEvent(mysql);
+
+  mysql.destroy();
+
+  return { lastEventId: lastEvent?.last_event_id };
+};
+
+export const validateNetwork = async () => {
+  return;
+};
```
glevco: Missing implementation?

src/machines/SyncMachine.ts
```diff
@@ -0,0 +1,307 @@
...
+          id: CONNECTED_STATES.handlingUnvoidedTx,
+          invoke: {
+            src: 'handleUnvoidedTx',
+            data: (_context: Context, event: Event) => event,
+            onDone: {
+              // The handleUnvoidedTx will remove the tx from the database, we should
+              // re-add it:
+              target: `#${CONNECTED_STATES.handlingVertexAccepted}`,
+              // We shouldn't send ACK, as we'll send the ACK after handlingVertexAccepted
+              actions: ['storeEvent', 'updateCache'],
```
glevco, line 229: Should we have a new test to cover this recently found bug?

src/utils.ts
andreabadesso: Done in f470b901835644030fc9dffd9d8cb805bb9d9d7f

src/utils.ts
andreabadesso: I agree we should, done in b4f68e26e11974b71b755106481e2499a63683e6

src/utils.ts
andreabadesso: Fixed in f470b901835644030fc9dffd9d8cb805bb9d9d7f

src/services/index.ts
andreabadesso: As we discussed in slack, I think that would be great, would reduce a lot of the complexity with calculating balances here

src/machines/types.ts
andreabadesso: Yes we should, done in 10975e37792974ae73051d223c21a4ec12f3d34f

src/machines/types.ts
andreabadesso: Not sure, this should have been typed

Fixed in 4a5881a855bd30a5c74d617ff35ab82e6e8bd79b

src/services/index.ts
andreabadesso: It's not, we create a connection pool the first time this method is called and just return a connection from it on subsequent calls

src/services/index.ts
andreabadesso: Since we're removing all traces from it and then re-adding it from the data that came in the event, it should not be a problem, the transaction will be added with the height

src/services/index.ts
andreabadesso: That's a good point, it should

Added a throw in 73c8c91941e33c6c8ad8119426be7af7751c7c31

src/services/index.ts
andreabadesso: We should, I'm not sure why I were using process.exit, refactored in f40953f5e5699f530178b8d6868e60dfd6c8589a

andreabadesso: We absolutely should, fixed in f40953f5e5699f530178b8d6868e60dfd6c8589a

src/services/index.ts
andreabadesso: The comment is misleading, this comment is explaining the if case, genesis tx does not have inputs nor outputs

Refactored the comment in 3a8f4ee660eb344ea56439c0b5270c128d48b950

src/services/index.ts
andreabadesso: This comment is also misleading, we're not locking the outputs, we're just setting their locked status as it is in the database

Refactored in de59847c68babcbe00bc998df760924ac962ffb0

src/services/index.ts
andreabadesso: Yes we should, done in 67b6f912198a5dd023acc704cbded8584393776b

src/services/index.ts
andreabadesso: I would rather do it after the monorepo merge as this also requires some changes in the wallet-service

src/types.ts
andreabadesso: Yes we should, done in those two commits:

52cd7bd71d433ac0af04e5e97fde48e671c78fcd
a1580afe25096f638d20222fe61627c8a16fe0bb

src/guards/index.ts
andreabadesso: Yes we should, done in 2f4b1e7954d61b8c24ed46dfdf7ee4d8d0653891

src/guards/index.ts
andreabadesso: Yes we should done in 13676a315d387f2b62f554bfbfde1b5ac78797db, 30002cdee17bfb91ccf2a3a55746ce75f0d808aa and 599ffb3bae319d42bc269221a928e83911bc0172

src/services/index.ts
andreabadesso: Refactored in 517953574c153b62a912b6505bcd6259ef5766e4

src/actions/index.ts
andreabadesso: Done in 23d9ec93a199b7c8af4da315dd30bc7df336c71d

src/actions/index.ts
andreabadesso: Refactored at 919175acc8fdfab00a36e196d092c9bdfa4c2591

What do you think?

src/config.ts
andreabadesso: That would be great, is there any APIs we can fetch it from?

src/utils.ts
andreabadesso: Done in 00a24596b3a93c8f488e42650d0662b130a721a9

src/types.ts
andreabadesso: We can actually throw an error here, done in 4f2cecd11cae5de57147cd432a3da74150656d70

src/machines/SyncMachine.ts
andreabadesso: Done in 83bbd6add31a1fbc8ff67dcfdaeb8d946f0f2e57

src/machines/SyncMachine.ts
andreabadesso: Done in 83bbd6add31a1fbc8ff67dcfdaeb8d946f0f2e57

src/machines/SyncMachine.ts
andreabadesso: Done in 83bbd6add31a1fbc8ff67dcfdaeb8d946f0f2e57

src/actors/WebSocketActor.ts
andreabadesso: Yes we should ifxed in 0dc5f661c659cecb819d8765ee6c831b7cc1d896

src/actions/index.ts
andreabadesso: I guess `context.event` would work here, to return the old event, but I think we should throw because this should never happen

Done in 420973d25f648f4995ac75a8a74789445a4ceda3

src/actions/index.ts
andreabadesso: Done in f3f029b4076e6e9df3c398a19a197488bf6a4124

src/config.ts
glevco: Yes, it's the `reward_spend_min_blocks` key in `/v1a/version`. Example: https://node1.mainnet.hathor.network/v1a/version

andreabadesso: Great, I'll refactor it to use this API in another PR

src/machines/SyncMachine.ts
andreabadesso: I will create a integration test that cover this and the other bugs we found in another PR, after the monorepo PR is merged

src/machines/SyncMachine.ts
andreabadesso: Even though I agree with your point that it could be easy to change one of them and forget the others, I think I prefer being explicit about what actions are taken on each state migration so it's clear when reading the machine

src/config.ts
glevco: Could you create a KTLO issue for this and link it here, please? So we can keep track of it

src/types.ts
glevco: @andreabadesso Could you create a KTLO issue for this and link it here, please? So we can keep track of it

src/machines/SyncMachine.ts
glevco: Ok, thanks!

src/machines/SyncMachine.ts
glevco: Could you create a KTLO issue for this and link it here, please? So we can keep track of it

r4mmer (approved):

src/types.ts
andreabadesso: Issue: https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon/issues/94

src/machines/SyncMachine.ts
andreabadesso: Issue: https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon/issues/95

src/config.ts
andreabadesso: Issue: https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon/issues/95

src/machines/SyncMachine.ts
andreabadesso: Yes

Issue: https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon/issues/96

src/config.ts
glevco: Thanks!

src/types.ts
glevco: Thanks!

src/machines/SyncMachine.ts
glevco: Thanks!

glevco (approved):

PR #91, opened 2023-10-02
feat: reliable integrations [part 1]

andreabadesso:
### Acceptance Criteria
- We should remove all files from the old daemon

### Security Checklist


package.json
```diff
@@ -8,60 +8,45 @@
...
-    "aws-sdk": "^2.878.0",
-    "axios": "^0.21.1",
+    "@hathor/wallet-lib": "^1.0.2",
+    "@types/jest": "^29.5.4",
+    "@types/ws": "^8.5.5",
+    "assert": "^2.1.0",
+    "aws-sdk": "^2.1454.0",
+    "axios": "^0.21.4",
     "dotenv": "^8.2.0",
+    "jest": "^29.6.4",
```
glevco: Shouldn't jest be a dev dependency?

glevco (approved):

package.json
```diff
@@ -8,60 +8,45 @@
     "src"
   ],
   "engines": {
-    "node": ">=10"
+    "node": ">=18"
```
r4mmer, line 11: We need to check if the lib works well with node 18.

package.json
r4mmer: Same for the @types/jest

r4mmer (approved):

package.json
andreabadesso: Fixed in aca440380aea3b7eb464f02a9d2a3a10fd1b9eb2

package.json
andreabadesso: I did a full sync using it, apparently it's working properly

PR #99, opened 2023-11-30
chore: wallet-service monorepo [part 2] - CI

andreabadesso:
### Acceptance Criteria

- We should run tests for the wallet-service lambdas on every push
- We should run tests for the daemon on every push
- We should update the aws-sdk to v3 as v2 was deprecated
- We should update jest to the latest version

### Security Checklist


packages/daemon/__tests__/db/index.test.ts
```diff
@@ -843,7 +843,7 @@ describe('address and wallet related tests', () => {
     for (const [index, address] of addressesInfo.addresses.entries()) {
       expect(ADDRESSES[index]).toBe(address);
     }
-  } );
+  }, 15000);
```
andreabadesso, line 846: It times out with less than 15s

packages/daemon/__tests__/services/services.test.ts
```diff
@@ -401,24 +401,6 @@ describe('handleVertexAccepted', () => {
...
-          id: 'idValue',
-        },
-      },
-    };
-
-    const mockTransaction = { tx_id: 'hashValue' };
-    (getTransactionById as jest.Mock).mockResolvedValue(mockTransaction);
-
-    await expect(handleVertexAccepted(context as any, {} as any)).rejects.toThrow('Transaction hashValue already in the database, this should never happen');
-  });
```
andreabadesso, line 421: This condition was removed on the reliable-integrations part 2 PR, this test was failing

packages/daemon/package.json
```diff
@@ -35,7 +35,7 @@
     "typescript": "^4.9.5"
   },
   "dependencies": {
-    "@hathor/wallet-lib": "^1.0.2",
+    "@hathor/wallet-lib": "^0.39.0",
```
andreabadesso, line 38: We should update the wallet-lib on another PR for both the daemon and the wallet-service

packages/wallet-service/src/utils.ts
```diff
@@ -7,6 +7,7 @@
 
 import { BinaryToTextEncoding, createHash } from 'crypto';
 
+import { Logger } from 'winston';
```
andreabadesso, line 10: Not sure why this happened, but the build was failing because of this missing dependency

packages/wallet-service/tests/utils.ts
```diff
@@ -1121,3 +1122,22 @@ export const insertPushDevice = async (mysql: ServerlessMysql, pushRegister: {
...
+    passphrase?: string,
+    networkName?: string
+  } = {}): bitcore.HDPrivateKey => {
+  const methodOptions = Object.assign({passphrase: '', networkName: 'mainnet'}, options);
+  const { passphrase, networkName } = methodOptions;
+
+  const network = new Network(networkName);
+  const code = new Mnemonic(seed);
+  return code.toHDPrivateKey(passphrase, network.bitcoreNetwork);
+};
```
andreabadesso, line 1142: I had to create my own `getXPrivKeyFromSeed` because the lib was having issues with its own bitcore-mnemonic dependency

I also had to add Hathor's network manually to be able to use the wallet-lib methods

packages/wallet-service/tests/api.test.ts
```diff
@@ -181,7 +182,7 @@ test('GET /addresses', async () => {
   expect(result.statusCode).toBe(STATUS_CODE_TABLE[ApiError.INVALID_PAYLOAD]);
   expect(returnBody.details).toHaveLength(1);
   expect(returnBody.details[0].message)
-    .toMatchInlineSnapshot('"\\"index\\" must be greater than or equal to 0"');
+    .toMatchInlineSnapshot('"\"index\" must be greater than or equal to 0"');
```
andreabadesso, line 185: Had to update all inline snapshots after jest update

packages/wallet-service/src/utils.ts
r4mmer: I believe its used as an argument type, maybe the LSP does not count type annotations as usage of the import.

r4mmer (approved):

glevco (approved):

PR #101, opened 2023-12-08
refactor: fetch reward min blocks

andreabadesso:
Fixes https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon/issues/95

### Acceptance Criteria
- We should fetch the reward_min_blocks from the fullnode and use it instead of an env variable

### Security Checklist

### DIFF
--- a/packages/daemon/src/config.ts
+++ b/packages/daemon/src/config.ts
@@ -33,8 +33,6 @@ export const SERVICE_NAME = process.env.SERVICE_NAME ?? 'wallet-service-daemon';
 export const CONSOLE_LEVEL = process.env.CONSOLE_LEVEL ?? 'debug';
 // Number of transactions to cache in the LRU in-memory cache
 export const TX_CACHE_SIZE = parseInt(process.env.TX_CACHE_SIZE ?? '10000', 10);
-// Number of blocks before unlocking a block utxo
-export const BLOCK_REWARD_LOCK = parseInt(process.env.BLOCK_REWARD_LOCK ?? '10', 10);
 
 // Fullnode information, used to make sure we're connected to the same fullnode
 export const FULLNODE_PEER_ID = process.env.FULLNODE_PEER_ID;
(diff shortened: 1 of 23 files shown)


packages/daemon/src/config.ts
```diff
@@ -48,13 +50,18 @@ export const DB_USER = process.env.DB_USER;
...
 
+// Other
+export const USE_SSL = process.env.USE_SSL;
+
 
 export default () => ({
   SERVICE_NAME,
   CONSOLE_LEVEL,
   TX_CACHE_SIZE,
   BLOCK_REWARD_LOCK,
```
glevco: Should we remove this, and its definition above?

packages/daemon/src/utils/helpers.ts
```diff
@@ -0,0 +1,20 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { USE_SSL, FULLNODE_HOST } from '../config';
+
+export const getFullnodeHttpUrl = () => {
+  const protocol = USE_SSL ? 'https://' : 'http://';
+
+  return `${protocol}${FULLNODE_HOST}/v1a/`;
```
r4mmer: If the wallet-service runs on nodejs we can use `path.join` to safely join the url parts.

r4mmer (commented):
The data from the fullnode is used with the name `rewardMinBlocks` but it is set on the initial state as `minRewardBlocks`, but I could not find where one is converted to the other.

packages/daemon/src/config.ts
andreabadesso: Yes we should, done in f4c2fd24fae2068e1961b52012015fb90f27b8a2

packages/daemon/src/utils/helpers.ts
andreabadesso: Updated it to use URL in dfe48c88978f587fad943ad167571f7a438e694f

andreabadesso:
> The data from the fullnode is used with the name `rewardMinBlocks` but it is set on the initial state as `minRewardBlocks`, but I could not find where one is converted to the other.

Refactored it to use `rewardMinBlocks` in 7c4e900b71a815bc8b17e15653b3675d7cc7030e

r4mmer (approved):

packages/daemon/src/config.ts
glevco: Thanks!

glevco (approved):

PR #103, opened 2023-12-13
tests: integration tests

andreabadesso:
### Acceptance Criteria
- We should setup an environment for integration tests to run on
- We should have integration tests to validate the final address balances from the fullnode scenarios (SINGLE_CHAIN_BLOCKS_AND_TRANSCTIONS, REORG, UNVOIDED_TRANSACTIONS)
- We should run integration tests on the CI

### Security Checklist


packages/daemon/__tests__/integration/config.ts
```diff
@@ -0,0 +1,20 @@
...
+export const UNVOIDED_SCENARIO_PORT = 8081;
+export const UNVOIDED_SCENARIO_LAST_EVENT = 38;
+
+// reorg
+export const REORG_SCENARIO_PORT = 8082;
+export const REORG_SCENARIO_LAST_EVENT = 19;
+
+// single chain blocks and transactions port
+export const SINGLE_CHAIN_BLOCKS_AND_TRANSACTIONS_PORT = 8083;
+export const SINGLE_CHAIN_BLOCKS_AND_TRANSACTIONS_LAST_EVENT = 37;
```
glevco: It looks like all `LAST_EVENT` numbers are short by one in comparison to event IDs in hathor-core tests 🤔

packages/daemon/__tests__/integration/docker-compose.yml
```diff
@@ -0,0 +1,41 @@
...
+      "--seed", "1"
+    ]
+    ports:
+      - "8082:8080"
+  single_chain_blocks_and_transactions:
+    image: hathornetwork/hathor-core:stable
+    command: [
+      "events_simulator",
+      "--scenario", "SINGLE_CHAIN_BLOCKS_AND_TRANSACTIONS",
+      "--seed", "1"
```
glevco: Shouldn't we let the seeds be random?

packages/daemon/__tests__/integration/utils.ts
```diff
@@ -0,0 +1,96 @@
...
+
+
+
+
+
+
+
+
+
+
```
glevco: Shouldn't these lines be reject by a linter?

packages/daemon/__tests__/integration/config.ts
andreabadesso: Yes, they're short by one because we don't store event ids on the database when we ignore an event and on all scenarios, the last event is an event we ignore

I will add a comment explaining this

packages/daemon/__tests__/integration/docker-compose.yml
andreabadesso: We could, but then we would have no way to check if the balance is correct as the fullnode's API is not active when using the simulator

It's fixed so we can have a list of addresses and their balances

packages/daemon/__tests__/integration/utils.ts
andreabadesso: Fixed the multiple empty lines

Just checked, and the linter is not running on the packages, I'll open another PR adding it

packages/daemon/__tests__/integration/config.ts
andreabadesso: Done in 97ead7b1a9f691a24d6f79da2e9ded84449c9b7f

glevco: Thanks!

packages/daemon/__tests__/integration/docker-compose.yml
glevco: Ok!

packages/daemon/__tests__/integration/utils.ts
glevco: Ok, thanks!

glevco (approved):

packages/daemon/__tests__/integration/docker-compose.yml
```diff
@@ -0,0 +1,41 @@
...
+      - "3380:3306"
+  unvoided_transaction:
+    image: hathornetwork/hathor-core:stable
+    command: [
+      "events_simulator",
+      "--scenario", "UNVOIDED_TRANSACTION",
+      "--seed", "1"
+    ]
+    ports:
+      - "8081:8080"
```
r4mmer: question(non-blocking): Since we are starting 3 hathor-cores (in simulation mode) should we check the memory and cpu requirements of each container?

packages/daemon/__tests__/integration/wait-for-db-up.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  while (attempts < maxAttempts) {
+    try {
+      const conn = await mysql.createConnection({
+        host: DB_ENDPOINT,
+        user: DB_USER,
+        password: DB_PASS,
+        port: DB_PORT,
+      });
+      console.log('Successfully connected to the database!');
```
r4mmer: suggestion: Should we make a `SELECT 1` query to test the connection?

packages/daemon/__tests__/integration/wait-for-db-up.ts
```diff
@@ -0,0 +1,42 @@
...
+  DB_USER,
+  DB_PASS,
+  DB_PORT,
+  DB_ENDPOINT,
+} from './config';
+
+const attemptConnection = async (maxAttempts: number, interval: number): Promise<void> => {
+  let attempts = 0;
+
+  while (attempts < maxAttempts) {
```
r4mmer: This will never trigger, should we just make it `while (true)`?

packages/daemon/__tests__/integration/scenarios/balances.test.ts
```diff
@@ -0,0 +1,201 @@
+/**
```
r4mmer: suggestion: Maybe we should have this test (the only one so far) on the `integration` folder and keep utilities, scenario configs and docker configuration on folders that indicate whats inside them.

what do you think?

packages/daemon/__tests__/integration/wait-for-db-up.ts
andreabadesso: Why wouldn't it trigger? It might if something wrong happens with the db instance

packages/daemon/__tests__/integration/wait-for-db-up.ts
andreabadesso: Yes! Done in 5810e307c9b6ddb68fafd25c08133644396847f0

packages/daemon/__tests__/integration/scenarios/balances.test.ts
andreabadesso: I agree, done in f94a27c9243a70c4ecf40c2602425ff6e93210c9

packages/daemon/__tests__/integration/wait-for-db-up.ts
r4mmer: Because if this is false it would enter the else in the catch block below, throwing an error before checking this condition.

Also, if we exit the while loop we would make the function silently return without errors but without making a connection to the database, maybe we should remove the else condition and throw the error outside the while.
What do you think?

packages/daemon/__tests__/integration/docker-compose.yml
r4mmer: From our conversation we gathered that each scenario adds up 60mbs of memory and an average of 0.04% cpu which makes running all three scenarios very light.

packages/daemon/__tests__/integration/wait-for-db-up.ts
andreabadesso: I agree! Done in ef3e8eeb19adee112ece38c661f21f2f480aa366

r4mmer (approved):

PR #104, opened 2023-12-18
fix: real time txs + push notification

andreabadesso:
### Acceptance Criteria
- We should publish new transactions to the NEW_TX_SQS queue

Note: I didn't migrate the tests in this PR, I suggest we do this in another PR so we can have shared code between the daemon and the lambdas

### Security Checklist

### DIFF
--- a/packages/daemon/src/db/index.ts
+++ b/packages/daemon/src/db/index.ts
@@ -4,7 +4,7 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  */
-import mysql, { Connection as MysqlConnection, Pool } from 'mysql2/promise';
+import mysql, { Connection as MysqlConnection, Pool, RowDataPacket } from 'mysql2/promise';
 import {
   TokenBalanceMap,
   DbTxOutput,
@@ -21,6 +21,7 @@ import {
   DbTransaction,
   TokenInfo,
   Miner,
+  TokenSymbolsRow,
 } from '../types';
 import { isAuthority } from '../utils';
 import {
@@ -1531,4 +1532,33 @@ export const cleanupVoidedTx = async (mysql: MysqlConnection, txId: string): Pro
         AND voided = true`,
     [txId],
   );
+};
+
+/**
+ * Get token symbol map, correlating token id to its symbol.
+ *
+ * @param mysql - Database connection
+ * @param tokenIdList - A list of token id
+ * @returns The token information (or null if id is not found)
+ *
+ * TODO: This method is duplicated from the wallet-service lambdas,
+ * we should have common methods for both packages
+ */
+export const getTokenSymbols = async (
+  mysql: MysqlConnection,
+  tokenIdList: string[],
+): Promise<StringMap<string> | null> => {
+  if (tokenIdList.length === 0) return null;
+
+  const [results] = await mysql.query<TokenSymbolsRow[]>(
+    'SELECT `id`, `symbol` FROM `token` WHERE `id` IN (?)',
+    [tokenIdList],
+  );
+
+  if (results.length === 0) return null;
+  return results.reduce((prev: Record<string, string>, token: { id: string, symbol: string }) => {
+    // eslint-disable-next-line no-param-reassign
+    prev[token.id] = token.symbol;
+    return prev;
+  }, {}) as unknown as StringMap<string>;
 };
--- a/packages/daemon/src/db/index.ts
+++ b/packages/daemon/src/db/index.ts
@@ -1538,7 +1538,7 @@ export const cleanupVoidedTx = async (mysql: MysqlConnection, txId: string): Pro
  * Get token symbol map, correlating token id to its symbol.
  *
  * @param mysql - Database connection
- * @param tokenIdList - A list of token id
+ * @param tokenIdList - A list of token ids
  * @returns The token information (or null if id is not found)
  *
  * @todo This method is duplicated from the wallet-service lambdas,
--- a/packages/daemon/src/db/index.ts
+++ b/packages/daemon/src/db/index.ts
@@ -4,7 +4,7 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  */
-import mysql, { Connection as MysqlConnection, Pool, RowDataPacket } from 'mysql2/promise';
+import mysql, { Connection as MysqlConnection, Pool } from 'mysql2/promise';
 import {
   TokenBalanceMap,
   DbTxOutput,
@@ -1541,7 +1541,7 @@ export const cleanupVoidedTx = async (mysql: MysqlConnection, txId: string): Pro
  * @param tokenIdList - A list of token id
  * @returns The token information (or null if id is not found)
  *
- * TODO: This method is duplicated from the wallet-service lambdas,
+ * @todo This method is duplicated from the wallet-service lambdas,
  * we should have common methods for both packages
  */
 export const getTokenSymbols = async (
--- a/packages/daemon/src/services/index.ts
+++ b/packages/daemon/src/services/index.ts
@@ -60,8 +60,7 @@ import {
 } from '../db';
 import getConfig from '../config';
 import logger from '../logger';
-import { SendMessageCommand, SQSClient } from '@aws-sdk/client-sqs';
-import { invokeOnTxPushNotificationRequestedLambda } from '../utils/aws';
+import { invokeOnTxPushNotificationRequestedLambda, sendMessageSQS } from '../utils/aws';
 
 export const METADATA_DIFF_EVENT_TYPES = {
   IGNORE: 'IGNORE',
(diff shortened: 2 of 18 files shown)


packages/daemon/src/db/index.ts
```diff
@@ -1532,3 +1533,32 @@ export const cleanupVoidedTx = async (mysql: MysqlConnection, txId: string): Pro
...
+
+/**
+ * Get token symbol map, correlating token id to its symbol.
+ *
+ * @param mysql - Database connection
+ * @param tokenIdList - A list of token id
+ * @returns The token information (or null if id is not found)
+ *
+ * TODO: This method is duplicated from the wallet-service lambdas,
+ * we should have common methods for both packages
```
r4mmer: suggestion(non-blocking): Should we use [@todo](https://jsdoc.app/tags-todo) since this is a js doc?

Maybe something like `@todo duplicate_code This method...` This way we can later search for `@todo duplicate_code` and find all instances of duplicated code we want to extract to a common method.

packages/daemon/src/services/index.ts
```diff
@@ -285,6 +292,52 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
...
+        const queueUrl = NEW_TX_SQS;
+        const client = new SQSClient({});
+        const command = new SendMessageCommand({
+          QueueUrl: queueUrl,
+          MessageBody: JSON.stringify({
+            wallets: Array.from(seenWallets),
+            tx,
+          }),
+        });
+
```
r4mmer: suggestion(non-blocking): Although it is simple, maybe we should extract this as a method `sendMessageSQS`?

packages/daemon/src/utils/aws.ts
```diff
@@ -0,0 +1,53 @@
...
+  const {
+    PUSH_NOTIFICATION_ENABLED,
+    WALLET_SERVICE_LAMBDA_ENDPOINT,
+    ON_TX_PUSH_NOTIFICATION_REQUESTED_FUNCTION_NAME,
+  } = getConfig();
+
+  if (PUSH_NOTIFICATION_ENABLED) {
+    logger.debug('Push notification is disabled. Skipping invocation of OnTxPushNotificationRequestedLambda lambda.');
+    return;
+  }
```
r4mmer:
```suggestion
  if (!PUSH_NOTIFICATION_ENABLED) {
    logger.debug('Push notification is disabled. Skipping invocation of OnTxPushNotificationRequestedLambda lambda.');
    return;
  }
```

packages/daemon/src/utils/helpers.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+ import { StringMap } from '../types';
+
+export const stringMapIterator = (stringMap: StringMap<unknown>): [string, unknown][] => (Object.entries(stringMap));
```
r4mmer:
```suggestion
export function stringMapIterator<T>(stringMap: StringMap<T>): [string, T][] => (Object.entries(stringMap));
```

r4mmer (changes requested):

packages/daemon/src/db/index.ts
```diff
@@ -1532,3 +1533,32 @@ export const cleanupVoidedTx = async (mysql: MysqlConnection, txId: string): Pro
     [txId],
   );
 };
+
+/**
+ * Get token symbol map, correlating token id to its symbol.
+ *
+ * @param mysql - Database connection
+ * @param tokenIdList - A list of token id
```
tuliomir: nitpick: Use the plural: "A list of token ids"

packages/daemon/src/types/transaction.ts
```diff
@@ -424,17 +424,16 @@ export interface Transaction {
...
-  signal_bits: number;
   version: number;
   weight: number;
   parents: string[];
   inputs: TxInput[];
   outputs: TxOutput[];
   height?: number;
   // eslint-disable-next-line camelcase
-  token_name?: string;
+  token_name?: string | null;
```
tuliomir, line 434: If this property is always `string` or `null`, do we still have to also inform it is _optional_?
In other words, can it be `undefined` too?

If not, we can just remove the `?` and assume it will always be present, even if `null`

packages/daemon/src/utils/alerting.ts
```diff
@@ -0,0 +1,61 @@
...
+import { Severity } from '../types';
+import getConfig from '../config';
+import logger from '../logger';
+
+/**
+ * Adds a message to the SQS alerting queue
+ *
+ * @param fnName - The lambda function name
+ * @param payload - The payload to be sent
+ */
```
tuliomir: fix: These jsdocs don't match the function parameters

packages/daemon/src/utils/alerting.ts
```diff
@@ -0,0 +1,61 @@
...
+export const addAlert = async (
+  title: string,
+  message: string,
+  severity: Severity,
```
tuliomir: question: Should we have a default severity?
Or maybe a protection against empty or invalid calls to this function?

packages/daemon/src/utils/aws.ts
```diff
@@ -0,0 +1,53 @@
...
+  const command = new InvokeCommand({
+    FunctionName: ON_TX_PUSH_NOTIFICATION_REQUESTED_FUNCTION_NAME,
+    InvocationType: 'Event',
+    Payload: JSON.stringify(walletBalanceValueMap),
+  });
+
+  const response: InvokeCommandOutput = await client.send(command);
+
+  // Event InvocationType returns 202 for a successful invokation
+  const walletIdList = Object.keys(walletBalanceValueMap);
```
tuliomir: suggestion: I think this command can be moved to inside the `if` context. It doesn't need to be called if the response status is `202`

packages/daemon/src/db/index.ts
andreabadesso: Done in 0f450683a5cdefb48b5e5063524ce07952f187f1

packages/daemon/src/services/index.ts
andreabadesso: I agree, done in 52d1cdad79d93803d653f1cd0ce252b915635da8

packages/daemon/src/utils/aws.ts
andreabadesso: This is weird, but fixed in d8a39670dc0bf2fc0293eecc2bc3e7aa57721316

packages/daemon/src/utils/helpers.ts
andreabadesso: Thanks, done in 2f912541f296b7c34ce3555b76340517444338f6

packages/daemon/src/db/index.ts
andreabadesso: b95be51dbba61db95cafe175504c7e6627ab3cdb

packages/daemon/src/types/transaction.ts
andreabadesso: It might also be undefined

packages/daemon/src/utils/alerting.ts
andreabadesso: Whoops, fixed in efd0bac5474e636e4a24fa7ca9f0d7eac6dc4a7b

packages/daemon/src/utils/alerting.ts
andreabadesso: I think we can have a default severity, defaulting to INFO in efd0bac5474e636e4a24fa7ca9f0d7eac6dc4a7b

packages/daemon/src/utils/aws.ts
andreabadesso: You're right, fixed in dad1fc21ad3bed1951e4dc73d073a2f08f051d9e

tuliomir (approved):

packages/daemon/src/utils/alerting.ts
```diff
@@ -0,0 +1,63 @@
...
+  const command = new SendMessageCommand({
+    QueueUrl: QUEUE_URL,
+    MessageBody: JSON.stringify(preparedMessage),
+    MessageAttributes: {
+      None: {
+        DataType: 'String',
+        StringValue: '--',
+      },
+    },
+  });
```
r4mmer: We should use the `sendMessageSQS` util here

packages/daemon/src/utils/wallet.ts
```diff
@@ -393,3 +399,114 @@ export const validateAddressBalances = async (mysql: MysqlConnection, addresses:
...
+  const addressWalletMap: StringMap<Wallet> = await getAddressWalletInfo(mysql, Object.keys(addressBalanceMap));
+
+  // Create a new map focused on the walletId and storing its balance variation from this tx
+  const walletsMap: StringMap<WalletBalance> = {};
+
+  // Accumulation of tokenId to be used to extract its symbols.
+  const tokenIdAccumulation = [];
+
+  // Iterates all the addresses to populate the map's data
+  const addressWalletEntries = stringMapIterator(addressWalletMap) as [string, Wallet][];
```
r4mmer: We don't need to cast `as [string, Wallet][]` anymore since `stringMapIterator` uses generic typing the `addressWalletEntries` will be `[string, Wallet][]`.

image

andreabadesso: Done in 08356ae607650478185b1e1d1996df5ff6e2c15c

packages/daemon/src/utils/alerting.ts
andreabadesso: Done in 98d774c4a101324e126692a93b6c8dfebc88f275

r4mmer (approved):

PR #105, opened 2023-12-19
feat: ping/pong mechanism

andreabadesso:
### Acceptance Criteria
- We should add a ping mechanism on the client side to detect if the connection was broken and the client is unaware of it

### Security Checklist

### DIFF
--- a/packages/daemon/src/actors/WebSocketActor.ts
+++ b/packages/daemon/src/actors/WebSocketActor.ts
@@ -11,10 +11,29 @@ import { get } from 'lodash';
 import logger from '../logger';
 import getConfig from '../config';
 
+const PING_TIMEOUT = 30000; // 30s timeout
+const PING_INTERVAL = 5000; // Will ping every 5s
+
 export default (callback: any, receive: any) => {
   const { WS_URL } = getConfig();
+  const createPingTimeout = (): NodeJS.Timeout => setTimeout(() => {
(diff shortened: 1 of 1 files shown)


packages/daemon/src/actors/WebSocketActor.ts
```diff
@@ -11,10 +11,29 @@ import { get } from 'lodash';
...
+
   // @ts-ignore: We already check for missing envs in startup
   const socket: WebSocket = new WebSocket(WS_URL);
+  let pingTimeout: NodeJS.Timeout = createPingTimeout();
+  let pingTimer: NodeJS.Timer;
+
+  const heartbeat = () => {
+    logger.debug('Pong received from server');
+    clearTimeout(pingTimeout);
+    pingTimeout = createPingTimeout();
```
glevco, line 35: Shouldn't this new `pingTimeout` instance be stored somewhere? It looks like we will be calling `clearTimeout(pingTimeout)` with the original instance from line 30 every time

luislhl (approved):

packages/daemon/src/actors/WebSocketActor.ts
andreabadesso: `setTimeout`, different from `setInterval`, runs only once

So what we're doing here is scheduling the next check

glevco (approved):

PR #110, opened 2024-01-16
feat: healthcheck API

luislhl:
Design: https://github.com/HathorNetwork/internal-rfcs/pull/5

### Acceptance Criteria
We should have a `/health` endpoint on the API that will:

- Be protected by an api key
- Be called every 60s by our Healthcheck Dashboard
- Use the `hathor-healthcheck-lib` to build a healthcheck response about the wallet-service health. The things to be checked include:
    - A comparison between the highest block height in MySQL and in the fullnode. If they differ, the service is reported as unhealthy. Doing this will also make sure the database and the fullnode are responding.
    - A check on the `/health` endpoint of the fullnode, to make sure it's reporting as healthy (internally hathor-core checks whether it's synced with other fullnodes to respond this).
    - A ping on Redis to make sure it's accessible and responding

### Security Checklist


packages/wallet-service/src/api/healthcheck.ts
```diff
@@ -0,0 +1,143 @@
...
+const mysql = getDbConnection();
+
+const checkDatabaseHeight: HealthcheckCallbackResponse = async () => {
+    try {
+        const [currentHeight, fullnodeStatus] = await Promise.all([
+            getLatestHeight(mysql),
+            fullnode.getStatus()
+        ]);
+
+        const currentFullnodeHeight = fullnodeStatus["dag"]["best_block"]["height"];
```
andreabadesso: s/"/'/g

```suggestion
        const currentFullnodeHeight = fullnodeStatus['dag']['best_block']['height'];
```

packages/wallet-service/src/api/healthcheck.ts
```diff
@@ -0,0 +1,143 @@
...
+
+        if (currentHeight === currentFullnodeHeight) {
```
andreabadesso: Maybe we should have a threshold here, we might still be syncing

andreabadesso (changes requested):

packages/wallet-service/src/api/healthcheck.ts
luislhl: Do we have an automatted formatter for the project? I think it would be the best solution to keep a style standard.

packages/wallet-service/src/api/healthcheck.ts
luislhl: Done in https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon/pull/110/commits/8fb42a40c15416a0c369af8e99286b5f303314c7

packages/wallet-service/src/api/healthcheck.ts
luislhl: Fixed in https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon/pull/110/commits/555184bdbaf034a17b2337e868ef0dc5476fd4d9

packages/wallet-service/src/api/healthcheck.ts
```diff
@@ -0,0 +1,145 @@
...
+            getLatestHeight(mysql),
+            fullnode.getStatus()
+        ]);
+
+        const currentFullnodeHeight = fullnodeStatus['dag']['best_block']['height'];
+
+        if (currentFullnodeHeight - currentHeight < HEALTHCHECK_MAXIMUM_HEIGHT_DIFFERENCE) {
+            return new HealthcheckCallbackResponse({
+                status: HealthcheckStatus.PASS,
+                output: `Database and fullnode heaights are within ${HEALTHCHECK_MAXIMUM_HEIGHT_DIFFERENCE} blocks difference`,
```
andreabadesso:
```suggestion
                output: `Database and fullnode heights are within ${HEALTHCHECK_MAXIMUM_HEIGHT_DIFFERENCE} blocks difference`,
```

andreabadesso (approved):

packages/wallet-service/src/redis.ts
```diff
@@ -22,6 +22,13 @@ export const closeRedisClient = (
   return quit();
 };
 
+export const ping = (
+  client: redis.RedisClient,
+): Promise<string> => {
```
pedroferreira1, line 31: What's the return of this method?

packages/wallet-service/src/api/healthcheck.ts
```diff
@@ -0,0 +1,145 @@
...
+} from '@hathor/healthcheck-lib';
+import { getLatestHeight } from '@src/db';
+import fullnode from '@src/fullnode';
+import { closeDbConnection, getDbConnection } from '@src/utils';
+import { APIGatewayProxyHandler } from 'aws-lambda';
+import { getRedisClient, ping } from '@src/redis';
+
+const mysql = getDbConnection();
+
+const HEALTHCHECK_MAXIMUM_HEIGHT_DIFFERENCE = Number(process.env.HEALTHCHECK_MAXIMUM_HEIGHT_DIFFERENCE) || 5;
```
pedroferreira1: Is there any chance that we want to use `process.env.HEALTHCHECK_MAXIMUM_HEIGHT_DIFFERENCE = '0'` someday?

In this case this would be `5`, not `0`.

luislhl: You're right. I think we should allow setting as 0, this is a bug. I'll fix this.

packages/wallet-service/src/redis.ts
luislhl: 'PONG'. See https://redis.io/commands/ping/

packages/wallet-service/src/api/healthcheck.ts
luislhl: Fixed in https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon/pull/110/commits/e078a4f4f7e3ff8114b37e9bfce8cad6ccac78d5

pedroferreira1 (approved):

packages/wallet-service/src/api/healthcheck.ts
```diff
@@ -0,0 +1,145 @@
...
+      });
+    } else if (health['status'] === HealthcheckStatus.WARN) {
+      return new HealthcheckCallbackResponse({
+        status: HealthcheckStatus.WARN,
+        output: `Fullnode has health warnings: ${health}`,
+      });
+    } else {
+      return new HealthcheckCallbackResponse({
+        status: HealthcheckStatus.FAIL,
+        output: `Fullnode is unhealthy: ${JSON.stringify(health)}`,
```
luislhl, line 93: We should review whether we should fail our healthcheck when the fullnode has just not received blocks for a while (which is a possible situation)

PR #114, opened 2024-02-14
fix: broken deploy

andreabadesso:
### Acceptance Criteria
- We should fix the deploy script on github actions as it had a typo
- We should fix the docker image build, it was using the `node_modules` folder from the host instead of the locally installed
- We should update nodejs to v20

### Security Checklist


packages/daemon/.dockerignore
```diff
@@ -0,0 +1,4 @@
+dist/
+node_modules/
+__tests__/
+node_modules/
```
luislhl: I believe this has no effect, but this one is duplicated.

packages/daemon/Dockerfile
```diff
@@ -15,22 +15,29 @@ COPY package.json ./
...
 RUN which python
 
-RUN yarn set version berry
+RUN yarn set version 3.4.1
 
 RUN yarn install
 
 COPY . ./
 
+RUN ls -la /app
```
luislhl: Should we keep this?

packages/daemon/.dockerignore
andreabadesso: Fixed, thanks!

packages/daemon/Dockerfile
andreabadesso: We shouldn't, fixed!

luislhl (approved):

Dockerfile
```diff
@@ -0,0 +1,34 @@
+# Copyright 2020 Hathor Labs
```
luislhl: Since you're adding this, should we use 2024?

packages/daemon/.dockerignore
luislhl, line 1: Is this file still needed, since there is one in the root folder, and the docker build is made there now?

luislhl (approved):

packages/daemon/.dockerignore
andreabadesso: It is not! Removed

Dockerfile
andreabadesso: Sure! Updated

.gitignore
```diff
@@ -12,3 +12,4 @@ packages/wallet-service/.serverless
 packages/wallet-service/.webpack
 packages/wallet-service/.env*
 .yarn/
+.git
```
tuliomir: Question: I'm unsure if this has the expected behavior. The `.git` folder is always ignored by default, so why should we add it here?

If we want to ignore nested `.git` folders, will this command work? Or should we use `**/.git`?

andreabadesso: This wasn't supposed to be there, fixed!

Thanks

tuliomir (approved):

PR #115, opened 2024-02-16
fix: use fullnode network

andreabadesso:
### Acceptance Criteria
- The network that comes from the fullnode events is the official name of the network, e.g. `testnet-golf`, we should compare it to a different env varialbe other than the one we use to derive addresses 

### Security Checklist


packages/daemon/src/config.ts
```diff
@@ -49,6 +50,7 @@ export const FULLNODE_PEER_ID = process.env.FULLNODE_PEER_ID;
 export const FULLNODE_HOST = process.env.FULLNODE_HOST;
 export const STREAM_ID = process.env.STREAM_ID;
 export const NETWORK = process.env.NETWORK;
+export const FULLNODE_NETWORK = process.env.FULLNODE_NETWORK;
```
luislhl, line 57: Maybe we should add a comment explaining the reason why we needed 2 env vars (NETWORK and FULLNODE_NETWORK)

luislhl (approved):

packages/daemon/src/config.ts
andreabadesso: Done! Thanks

tuliomir (approved):

PR #120, opened 2024-02-19
chore: bump v1.4.3

andreabadesso:
### Acceptance Criteria
- We should bump the version to `v1.4.3+beta`
- We should have a single version for both projects from now on, so the version should be removed on the specific packages

### Security Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "name": "hathor-wallet-service",
+  "version": "1.4.3-beta",
   "workspaces": [
     "packages/daemon",
     "packages/wallet-service"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hathor-wallet-service",
-  "version": "1.4.3-beta",
+  "version": "1.4.3+beta",
   "workspaces": [
     "packages/daemon",
     "packages/wallet-service"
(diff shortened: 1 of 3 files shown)


luislhl (approved):

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "hathor-wallet-service",
+  "version": "1.4.3+beta",
```
tuliomir, line 3: question: Do we really use `+` on the beta versions? Or should we have `1.4.3-beta`?

tuliomir (approved):

package.json
andreabadesso: It's because in the release guide we use `-rc` for release candidate versions so we agreed upon using `+` for the beta, e.g. `v1.4.3-rc1+beta`

luislhl: We really use it. Mostly because we don't want it to be considered as a pre-release (as in https://semver.org/#spec-item-9), but just as metadata (as in https://semver.org/#spec-item-10).

This makes it easier for some tools in our pipeline to detect the release-candidates, which are our actual pre-releases, and not mistake releases with pre-releases because of the `-beta`. Otherwise the `-beta` could mess with version precedence.

tuliomir: Thanks for the reference to the semver docs! It's pretty clear to me now why the `+`

PR #130, opened 2024-02-22
chore: single deploy process for both projects in the monorepo

andreabadesso:
### Acceptance Criteria
- We should merge the deploy scripts from both projects to be in codebuild
- If the build for any of the two projects fail, we shouldn't build the other one

### Security Checklist


Makefile
```diff
@@ -2,21 +2,61 @@
...
+
+# mainnet
 
 .PHONY: deploy-lambdas-mainnet
 deploy-lambdas-mainnet:
-	npx serverless deploy --stage mainnet --region eu-central-1
+	yarn workspace wallet-service run serverless deploy --stage mainnet --region eu-central-1
+
+.PHONY: push-daemon-mainnet
+push-daemon-mainnet:
```
luislhl: We are missing `build-daemon-mainnet`, right?

Makefile
```diff
@@ -2,21 +2,61 @@
...
+build-daemon-dev-testnet:
+	bash scripts/build-daemon.sh dev-testnet
+
+.PHONY: push-daemon-dev-testnet
+push-daemon-dev-testnet:
+	bash scripts/push-daemon.sh
+
 .PHONY: deploy-lambdas-dev-testnet
 deploy-lambdas-dev-testnet:
 	yarn workspace wallet-service run serverless deploy --stage dev-testnet --region eu-central-1
```
luislhl: At some point, I believe we had renamed  this one into `ci-deploy-lambdas-dev-testnet`, and created a new one like this:

```
.PHONY: deploy-lambdas-dev-testnet
deploy-lambdas-dev-testnet:
	   AWS_SDK_LOAD_CONFIG=1 yarn workspace wallet-service run serverless deploy --stage dev-testnet --region eu-central-1
```

Without the env var, the deploy doesn't work from a local machine using `aws sso` to authenticate, right?

Should we do this again?

scripts/build-daemon.sh
```diff
@@ -0,0 +1,25 @@
...
+fi
+
+if [ -z "$ACCOUNT_ID" ]; then
+    echo "Please export a ACCOUNT_ID env var before running this";
+    exit 1;
+fi
+
+commit=`git rev-parse HEAD`;
+timestamp=`date +%s`;
+export DOCKER_IMAGE_TAG="$1-$commit-$timestamp";
```
luislhl: Maybe you should use `$STAGE` here, instead of `$1` directly, for readability.

scripts/push-daemon.sh
```diff
@@ -0,0 +1,10 @@
+set -e
+set -o pipefail
+
+DOCKER_IMAGE_TAG=$(cat /tmp/docker_image_tag)
```
luislhl, line 4: Who will write to `/tmp/docker_image_tag`?

scripts/build-daemon.sh
luislhl: There is an important difference between what you're doing, and the way we did it in the past (in the `.github/workflows/deploy.yml` file that was deleted).

The difference is that the Git version tags won't be used as Docker tags anymore.

I believe we should keep the old behavior. Git tags should be used as Docker tags, because it's easier to identify the images later.

But regardless, we need to make sure the same patterns used to build the images here will be used by the rest of our automation tools to detect and deploy the images.

Makefile
andreabadesso: Fixed, thanks

Makefile
andreabadesso: We should, done!

Thanks

scripts/build-daemon.sh
andreabadesso: Fixed, thanks

andreabadesso: I agree.

Added the `DOCKER_IMAGE_TAG` on the build.sh and using it on the codebuild script

scripts/build-daemon.sh
```diff
@@ -0,0 +1,29 @@
+set -e
+set -o pipefail
+
+STAGE=$1
+
+if [ -z $STAGE ]; then
```
luislhl: I think that we don't need this variable anymore, right?

scripts/push-daemon.sh
luislhl, line 4: Maybe we should check whether this is not empty, just in case, and return error if it is.

Someone could try to run locally without running the build first, for instance.
Or someone could mess with the build and break the right order of execution of scripts without noticing.

Makefile
```diff
@@ -2,21 +2,61 @@
...
 .PHONY: deploy-lambdas-testnet
 deploy-lambdas-testnet:
-	yarn workspace wallet-service run serverless deploy --stage testnet --region eu-central-1
+	AWS_SDK_LOAD_CONFIG=1 yarn workspace wallet-service run serverless deploy --stage testnet --region eu-central-1
+
+# mainnet-staging
+
+.PHONY: build-daemon-mainnet-staging
+build-daemon-mainnet-staging:
+	bash scripts/build-daemon.sh mainnet_staging
```
luislhl: Do we still need to pass this argument to the script?

Also, if not, do we still need a build and a push commands in the Makefile for each environment?

luislhl (approved):

luislhl (approved):

scripts/build-daemon.sh
tuliomir, line 1: suggestion(non-blocking): I think adding some comments here would be nice to clarify what each step is making on this script

tuliomir (approved):

PR #143, opened 2024-03-11
chore: updated wallet-service docs to root

andreabadesso:
### Motivation

Moved wallet-service docs to the root folder of the monorepo and updated it with the new changes done after the repositories were merged

### Acceptance Criteria

- We should move the docs to the root dir
- We should update docs to reflect the new monorepo reality
- We should update the feature toggles to reflect the ones created in Unleash

### Checklist


luislhl (approved):

docs/2022-04-18-feature-toggles.md
```diff
@@ -10,18 +10,13 @@ This document describes the main features we are using and how to interact with
 
 ## Feature toggles
 
-We have a list of [feature toggles](https://docs.getunleash.io/advanced/feature_toggle_types) that are queried by the wallets:
+We have two [feature toggles](https://docs.getunleash.io/advanced/feature_toggle_types) that are queried by the wallets:
```
r4mmer, line 13: Unrelated to PR.
In unleash we have a single project with all feature toggles, should we create a separate project for wallet-service and if we require more feature toggles we can add them there without having to update these docs?

r4mmer (approved):

docs/2022-04-18-feature-toggles.md
andreabadesso: It would be great, but we're running the community version of Unleash and creating multiple projects are only available in the "Pro & Enterprise" version:

image

But that's really valid point, as we're currently receiving feature toggles for all projects, it's not that bad atm because we have only a few feature toggles, but it's wasted bandwidth

I'll open an issue to study if we would benefit from upgrading and the pricing

PR #145, opened 2024-03-18
chore: increased memory and lambda timeout

andreabadesso:
### Motivation

Wallets with a large number of transactions cause the `loadWalletAsync` lambda to run out of memory, this is yet to be investigated, so this is a short-term solution until we fully investigate and solve the increased use of memory

1gb of memory is a safe margin on the average use of memory used while loading a wallet with 380k transactions.

### Acceptance Criteria

- We should increase the memory limit of the `loadWalletAsync` lambda to 1gb
- We should increase the `loadWalletAsync` lambda timeout to 10 minutes

### Checklist

### DIFF
--- a/packages/wallet-service/serverless.yml
+++ b/packages/wallet-service/serverless.yml
@@ -242,6 +242,11 @@ functions:
           arn:aws:lambda:eu-central-1:${self:provider.environment.ACCOUNT_ID}:function:hathor-explorer-service-${self:custom.explorerServiceStage}-create_or_update_dag_metadata
   loadWalletAsync:
     handler: src/api/wallet.loadWallet
+    # This lambda is currently running out of memory when wallets with a big
+    # number of transactions (> 300k) is being loaded. I chose 1024 because the
+    # average memory usage of the largest test wallet we have is 800mb.
+    memorySize: 1024
+    timeout: 600 # 10 minutes should be enough for most wallets
     warmup:
       walletWarmer:
         enabled: false


pedroferreira1 (approved):

packages/wallet-service/serverless.yml
```diff
@@ -242,6 +242,11 @@ functions:
           arn:aws:lambda:eu-central-1:${self:provider.environment.ACCOUNT_ID}:function:hathor-explorer-service-${self:custom.explorerServiceStage}-create_or_update_dag_metadata
   loadWalletAsync:
     handler: src/api/wallet.loadWallet
+    # This lambda is currently running out of memory when wallets with a big
+    # number of transactions (> 300k) is being loaded. I chose 1024 because the
+    # average memory usage of the largest test wallet we have is 800mb.
+    memorySize: 1024
+    timeout: 600 # 10 minutes should be enough for most wallets
```
alexruzenhack, line 249: [suggestion] Can we use a cache to store all these data outside the lambda and yet make it available extremely fast? Maybe another product from AWS?

alexruzenhack (approved):

packages/wallet-service/serverless.yml
andreabadesso: The wallet-service is the cache, this method creates it

This lambda receives the wallet's `xpubkey` and uses it to relate all addresses and their transactions and utxos (which are already processed) to the newly created wallet

This process is what is using more memory than it should 

Without the `xpubkey` we can never know which address belongs to which wallet, so there is no way to do it before the user calls this lambda

PR #146, opened 2024-03-19
feat: added handler to handle DLQ messages from loadWalletAsync

andreabadesso:
### Motivation

Currently, if the `loadWalletAsync` lambda is terminated, we never set the wallet state to `error` in the database, causing the wallets to never fallback to the fullnode facade.

This PR will introduce a DLQ on the `loadWalletAsync` lambda and a handler to handle its messages

### Acceptance Criteria

- We should send a message to the DLQ SNS topic when the lambda fails to execute for whatever reason
- We should send an alert to our alerting SQS with metadata on what caused the fail
- We should fix the sqs client on both wallet-service and the daemon

### Checklist

### DIFF
--- a/packages/daemon/src/utils/aws.ts
+++ b/packages/daemon/src/utils/aws.ts
@@ -61,8 +61,17 @@ export const invokeOnTxPushNotificationRequestedLambda = async (walletBalanceVal
  * @param messageBody - A string with the message body
  * @param queueUrl - The queue URL
  */
-export const sendMessageSQS = async (messageBody: string, queueUrl: string, messageAttributes?: Record<string, MessageAttributeValue>): Promise<SendMessageCommandOutput> => {
-  const client = new SQSClient({});
+export const sendMessageSQS = async (messageBody: string, queueUrl: string, messageAttributes?: Record<string, MessageAttributeValue>, region?: string): Promise<SendMessageCommandOutput> => {
+  const { AWS_REGION } = getConfig();
+
+  if (!region) {
+    region = AWS_REGION;
+  }
+
+  const client = new SQSClient({
+    endpoint: queueUrl,
+    region,
+  });
   const command = new SendMessageCommand({
     QueueUrl: queueUrl,
     MessageBody: messageBody,
--- a/packages/wallet-service/serverless.yml
+++ b/packages/wallet-service/serverless.yml
@@ -120,6 +120,11 @@ resources:
       Properties:
         QueueName:
             WalletServiceNewTxQueue_${self:custom.stage}
+    WalletServiceLoadAsyncFailedTopic:
+        Type: "AWS::SNS::Topic"
+        Properties:
+          DisplayName: 'Messages published when the loadWalletAsync lambda fails'
+          TopicName: WalletServiceLoadAsyncFailed_${self:custom.stage}
 
 provider:
   name: aws
@@ -247,9 +252,33 @@ functions:
     # average memory usage of the largest test wallet we have is 800mb.
     memorySize: 1024
     timeout: 600 # 10 minutes should be enough for most wallets
+    onError: arn:aws:sns:eu-central-1:${self:provider.environment.ACCOUNT_ID}:WalletServiceLoadAsyncFailed_${self:custom.stage}
     warmup:
       walletWarmer:
         enabled: false
+    iamRoleStatements:
+      - Effect: Allow
+        Action:
+          - SNS:Publish
+        Resource:
+          arn:aws:sns:eu-central-1:${self:provider.environment.ACCOUNT_ID}:WalletServiceLoadAsyncFailed_${self:custom.stage}
+  handleLoadWalletFailed:
+    handler: src/api/wallet.loadWalletFailed
+    events:
+      - sns: arn:aws:sns:eu-central-1:${self:provider.environment.ACCOUNT_ID}:WalletServiceLoadAsyncFailed_${self:custom.stage}
+    warmup:
+      walletWarmer:
+        enabled: false
+    iamRoleStatementsInherit: true
+    iamRoleStatementsName: hathor-wallet-service-${self:custom.stage}-handleLoadFailed-snsRole
+    iamRoleStatements:
+      - Effect: Allow
+        Action:
+          - SNS:Publish
+          - SNS:Subscribe
+          - SNS:Unsubscribe
+        Resource:
+          arn:aws:sns:eu-central-1:${self:provider.environment.ACCOUNT_ID}:WalletServiceLoadAsyncFailed_${self:custom.stage}
   loadWalletApi:
     role: arn:aws:iam::${self:provider.environment.ACCOUNT_ID}:role/WalletServiceLoadWalletLambda
     handler: src/api/wallet.load
--- a/packages/wallet-service/serverless.yml
+++ b/packages/wallet-service/serverless.yml
@@ -36,10 +36,10 @@ custom:
     topics: # SNS Topics to send alerts to
       major:
         alarm:
-          topic: arn:aws:sns:eu-central-1:${self:provider.environment.ACCOUNT_ID}:opsgenie-cloudwatch-integration-production-major
+          topic: arn:aws:sns:${self:provider.region}:${self:provider.environment.ACCOUNT_ID}:opsgenie-cloudwatch-integration-production-major
       minor:
         alarm:
-          topic: arn:aws:sns:eu-central-1:${self:provider.environment.ACCOUNT_ID}:opsgenie-cloudwatch-integration-production-minor
+          topic: arn:aws:sns:${self:provider.region}:${self:provider.environment.ACCOUNT_ID}:opsgenie-cloudwatch-integration-production-minor
     definitions: # Definition of alarms
       majorFunctionErrors:
         description: "Too many errors in hathor-wallet-service. Runbook: https://github.com/HathorNetwork/ops-tools/blob/master/docs/runbooks/wallet-service/errors-in-logs.md"
@@ -129,6 +129,7 @@ resources:
 provider:
   name: aws
   runtime: nodejs18.x
+  region: ${opt:region, 'eu-central-1'}
   # In MB. This is the memory allocated for the Lambdas, they cannot use more than this
   # and will break if they try.
   memorySize: 256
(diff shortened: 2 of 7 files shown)


packages/daemon/src/utils/aws.ts
```diff
@@ -61,8 +61,17 @@ export const invokeOnTxPushNotificationRequestedLambda = async (walletBalanceVal
...
+  const { AWS_REGION } = getConfig();
+
+  if (!region) {
+    region = AWS_REGION;
+  }
+
+  const client = new SQSClient({
+    endpoint: queueUrl,
+    region,
+  });
```
andreabadesso, line 74: This is unrelated to the fix, but since I fixed this same issue in the wallet-service, I added it here as well

packages/wallet-service/serverless.yml
```diff
@@ -242,9 +247,33 @@ functions:
...
+      - sns: arn:aws:sns:eu-central-1:${self:provider.environment.ACCOUNT_ID}:WalletServiceLoadAsyncFailed_${self:custom.stage}
+    warmup:
+      walletWarmer:
+        enabled: false
+    iamRoleStatementsInherit: true
+    iamRoleStatementsName: hathor-wallet-service-${self:custom.stage}-handleLoadFailed-snsRole
+    iamRoleStatements:
+      - Effect: Allow
+        Action:
+          - SNS:Publish
```
alexruzenhack: [doubt] Will this lambda also publish something to the topic again?

packages/wallet-service/src/api/wallet.ts
```diff
@@ -380,6 +381,65 @@ interface LoadResult {
...
+    for (let i = 0; i < records.length; i++) {
+      const snsEvent = records[i].Sns;
+      const { RequestID, ErrorMessage } = snsEvent.MessageAttributes;
+
+      // Process each failed load wallet event
+      const loadEvent: LoadEvent = JSON.parse(snsEvent.Message) as unknown as LoadEvent;
+
+      if (!loadEvent.xpubkey) {
+        logger.error('Received wallet load fail message from SNS but no xpubkey received');
+        continue;
```
alexruzenhack: [doubt] What are the possible causes for this situation?

packages/wallet-service/src/api/wallet.ts
```diff
@@ -380,6 +381,65 @@ interface LoadResult {
...
+
+      const walletId = getWalletId(loadEvent.xpubkey);
+
+      // update wallet status to 'error' and set the number of retries to MAX so
+      // it doesn't get retried
+      await updateWalletStatus(mysql, walletId, WalletStatus.ERROR, MAX_LOAD_WALLET_RETRIES);
+
+      logger.error(`${walletId} failed to load.`);
+      logger.error({
+        xpubkey: loadEvent.xpubkey,
```
alexruzenhack: [doubt] Should we log the xpubkey? If this log falls on the hand of malicious actor, could it harm the user in any way?

alexruzenhack: I believe the `walletId` is sufficient to uniquely identify the wallet on the database, right?

packages/wallet-service/src/api/wallet.ts
```diff
@@ -380,6 +381,65 @@ interface LoadResult {
...
+        'A wallet failed to load in the wallet-service',
+        `The wallet with id ${walletId} failed to load on the wallet-service. Please check the logs.`,
+        Severity.MINOR,
+        {
+          xpubkey: loadEvent.xpubkey,
+          walletId,
+          RequestID: RequestID.Value,
+          ErrorMessage: ErrorMessage.Value,
+        },
+      );
```
alexruzenhack: [doubt] What is the purpose of this alert, to build awareness? What are the possible causes for this issue?

packages/wallet-service/src/api/wallet.ts
```diff
@@ -380,6 +381,65 @@ interface LoadResult {
...
+    }
+  } catch (e) {
+    await addAlert(
+      'Failed to handle loadWalletFailed event',
+      `Failed to process the loadWalletFailed event. Please check the logs.`,
+      // This is major because the user will be stuck in a loading cycle
```
alexruzenhack, line 446: [suggestion] I believe this explanation should be on the alert message as well to add context.

packages/wallet-service/src/api/wallet.ts
alexruzenhack: This possibility would be a problem on the lambda itself, right? Or an invoke on lambda without correct params, right?

alexruzenhack: Asking because if it is a possible problem on the lambda itself, it would be nice to alert us to investigate, because it may impact other calls to the lambda.

alexruzenhack: Or we should build awareness over this error logs, turn it more visible for us.

andreabadesso: I added this try/catch thinking on a potential database failure which would cause the wallet to never be set to `error`, requiring manual intervetion

But it might happen for other reasons, like a failed deploy with missing env variables

I updated the alert message, indicating that this is important. It is also a MAJOR alert, so the on-call should treat it with utmost importance

e9b1f6d03a56c145e170a74754afc7d9ab8f2b11


I will create an issue to write a runbook for this situation

packages/wallet-service/src/api/wallet.ts
andreabadesso: I don't think the user can be harmed, but if this is leaked, it will indeed be a privacy issue as it will reveal information about the wallet addresses

I agree the `walletId` should be enough, fixed in 9bb99e65b69daf2037241252016cd9a514ba1b07


Thanks

packages/wallet-service/src/api/wallet.ts
andreabadesso: I improved the log in e9b1f6d03a56c145e170a74754afc7d9ab8f2b11, can you please check?

packages/wallet-service/serverless.yml
andreabadesso: It won't! Removed the permission in 63ec7cb3576bc6768526db68bf05fa7444a39aea


Thanks

alexruzenhack (approved):

packages/wallet-service/serverless.yml
```diff
@@ -242,9 +247,32 @@ functions:
           arn:aws:lambda:eu-central-1:${self:provider.environment.ACCOUNT_ID}:function:hathor-explorer-service-${self:custom.explorerServiceStage}-create_or_update_dag_metadata
   loadWalletAsync:
     handler: src/api/wallet.loadWallet
+    onError: arn:aws:sns:eu-central-1:${self:provider.environment.ACCOUNT_ID}:WalletServiceLoadAsyncFailed_${self:custom.stage}
```
pedroferreira1: Can't we use the region from the env?

packages/wallet-service/src/api/wallet.ts
```diff
@@ -380,6 +381,63 @@ interface LoadResult {
   xpubkey: string;
 }
 
+/*
+ * This lambda will be started by a SNSMessage on the load failed SNS configured
+ * in serverless.yml. It will receive all wallet load failed events published on
+ * the loadWalletAsync DLQ SNS topic
```
pedroferreira1, line 387: It would be interesting to have here an example of the event received.

packages/wallet-service/src/api/wallet.ts
```diff
@@ -380,6 +381,63 @@ interface LoadResult {
...
+          walletId,
+          RequestID: RequestID.Value,
+          ErrorMessage: ErrorMessage.Value,
+        },
+      );
+    }
+  } catch (e) {
+    await addAlert(
+      'Failed to handle loadWalletFailed event',
+      `Failed to process the loadWalletFailed event. This indicates that wallets failed to load and we wasn't able to recover, please check the logs as soon as possible.`,
```
pedroferreira1:
```suggestion
      `Failed to process the loadWalletFailed event. This indicates that wallets failed to load and we weren't able to recover, please check the logs as soon as possible.`,
```

packages/wallet-service/src/api/wallet.ts
pedroferreira1: > I updated the alert message, indicating that this is important. It is also a MAJOR alert, so the on-call should treat it with utmost importance

You are not alerting anything here. You just continue to the next event record.

andreabadesso: @pedroferreira1 you're right, I added an alert in 78ba032aa169fa6018decded99e39e45651bb670

packages/wallet-service/src/api/wallet.ts
andreabadesso: Done in 33f81483e042dae9bc6a820a8eca52c8213a74b7

packages/wallet-service/src/api/wallet.ts
andreabadesso: Thanks 0ad0f176db0f755641a14d1b775e9d0c6a9a6118

packages/wallet-service/serverless.yml
andreabadesso: Not from the env, but we can get it from serverless:

dc53e307208de6753fe3cea7afbf9dbd54379767

packages/wallet-service/src/api/wallet.ts
```diff
@@ -380,6 +381,75 @@ interface LoadResult {
...
+      const { RequestID, ErrorMessage } = snsEvent.MessageAttributes;
+
+      // Process each failed load wallet event
+      const loadEvent: LoadEvent = JSON.parse(snsEvent.Message) as unknown as LoadEvent;
+
+      if (!loadEvent.xpubkey) {
+        logger.error('Received wallet load fail message from SNS but no xpubkey received');
+        await addAlert(
+          'Wallet failed to load, but no xpubkey received.',
+          `A event reached loadWalletFailed lambda but the xpubkey was not sent. This indicates that wallets failed to load and we weren't able to recover, please check the logs as soon as possible.`,
```
pedroferreira1:
```suggestion
          `An event reached loadWalletFailed lambda but the xpubkey was not sent. This indicates that the wallet has failed to load and we weren't able to recover, please check the logs as soon as possible.`,
```

pedroferreira1 (approved):

PR #150, opened 2024-03-28
feat: call onNewNftEvent when a new NFT tx is received

andreabadesso:
### Motivation

This fixes https://github.com/HathorNetwork/hathor-wallet-service/issues/153

After the wallet-service refactor, we're no longer calling the old `onNewTxRequest` lambda (as what it used to do is now done by the daemon), so the logic that deals with NFTs should be called from the daemon.

This PR is not attempting to refactor the wallet-service and move all shared code to the `common` package, but to move only the minimum necessary code so that the nft util works both on the wallet-service and the daemon package.

### Acceptance Criteria

- We should move the nft utils from the wallet-service repository to the `common` package, that has methods common to both the lambdas and the daemon
- We should move all transaction types to the `common` package
- We should call use the nft util methods from the `common` package to call the `onNewNftEvent` lambda in the `newVertexAccepted` service of the daemon

### Checklist


package.json
```diff
@@ -3,15 +3,16 @@
...
+    "packages/common"
   ],
   "engines": {
     "node": ">=18"
   },
   "nohoist": [
     "**"
   ],
-  "repository": "git@github.com:HathorNetwork/hathor-wallet-service-sync_daemon.git",
+  "repository": "git@github.com:HathorNetwork/hathor-wallet-service.git",
```
andreabadesso: I renamed the repository

packages/common/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "@wallet-service/common",
+  "packageManager": "yarn@4.1.0",
+  "peerDependencies": {
+    "@aws-sdk/client-lambda": "3.540.0",
+    "@hathor/wallet-lib": "0.39.0",
+    "winston": "^3.13.0"
```
andreabadesso: These are defined as `peerDependencies` so they use the same version from the root `package.json`

packages/daemon/src/types/event.ts
```diff
@@ -75,6 +75,7 @@ export type FullNodeEvent = {
       tokens: string[];
       token_name: null | string;
       token_symbol: null | string;
+      signal_bits: number;
```
andreabadesso, line 78: We should open a PR to start storing this in the transactions table, but this is not yet coming from the event stream.

packages/daemon/src/services/index.ts
```diff
@@ -343,9 +347,23 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
...
+      // Validating for NFTs only after the tx is successfully added
+      if (NftUtils.shouldInvokeNftHandlerForTx(tx, network, logger)) {
+        // This process is not critical, so we run it in a fire-and-forget manner, not waiting for the promise.
+        // In case of errors, just log the asynchronous exception and take no action on it.
+        NftUtils.invokeNftHandlerLambda(tx.tx_id, STAGE)
+          .catch((err) => logger.error('[ALERT] Errored on nftHandlerLambda invocation', err));
+      }
     }
 
-    // TODO: Send message on SQS  for real-time update
```
andreabadesso, line 348: We already did this in the past, it's on line 327

packages/common/__tests__/utils/nft.utils.test.ts
```diff
@@ -128,7 +152,8 @@ describe('isTransactionNFTCreation', () => {
 
     // Validation
     const tx = getTransaction();
-    const result = NftUtils.isTransactionNFTCreation(tx);
+    // @ts-ignore
```
tuliomir: question: Is this flag still needed?

packages/common/package.json
```diff
@@ -1,12 +1,20 @@
 {
   "name": "@wallet-service/common",
   "packageManager": "yarn@4.1.0",
+  "scripts": {
+    "jest": "jest --runInBand --collectCoverage --detectOpenHandles --forceExit",
+    "test": "jest"
```
tuliomir: question(non-blocking): Do we have any benefit of having this `jest` script instead of just a `test` one?

packages/common/src/types.ts
tuliomir, line 1: We should have the Hathor header on this file.

Also, this is a types file that is half filled with _utils_ functions. Maybe this is not the time to extract all those functions into separate dedicated files, but I believe at least an issue would be important to do so. In the future we may have circular dependencies because of these mixed responsibilities.

packages/common/__tests__/utils/nft.utils.test.ts
andreabadesso: It's not, removed it in c92fafac7661553185d9992a25f409169d59b4e2

packages/common/package.json
andreabadesso: We don't... Removed the jest script in 13f98b61cca8c4b3f6f5af88dacc189fe17832c9

packages/common/src/types.ts
andreabadesso: Adde the header in 61fe30e710758bce5e9ca74fc6b41a54990022c6


I'm writing a project listing all refactors I think we need to do to share code between projects, I'll add this there

tuliomir (approved):

packages/common/__tests__/events/nftCreationTx.ts
```diff
@@ -0,0 +1,181 @@
+/**
```
pedroferreira1: Should we delete the file in the wallet service package that is exactly this one?

https://github.com/HathorNetwork/hathor-wallet-service/blob/master/packages/wallet-service/events/nftCreationTx.ts

packages/wallet-service/src/txProcessor.ts
```diff
@@ -81,6 +81,8 @@ export const IGNORE_TXS = {
   ],
 };
 
+export const CREATE_NFT_MAX_RETRIES = 5;
```
pedroferreira1: We used to have an env var for this, why did you change it?

```
export const MAX_METADATA_UPDATE_RETRIES: number = parseInt(process.env.MAX_METADATA_UPDATE_RETRIES || '3', 10);
```

packages/common/src/utils/nft.utils.ts
```diff
@@ -26,24 +23,30 @@ export class NftUtils {
   /**
    * Returns whether we should invoke our NFT handler for this tx
    * @param {Transaction} tx
+   * @param {string} network
```
pedroferreira1: Missing `logger` parameter and wrong `network` type. Since it's a typescript file I think we could omit the types in the docstring.

packages/common/src/utils/nft.utils.ts
```diff
@@ -124,29 +127,31 @@ export class NftUtils {
   /**
    * Identifies if the metadata for a NFT needs updating and, if it does, update it.
    * @param {string} nftUid
+   * @param {number} maxRetries
```
pedroferreira1: Missing `logger` parameter

packages/wallet-service/src/txProcessor.ts
andreabadesso: Refactored it in 4448e29f567e858ebbc5950d461e5ed366ce0289

We were not using this env, it was always defaulting to 3, added it to codebuild as well

packages/common/src/utils/nft.utils.ts
andreabadesso: Fixed in 16efa8fdcf591e6e3c8b74a7e8e9e7bcb2f2d9e2

packages/common/src/utils/nft.utils.ts
andreabadesso: Fixed in 16efa8fdcf591e6e3c8b74a7e8e9e7bcb2f2d9e2

packages/common/__tests__/events/nftCreationTx.ts
andreabadesso: Yes, removed in 25bb3a319d0606b08076e033720f6b493e2d56b0

pedroferreira1 (approved):

PR #151, opened 2024-04-01
refactor: added a "common" package to share code between packages

andreabadesso:
### Motivation

We should have a `common` package so that the wallet-service and the daemon can share code between them.

This PR does not attempt to refactor both projects, but only to create the common project and include the bare minimum changes to enable code sharing of the nft util

### Acceptance Criteria

- We should migrate the `addAlert` from the daemon's package to the common package and use it on both the wallet-service and the daemon
- We should migrate types related to transactions and use them in the daemon

### Checklist


package.json
```diff
@@ -3,22 +3,40 @@
...
+  "dependencies": {
+    "@aws-sdk/client-apigatewaymanagementapi": "3.540.0",
+    "@aws-sdk/client-lambda": "3.540.0",
+    "@aws-sdk/client-sqs": "3.540.0",
+    "@hathor/wallet-lib": "0.39.0",
+    "bip32": "^4.0.0",
+    "bitcoinjs-lib": "^6.1.5",
+    "bitcoinjs-message": "^2.2.0",
+    "tiny-secp256k1": "^2.2.3",
+    "winston": "^3.13.0"
```
andreabadesso: Added those dependencies here so the projects can use `peerDependencies` and share the dependency

packages/common/src/utils/index.utils.ts
tuliomir, line 1: chore: We should have the Hathor header on this file

tuliomir (approved):

packages/common/src/utils/index.utils.ts
andreabadesso: Fixed in 497f843638571767f283d3cc58034173aa589de9

tuliomir (approved):

packages/common/src/utils/alerting.utils.ts
```diff
@@ -28,9 +28,9 @@ export const addAlert = async (
   title: string,
   message: string,
   severity: Severity,
-  metadata?: unknown,
+  metadata: unknown = null,
```
pedroferreira1: Why did you set this default value? Shouldn't we have the optional parameters in the end of the method?

packages/daemon/package.json
```diff
@@ -32,20 +32,26 @@
     "@types/mysql": "^2.15.21",
     "@types/node": "^17.0.45",
     "@types/ws": "^8.5.5",
-    "@typescript-eslint/eslint-plugin": "^6.7.3",
-    "@typescript-eslint/parser": "^6.7.3",
+    "@typescript-eslint/eslint-plugin": "^7.4.0",
```
pedroferreira1: Why did you upgrade this major version of this package? Shouldn't we do any upgrades in a separate PR to be better reviewed?

packages/daemon/package.json
```diff
@@ -32,20 +32,26 @@
...
+    "@typescript-eslint/parser": "^7.4.0",
+    "eslint": "^8.57.0",
```
pedroferreira1: Why do you need `eslint` now in the `daemon`?

packages/wallet-service/package.json
```diff
@@ -36,22 +33,21 @@
...
+    "uuid": "^8.3.0"
   },
   "devDependencies": {
     "@types/aws-lambda": "^8.10.95",
     "@types/jest": "^27.0.24",
     "@types/node": "^18.0.4",
-    "@typescript-eslint/eslint-plugin": "^6.7.4",
-    "@typescript-eslint/parser": "^3.3.0",
+    "@typescript-eslint/eslint-plugin": "^7.4.0",
+    "@typescript-eslint/parser": "^7.4.0",
```
pedroferreira1: Why did you upgrade this lib? It was 4 majors before.

packages/wallet-service/package.json
```diff
@@ -72,5 +68,12 @@
     "typescript-eslint": "0.0.1-alpha.0",
     "webpack": "^5.88.2",
     "webpack-node-externals": "^3.0.0"
+  },
+  "peerDependencies": {
+    "@aws-sdk/client-apigatewaymanagementapi": "3.540.0",
+    "@aws-sdk/client-lambda": "3.540.0",
+    "@aws-sdk/client-sqs": "3.540.0",
+    "@hathor/wallet-lib": "^0.39.0",
+    "winston": "^3.13.0"
```
pedroferreira1: Any reason to upgrade this package?

packages/wallet-service/src/types.ts
```diff
@@ -798,15 +751,4 @@ export interface WalletBalanceValue {
...
- */
-export enum Severity {
-  CRITICAL = 'critical',
-  MAJOR = 'major',
-  MEDIUM = 'medium',
-  MINOR = 'minor',
-  WARNING = 'warning',
-  INFO = 'info',
-}
+export { Severity, Transaction, TxInput, TxOutput, DecodedOutput };
```
pedroferreira1: Why do you need to export the imported types?

package.json
```diff
@@ -3,22 +3,40 @@
...
-  "packageManager": "yarn@4.1.0"
+  "packageManager": "yarn@4.1.0",
+  "dependencies": {
+    "@aws-sdk/client-apigatewaymanagementapi": "3.540.0",
+    "@aws-sdk/client-lambda": "3.540.0",
+    "@aws-sdk/client-sqs": "3.540.0",
+    "@hathor/wallet-lib": "0.39.0",
+    "bip32": "^4.0.0",
+    "bitcoinjs-lib": "^6.1.5",
+    "bitcoinjs-message": "^2.2.0",
```
pedroferreira1, line 40: Does it make sense to have this as peerDependency in the wallet service package.json?  You are duplicating this package

packages/daemon/package.json
andreabadesso: Reverted all package changes related to eslint, I'll do it in this PR:
https://github.com/HathorNetwork/hathor-wallet-service/pull/156

packages/daemon/package.json
andreabadesso: Reverted all package changes related to eslint, I'll do it in this PR:
https://github.com/HathorNetwork/hathor-wallet-service/pull/156

packages/wallet-service/package.json
andreabadesso: Reverted all package changes related to eslint, I'll do it in this PR:
https://github.com/HathorNetwork/hathor-wallet-service/pull/156

packages/wallet-service/package.json
andreabadesso: Since we were using caret ranges in the version (`^3.3.0`), we were already installing the latest non-major version (< `4.0.0`) after a `yarn install`

image

image

So this will not cause the package to get updated, it's just setting the version we're actually using

packages/wallet-service/src/types.ts
andreabadesso: We don't, this was done initially because I was only planning to move the types on the next PR

But refactored the code to import from the common project, can you please check again?

package.json
andreabadesso: It does, refactored in b6e4368e865b8947ab22ab329343da05891117c5

packages/common/src/utils/alerting.utils.ts
andreabadesso: Added a comment in this line: a410a8ca63de34d48327ca4e735cc2ae62f03d12

I'll refactor the `logger` to be shared between projects, so `logger` will stop being a parameter there, so this default to null is temporary until then

pedroferreira1: I still don't understand the default `null`. We use default values for parameters in the end of the list of parameters, which is not the case.

packages/wallet-service/tests/integration.test.ts
```diff
@@ -4,7 +4,8 @@ import eventTemplate from '@events/eventTemplate.json';
 import { loadWallet, loadWalletFailed } from '@src/api/wallet';
 import { createWallet, getMinersList } from '@src/db';
 import * as txProcessor from '@src/txProcessor';
-import { Transaction, WalletStatus, TxInput, Severity } from '@src/types';
+import { WalletStatus  } from '@src/types';
```
tuliomir: style: Lint fix
```suggestion
import { WalletStatus } from '@src/types';
```

packages/wallet-service/tests/utils/pushnotification.utils.test.ts
```diff
@@ -6,7 +6,8 @@ import { sendMulticastMock, messaging, initFirebaseAdminMock } from '@tests/util
 import { logger } from '@tests/winston.mock';
 import { PushNotificationUtils, PushNotificationError, buildFunctionName, FunctionName } from '@src/utils/pushnotification.utils';
 import * as pushnotificationUtils from '@src/utils/pushnotification.utils';
-import { SendNotificationToDevice, Severity } from '@src/types';
+import { SendNotificationToDevice } from '@src/types';
+import { Severity  } from '@wallet-service/common/src/types';
```
tuliomir: style: Lint fix
```suggestion
import { Severity } from '@wallet-service/common/src/types';
```

packages/common/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "@wallet-service/common",
+  "packageManager": "yarn@4.1.0",
+  "peerDependencies": {
+    "@aws-sdk/client-lambda": "3.540.0",
+    "@hathor/wallet-lib": "0.39.0",
+    "winston": "^3.13.0"
```
tuliomir, line 11: question(non-blocking): Here we have a mix of version definitions. Some are pinned to an exact value and some are with a flexible caret range.

Should we decide for one approach or another? Or can we leave this mixed as it currently is?

package.json
```diff
@@ -3,22 +3,40 @@
   "version": "1.5.0",
   "workspaces": [
     "packages/daemon",
-    "packages/wallet-service"
+    "packages/wallet-service",
+    "packages/common"
```
tuliomir: suggestion: Order this alphabetically, as it will probably be automatically fixed at some point in the future
```js,
    "packages/common",
    "packages/daemon",
    "packages/wallet-service"
```

tuliomir (approved):

packages/common/package.json
andreabadesso: I'll fix all versions in a future PR

package.json
andreabadesso: Done in 408d995a493a80f4ca8d17e1afe10bf8c669f2e1

packages/wallet-service/tests/utils/pushnotification.utils.test.ts
andreabadesso: Done in 408d995a493a80f4ca8d17e1afe10bf8c669f2e1

packages/wallet-service/tests/integration.test.ts
andreabadesso: Done in 408d995a493a80f4ca8d17e1afe10bf8c669f2e1

packages/common/src/utils/alerting.utils.ts
andreabadesso: Removed the null default in cae81acc04b833d561559f1f5f8156481057c40d

pedroferreira1 (approved):

PR #162, opened 2024-04-25
fix: serverless failing

andreabadesso:
### Motivation

Serverless was failing to package because of multiple issues related to the new yarn workspaces monorepo structure:

1. It wasn't able to find the common module:

```
Error: Cannot find module './@wallet-service/common/package.json'
```

Upon investigation, I found out that this was happening because it yarn workspaces symlinks internal modules to the package's `node_modules` and `serverless-monorepo` was cleaning it

The fix for this issue was adding it as a peer dependency, which actually copied the dependency to the `node_modules` folder and also adding it to be parsed by the `ts-loader`

2. Peer dependencies were not being transpiled by webpack

This is a weird one, but after we solved the issue with the common module, the package was failing in runtime because of errors on the imports

I wasn't to dive deep into why this was the case, but moving the peer dependencies to be dependencies fixed this issue.

My suggestion is to move on with this PR because it is stopping us from fixing a bug affecting our production server (NFT metadata is not being created on S3) and investigate in a KTLO in the near future.

The side effect is only that `yarn install` will take a little longer because it will not hoist the peer dependencies into the root package.

### Acceptance Criteria

- The wallet-service should be able to be packaged and deployed using the common module
- Dependencies should be properly transpiled by webpack

### Checklist

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -2,6 +2,7 @@ compressionLevel: mixed
 
 enableGlobalCache: false
 
+# Without this setting, the common package is hoisted in the root node_modules, causing the serverless-monorepo plugin to fail
 nmHoistingLimits: workspaces
 
 nodeLinker: node-modules
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -2,6 +2,6 @@ compressionLevel: mixed
 
 enableGlobalCache: false
 
-nmHoistingLimits: dependencies
+nmHoistingLimits: workspaces
 
 nodeLinker: node-modules
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "@aws-sdk/client-lambda": "3.540.0",
     "@aws-sdk/client-sqs": "3.540.0",
     "@hathor/wallet-lib": "0.39.0",
+    "@wallet-service/common": "1.5.0",
     "bip32": "^4.0.0",
     "bitcoinjs-lib": "^6.1.5",
     "bitcoinjs-message": "^2.2.0",
--- a/packages/wallet-service/src/api/wallet.ts
+++ b/packages/wallet-service/src/api/wallet.ts
@@ -351,8 +351,7 @@ export const load: APIGatewayProxyHandler = middy(async (event) => {
      */
     await invokeLoadWalletAsync(xpubkeyStr, maxGap);
   } catch (e) {
-    logger.error('Error on lambda wallet invoke', e);
-
+    logger.error(e);
     const newRetryCount = wallet.retryCount ? wallet.retryCount + 1 : 1;
     // update wallet status to 'error'
     await updateWalletStatus(mysql, walletId, WalletStatus.ERROR, newRetryCount);
--- a/packages/wallet-service/webpack.config.js
+++ b/packages/wallet-service/webpack.config.js
@@ -24,6 +24,11 @@ module.exports = {
     filename: '[name].js',
   },
   target: 'node',
+  // The bundle gets too big if we allow webpack to bundle all dependencies so
+  // we remove them from the bundle (they get loaded in runtime).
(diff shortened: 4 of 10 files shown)


.yarnrc.yml
```diff
@@ -2,6 +2,6 @@ compressionLevel: mixed
 
 enableGlobalCache: false
 
-nmHoistingLimits: dependencies
+nmHoistingLimits: workspaces
```
pedroferreira1, line 6: Why do you need to change this?

packages/wallet-service/src/api/wallet.ts
```diff
@@ -351,8 +351,7 @@ export const load: APIGatewayProxyHandler = middy(async (event) => {
      */
     await invokeLoadWalletAsync(xpubkeyStr, maxGap);
   } catch (e) {
-    logger.error('Error on lambda wallet invoke', e);
-
+    logger.error(e);
```
pedroferreira1, line 354: Why did you change this?

packages/wallet-service/webpack.config.js
```diff
@@ -24,7 +24,9 @@ module.exports = {
     filename: '[name].js',
   },
   target: 'node',
-  externals: [nodeExternals()],
+  externals: [nodeExternals({
+    allowlist: [new RegExp("@wallet-service/common*")],
```
pedroferreira1, line 33: Why do you need this?

pedroferreira1 (approved):

package.json
```diff
@@ -34,6 +34,7 @@
     "@aws-sdk/client-lambda": "3.540.0",
     "@aws-sdk/client-sqs": "3.540.0",
     "@hathor/wallet-lib": "0.39.0",
+    "@wallet-service/common": "1.5.0",
```
tuliomir, line 37: question(non-blocking): Will this require us to publish the wallet service common on `npm`? Or will `yarn` be able to identify that this package is not to be taken from an external source?

This is not clear to me because this is the root `package.json`, not a sub module.

tuliomir (commented):
suggestion: I agree with your suggestion to move on with this PR and investigate the peer dependencies issue through a KTLO.

My recommendation would be to add the issue number to this PR description now, so that it's easier to track these changes in the future.

package.json
andreabadesso: It won't, yarn will recognize it is a internal package and symlink it from the package

tuliomir (approved):

packages/wallet-service/src/api/wallet.ts
andreabadesso: That's because Winston as it is configured ignores the second parameter

Created an issue to refactor this: https://github.com/HathorNetwork/hathor-wallet-service/issues/165

.yarnrc.yml
andreabadesso: Without this setting, the common package is being hoisted in the root node_modules, causing the serverless-monorepo plugin to fail

packages/wallet-service/webpack.config.js
andreabadesso: Yarn workspaces hoists modules to the root node_modules directory leaving any incompatible (different semver, etc.) modules with the dependent workspace's node_modules directory, which is the case of our internal package

So we set it as a [external](https://webpack.js.org/configuration/externals/), which is loaded during runtime from the node_modules and not included in the webpack bundle

andreabadesso: The bundle gets too big if we allow webpack to bundle all dependencies so we remove them from the bundle (they [get loaded in runtime](https://webpack.js.org/configuration/externals/))

But we can't do this with the common module, otherwise it would not be seen by the serverless-monorepo package

pedroferreira1 (approved):

PR #163, opened 2024-05-01
tests: missing tests on websocket

andreabadesso:
### Motivation

Changes were made to the websocket utils without adding new tests, causing the coverage to drop below the minimum percentage, this PR should introduce new tests to the changed method

### Acceptance Criteria

- We should test the `sendMessageToClient` util

### Checklist


tuliomir (approved):
By looking into the PR's checks action runs I could validate the coverage percentage, however I had no way to check how it was before this PR ( other than also checking the last commit's action runs ).

Would it be possible to add `CodeCov` to this monorepo? I've seen that they already have [a _flags_ feature](https://docs.codecov.com/docs/flags) to help with that. It would also add automatic annotations to the PR to help identify untested code.

pedroferreira1 (approved):

PR #170, opened 2024-06-04
fix: broken daemon build

andreabadesso:
### Motivation

The daemon was failing to load the common module during runtime because the typescript was not being transpiled to javascript. This was not an issue for the lambdas because the wallet-service uses webpack which has a typescript loader

### Acceptance Criteria

- The build script from the daemon should build the common module using its own tsconfig
- We should have an index file on the common that exposes all methods

### Checklist


packages/common/src/utils/alerting.utils.ts
```diff
@@ -7,17 +7,8 @@
...
 import { Logger } from 'winston';
 
-assertEnvVariablesExistence([
-  'NETWORK',
-  'APPLICATION_NAME',
-  'ACCOUNT_ID',
-  'ALERT_MANAGER_REGION',
-  'ALERT_MANAGER_TOPIC',
-]);
-
```
andreabadesso, line 20: We have a config method that already checks for those variables and this check shouldn't be in a lib

packages/daemon/tsconfig.json
```diff
@@ -10,6 +10,9 @@
     "outDir": "./dist",
     "types": ["node", "jest"]
   },
+  "references": [
+      { "path": "../../node_modules/@wallet-service/common" }
+  ],
```
andreabadesso, line 18: This makes tsconfig compile the common module tsconfig

Dockerfile
```diff
@@ -20,18 +19,11 @@ RUN corepack enable
...
 
-FROM node:20-alpine
-
-WORKDIR /app
+RUN yarn workspace sync-daemon run build
 
-COPY --from=builder /app/packages/daemon/dist .
-COPY --from=builder /app/packages/daemon/node_modules ./node_modules
+# This will remove all dev dependencies and install production deps only
+RUN yarn workspaces focus -A --production
```
andreabadesso, line 28: I added the -A flag because otherwise we wouldn't be able to use wallet-lib as a peer dependency and would have issues with bitcore-lib

This is not installing dependencies from the wallet-service lambdas as we are ignoring its package in `.dockerignore`

packages/daemon/__tests__/integration/balances.test.ts
```diff
@@ -34,9 +34,14 @@ jest.mock('../../src/config', () => {
...
-jest.mock('@wallet-service/common/src/utils/index.utils', () => ({
-  assertEnvVariablesExistence: jest.fn(),
-}));
+jest.mock('@wallet-service/common', () => {
+  const originalModule = jest.requireActual('@wallet-service/common');
+
+  return {
+    ...originalModule,
+    assertEnvVariablesExistence: jest.fn(),
+  };
```
tuliomir: thought(non-blocking): There's an argument to only building explicitly what is being tested with these mocks, instead of using them whole and just changing some parts. However, this adds complexity to the testing phase when we're testing an entire module.

I don't think we should change this PR on this matter, but it's good to keep in mind that the least we implement on these mocks, the better our tests tend to be. Do you agree?

packages/daemon/package.json
```diff
@@ -20,7 +21,7 @@
     "test_images_wait_for_db": "yarn dlx ts-node ./__tests__/integration/scripts/wait-for-db-up.ts",
     "test_images_wait_for_ws": "yarn dlx ts-node ./__tests__/integration/scripts/wait-for-ws-up.ts",
     "test_images_setup_database": "yarn dlx ts-node ./__tests__/integration/scripts/setup-database.ts",
-    "test": "jest --coverage",
+    "test": "jest --coverage --verbose",
```
tuliomir: question(non-blocking): Should we keep the `--verbose` flag here?

packages/daemon/tsconfig.json
tuliomir: question(non-blocking): Should we add these instructions to a document on the repository itself?

There seem to be many of those small details that can break the production environment and delay future maintenance releases if left unchecked.

tuliomir (approved):

Dockerfile
```diff
@@ -20,18 +19,11 @@ RUN corepack enable
 RUN yarn set version 4.1.0
 
 # This will install dependencies for the sync-daemon, devDependencies included:
```
pedroferreira1: Is this comment still valid?

pedroferreira1 (approved):

Dockerfile
andreabadesso: No, updated in 3e1a48f162c8755e6042000f17493567e414a534

Thanks

packages/daemon/package.json
andreabadesso: No! Removed it in 3e1a48f162c8755e6042000f17493567e414a534

Thanks

packages/daemon/tsconfig.json
andreabadesso: I just found out that typescript implements its own JSON parser, which allows for comments in JSON

Added a comment in tsconfig.json in 536ee047812a2aeb8703120a5e3da64b016e0960

andreabadesso: Also created an issue so we document these details better

andreabadesso: https://github.com/HathorNetwork/hathor-wallet-service/issues/172

packages/daemon/__tests__/integration/balances.test.ts
andreabadesso: We actually don't need this mock anymore, as we removed it from the common module

e5988056b9ace7762f1728bfd1d8811da71a6f25

packages/daemon/tsconfig.json
tuliomir: Great! It will surely help with future maintenance

PR #177, opened 2024-06-14
fix: fetchMinRewardBlocks fails when reward_min_blocks is 0

andreabadesso:
### Motivation

`fetchMinRewardBlocks` is throwing an error when `reward_min_blocks` is 0 on the fullnode

### Acceptance Criteria

- `fetchMinRewardBlocks` should not throw when the `reward_min_blocks` returned by the fullnode is 0

### Checklist

### DIFF
--- a/packages/daemon/src/services/index.ts
+++ b/packages/daemon/src/services/index.ts
@@ -528,7 +528,7 @@ export const fetchMinRewardBlocks = async () => {
 
   const rewardSpendMinBlocks = get(response, 'data.reward_spend_min_blocks');
 
-  if (!rewardSpendMinBlocks) {
+  if (rewardSpendMinBlocks == null) {
     throw new Error('Failed to fetch reward spend min blocks');
   }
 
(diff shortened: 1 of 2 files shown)


luislhl (approved):

packages/daemon/src/services/index.ts
```diff
@@ -528,7 +528,7 @@ export const fetchMinRewardBlocks = async () => {
 
   const rewardSpendMinBlocks = get(response, 'data.reward_spend_min_blocks');
 
-  if (!rewardSpendMinBlocks) {
+  if (rewardSpendMinBlocks == null) {
```
glevco, line 531: Should we also check for `undefined`?

andreabadesso: This is also checking for `undefined`, `==` (as opposed to `===`) checks for both null and undefined in JavaScript due to type coercion

glevco: Ok, thanks!

glevco (approved):

PR #178, opened 2024-06-20
fix: handle tx with no inputs and outputs

andreabadesso:
### Motivation

This should fix https://github.com/HathorNetwork/on-call-incidents/issues/179

### Acceptance Criteria

- We should add `NODE_ENV` to the `test_integration` script so that it runs locally without having to manually export it
- We should fix an issue when a transaction has no inputs and outputs, like PoA blocks from PoA side-dags

### Checklist

### DIFF
--- a/packages/daemon/package.json
+++ b/packages/daemon/package.json
@@ -22,7 +22,7 @@
     "test_images_wait_for_ws": "yarn dlx ts-node ./__tests__/integration/scripts/wait-for-ws-up.ts",
     "test_images_setup_database": "yarn dlx ts-node ./__tests__/integration/scripts/setup-database.ts",
     "test": "jest --coverage",
-    "test_integration": "yarn run test_images_up && yarn run test_images_wait_for_db && yarn run test_images_wait_for_ws && yarn run test_images_setup_database && yarn run test_images_migrate && yarn run test_images_integration && yarn run test_images_down"
+    "test_integration": "yarn run test_images_up && yarn run test_images_wait_for_db && yarn run test_images_wait_for_ws && yarn run test_images_setup_database && yarn run test_images_migrate && yarn run test_images_integration"
   },
   "name": "sync-daemon",
   "author": "André Abadesso",
--- a/packages/daemon/package.json
+++ b/packages/daemon/package.json
@@ -22,7 +22,7 @@
     "test_images_wait_for_ws": "yarn dlx ts-node ./__tests__/integration/scripts/wait-for-ws-up.ts",
     "test_images_setup_database": "yarn dlx ts-node ./__tests__/integration/scripts/setup-database.ts",
     "test": "jest --coverage",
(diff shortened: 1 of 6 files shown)


packages/daemon/src/db/index.ts
```diff
@@ -438,25 +440,20 @@ export const voidTransaction = async (
       // for locked authorities, it doesn't make sense to perform the same operation. The authority needs to be
       // unlocked before it can be spent. In case we're just adding new locked authorities, this will be taken
       // care by the first sql query.
-
-      // update address_tx_history with one entry for each pair (address, token)
-      entries.push(txId);
```
andreabadesso, line 443: This doesn't even makes sense, because this method is called one tx at a time...

Probably lazy code by me when I migrated it from the old daemon

r4mmer (dismissed):

packages/daemon/package.json
```diff
@@ -17,12 +17,12 @@
     "test_images_up": "docker-compose -f ./__tests__/integration/scripts/docker-compose.yml up -d",
     "test_images_down": "docker-compose -f ./__tests__/integration/scripts/docker-compose.yml down",
     "test_images_integration": "jest --config ./jest_integration.config.js --runInBand --forceExit",
-    "test_images_migrate": "DB_NAME=hathor DB_PORT=3380 DB_PASS=hathor DB_USER=hathor yarn run sequelize-cli --migrations-path ../../db/migrations --config ./__tests__/integration/scripts/sequelize-db-config.js db:migrate",
+    "test_images_migrate": "NODE_ENV=test DB_NAME=hathor DB_PORT=3380 DB_PASS=hathor DB_USER=hathor yarn run sequelize-cli --migrations-path ../../db/migrations --config ./__tests__/integration/scripts/sequelize-db-config.js db:migrate",
```
andreabadesso, line 20: This was causing the tests to fail locally

packages/daemon/package.json
```diff
@@ -17,12 +17,12 @@
...
     "test_images_wait_for_db": "yarn dlx ts-node ./__tests__/integration/scripts/wait-for-db-up.ts",
     "test_images_wait_for_ws": "yarn dlx ts-node ./__tests__/integration/scripts/wait-for-ws-up.ts",
     "test_images_setup_database": "yarn dlx ts-node ./__tests__/integration/scripts/setup-database.ts",
     "test": "jest --coverage",
-    "test_integration": "yarn run test_images_up && yarn run test_images_wait_for_db && yarn run test_images_wait_for_ws && yarn run test_images_setup_database && yarn run test_images_migrate && yarn run test_images_integration && yarn run test_images_down"
+    "test_integration": "yarn run test_images_up && yarn run test_images_wait_for_db && yarn run test_images_wait_for_ws && yarn run test_images_setup_database && yarn run test_images_migrate && yarn run test_images_integration"
```
glevco: Why this was removed?

glevco (dismissed):

packages/daemon/package.json
andreabadesso: Re-added it! Thanks!

glevco (approved):

r4mmer (approved):

PR #182, opened 2024-07-30
feat: handling the vertex_removed event

andreabadesso:
### Motivation

- We should handle the `VERTEX_REMOVED` event
- We should have balance integration tests for the `invalid_mempool` scenario
- We should fix multiple db methods that were failing on empty params

### Acceptance Criteria

- Include here all things that this PR should solve

### Checklist


packages/daemon/src/db/index.ts
```diff
@@ -497,6 +497,10 @@ export const updateAddressTablesWithTx = async (
   timestamp: number,
   addressBalanceMap: StringMap<TokenBalanceMap>,
 ): Promise<void> => {
+  if (Object.keys(addressBalanceMap).length === 0) {
+    // No need to do anything here, the transaction most likely has no inputs and outputs
```
glevco: "most likely"? Is it possible that this is not the case?

glevco (dismissed):

packages/daemon/src/db/index.ts
andreabadesso: No.

Removed the comment

glevco: Did you push it?

r4mmer (dismissed):

packages/daemon/src/db/index.ts
andreabadesso: Ops, I pushed it to the invalid scripts branch, just cherry-picked it, can you check again?

r4mmer (approved):

glevco (approved):

PR #185, opened 2024-10-03
fix: non-standard scripts

andreabadesso:
### Motivation

Wallet Service will crash if a utxo with an non standard script (which is unable to be decoded) is spent in the network.

This currently never happens because there is a bug in the fullnode preventing those events from being broadcasted

### Acceptance Criteria

- We should skip balance calculation for undecoded scripts
- We should add a helper to reuse code in the tests

### Checklist


glevco (dismissed):
LGTM, just a comment, they're not "invalid scripts". They're completely valid, just non-standard

glevco (approved):

packages/daemon/src/utils/wallet.ts
```diff
@@ -294,11 +296,11 @@ export const prepareInputs = (inputs: EventTxInput[], tokens: string[]): TxInput
       // @ts-ignore
       script: utxo.script,
       token,
-      decoded: {
+      decoded: output.decoded ? {
         type: output.decoded.type,
         address: output.decoded.address,
         timelock: output.decoded.timelock,
-      },
```
r4mmer, line 301: Is this decoded from the fullnode or local?
A data output has a valid `decoded` with `type` and `data` but it does not have an `address`.

andreabadesso: The fullnode sends it already decoded, exactly with this structure (including the `address`). This comes from the event queue

r4mmer (approved):

PR #186, opened 2024-10-04
chore: using fixed dependencies for all projects

andreabadesso:
### Motivation

We should use fixed versions

### Acceptance Criteria

- Update all projects to use fixed versions instead of caret

### Checklist


package.json
```diff
@@ -16,7 +16,7 @@
   "author": "André Abadesso <andre.abadesso@gmail.com>",
   "private": true,
   "devDependencies": {
-    "@types/jest": "^29.5.12",
+    "@types/jest": "29.5.13",
```
andreabadesso, line 19: This is funny, using `29.5.12` won't work. They probably removed it from the registry

We were actually using `29.5.13`

glevco (approved):

luislhl (approved):

PR #199, opened 2024-12-04
fix: slow address generation

andreabadesso:
## Motivation

We currently iterate through all address indices of a wallet when a new address needs to be generated, this is taking up to 20 seconds when the wallet has many addresses

## Acceptance Criteria
- We should add a new composite index on `(wallet_id, index)`, improving query performance
- All existing functionality should be maintained (address generation, gap limit)
- Tests should be passing and should cover the new implementation

### Checklist

### DIFF
--- a/packages/daemon/src/db/index.ts
+++ b/packages/daemon/src/db/index.ts
@@ -1563,7 +1563,18 @@ export const getTokenSymbols = async (
   }, {}) as unknown as StringMap<string>;
 };
 
-// TODO: docstring
+/**
+ * Get the maximum index among a set of addresses for a specific wallet.
+ *
+ * @remarks
+ * This function is used to find the highest index used among a set of addresses that belong to a wallet.
+ * This is particularly useful when we need to determine where to start generating new addresses from.
+ *
+ * @param mysql - Database connection
+ * @param walletId - The ID of the wallet to check
+ * @param addresses - Array of addresses to check
+ * @returns The highest index found among the addresses, or null if no addresses are found
+ */
 export const getMaxIndexAmongAddresses = async (
   mysql: MysqlConnection,
   walletId: string,
@@ -1586,7 +1597,18 @@ export const getMaxIndexAmongAddresses = async (
   return results[0].max_index;
 };
 
-// TODO: docstring
+/**
+ * Get the maximum index used by any address in a wallet.
+ *
+ * @remarks
+ * This function retrieves the highest index used by any address in the given wallet.
+ * This is essential for wallet synchronization and address generation, as it helps determine
+ * the starting point for generating new addresses and maintaining the gap limit.
+ *
+ * @param mysql - Database connection
+ * @param walletId - The ID of the wallet to check
+ * @returns The highest index used in the wallet, or null if no addresses are found
+ */
 export const getMaxWalletAddressIndex = async (
   mysql: MysqlConnection,
   walletId: string,
--- a/packages/daemon/src/db/index.ts
+++ b/packages/daemon/src/db/index.ts
@@ -1592,7 +1592,7 @@ export const getMaxWalletAddressIndex = async (
   walletId: string,
 ): Promise<number | null> => {
   const [results] = await mysql.query<MaxAddressIndexRow[]>(
-    `SELECT MAX(\`index\`)
+    `SELECT MAX(\`index\`) AS max_index
        FROM address
       WHERE wallet_id = ?
      `, [walletId]
--- a/packages/daemon/src/db/index.ts
+++ b/packages/daemon/src/db/index.ts
@@ -1059,16 +1059,15 @@ export const incrementTokensTxCount = async (
 };
 
 /**
- * Given an xpubkey, generate its addresses.
+ * Generate a batch of addresses from a given xpubkey.
  *
  * @remarks
- * Also, check which addresses are used, taking into account the maximum gap of unused addresses (maxGap).
- * This function doesn't update anything on the database, just reads data from it.
- *
- * @param mysql - Database connection
- * @param xpubkey - The xpubkey
- * @param maxGap - Number of addresses that should have no transactions before we consider all addresses loaded
- * @returns Object with all addresses for the given xpubkey and corresponding index
+ * This function generates addresses starting from a specific index.
+ * 
+ * @param xpubkey - The extended public key to derive addresses from
+ * @param startIndex - The index to start generating addresses from
+ * @param count - How many addresses to generate
+ * @returns A map of addresses to their corresponding indices
  */
 export const generateAddresses = async (
   xpubkey: string,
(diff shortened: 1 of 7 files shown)


packages/daemon/src/db/index.ts
```diff
@@ -1058,78 +1059,29 @@ export const incrementTokensTxCount = async (
...
+ * @param startIndex - The index to start generating addresses from
+ * @param count - How many addresses to generate
+ * @returns A map of addresses to their corresponding indices
  */
-export const generateAddresses = async (mysql: MysqlConnection, xpubkey: string, maxGap: number): Promise<GenerateAddresses> => {
-  const existingAddresses: AddressIndexMap = {};
-  const newAddresses: AddressIndexMap = {};
-  const allAddresses: string[] = [];
-
+export const generateAddresses = async (
```
glevco: Should we move this function to another file since it doesn't access the database anymore?

packages/daemon/src/services/index.ts
```diff
@@ -289,31 +293,80 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
...
+        if (!maxAmongAddresses || !maxWalletIndex) {
+          // Do nothing, this is unexpected and an error should have been logged already
+          continue;
+        }
+
+        const diff = maxWalletIndex - maxAmongAddresses;
+
+        if (diff < walletDetails.maxGap) {
+          // We need to generate addresses
+          const addresses = await generateAddresses(walletDetails.xpubkey, maxWalletIndex + 1, walletDetails.maxGap);
```
pedroferreira1: The last parameter is the count of addresses to be generated. You don't need to always generate `maxGap` addresses, it can be `walletDetails.maxGap - diff`, right?

packages/daemon/src/services/index.ts
```diff
@@ -289,31 +293,80 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
...
+        }
       }
+
       // update wallet_balance and wallet_tx_history tables
       const walletBalanceMap: StringMap<TokenBalanceMap> = getWalletBalanceMap(addressWalletMap, addressBalanceMap);
       await updateWalletTablesWithTx(mysql, hash, timestamp, walletBalanceMap);
 
-      const tx: Transaction = {
+      // validate address balances
+      await validateAddressBalances(mysql, Object.keys(addressBalanceMap));
```
pedroferreira1: Why do you need this now?

packages/daemon/src/db/index.ts
```diff
@@ -1609,3 +1561,45 @@ export const getTokenSymbols = async (
...
+  walletData: Array<{walletId: string, addresses: string[]}>
+): Promise<Map<string, {maxAmongAddresses: number | null, maxWalletIndex: number | null}>> => {
+  if (walletData.length === 0) {
+    return new Map();
+  }
+
+  const allAddresses = walletData.flatMap(d => d.addresses);
+  const walletIds = walletData.map(d => d.walletId);
+
+  const [results] = await mysql.query<MaxAddressIndexRow[]>(
```
pedroferreira1, line 1572: I think this query deserves better comment description. It's not so easy to understand with the group by wallets to get the max of each wallet.

packages/daemon/src/db/index.ts
```diff
@@ -1609,3 +1561,45 @@ export const getTokenSymbols = async (
     return prev;
   }, {}) as unknown as StringMap<string>;
 };
+
+/**
+ * Get maximum indices for multiple wallets in a single query.
+ * 
+ * @remarks
+ * This is an optimized version that combines both getMaxIndexAmongAddresses and getMaxWalletAddressIndex
+ * into a single query for multiple wallets. This reduces the number of database round trips significantly.
```
glevco: Do these functions exist (`getMaxIndexAmongAddresses` and `getMaxWalletAddressIndex`)? I couldn't find them. Maybe explain what are the two queries you're combining instead of mentioning those functions

packages/daemon/src/db/index.ts
```diff
@@ -1609,3 +1561,45 @@ export const getTokenSymbols = async (
...
+ * 
+ * @param mysql - Database connection
+ * @param walletData - Array of objects containing wallet IDs and their associated addresses
+ * @returns Map of wallet IDs to their maximum indices (both among specific addresses and overall)
+ */
+export const getMaxIndicesForWallets = async (
+  mysql: MysqlConnection,
+  walletData: Array<{walletId: string, addresses: string[]}>
+): Promise<Map<string, {maxAmongAddresses: number | null, maxWalletIndex: number | null}>> => {
```
glevco, line 1564: Can the max indices ever be `null`? I mean, if a wallet exists it should always have a max index, no?

packages/daemon/src/services/index.ts
```diff
@@ -166,7 +168,7 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
   try {
     const fullNodeEvent = context.event as FullNodeEvent;
     const now = getUnixTimestamp();
-    const { NEW_TX_SQS, PUSH_NOTIFICATION_ENABLED } = getConfig();
+    const { PUSH_NOTIFICATION_ENABLED } = getConfig();
```
glevco: Was this moved for any specific reason or is it the same? Before we were using `getConfig()` and now it's a direct import

packages/daemon/src/services/index.ts
```diff
@@ -289,31 +293,80 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
...
       const seenWallets = new Set();
-      for (const wallet of Object.values(addressWalletMap)) {
-        const walletId = wallet.walletId;
+      const addressesPerWallet = Object.entries(addressWalletMap).reduce(
+        (result: StringMap<{ addresses: string[], walletDetails: Wallet }>, [address, wallet]: [string, Wallet]) => {
+        const { walletId } = wallet;
 
-        // this map might contain duplicate wallet values, as 2 different addresses might belong to the same wallet
-        if (seenWallets.has(walletId)) continue;
         seenWallets.add(walletId);
```
glevco: Do you need to keep this variable? You could just use the keys of `addressesPerWallet`, right?

packages/daemon/src/services/index.ts
```diff
@@ -289,31 +293,80 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
...
+        if (!result[walletId]) {
+          result[walletId] = {
+            addresses: [],
+            walletDetails: wallet,
+          }
+        }
+
+        // Add the current key to the array
+        result[walletId].addresses.push(address);
+        result[walletId].walletDetails = wallet;
```
glevco: This line can be removed, right?

packages/daemon/src/services/index.ts
```diff
@@ -289,31 +293,80 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
...
+          logger.error('Failed to get indices for wallet', { walletId });
+          continue;
+        }
+
+        const { maxAmongAddresses, maxWalletIndex } = indices;
+
+        if (!maxAmongAddresses || !maxWalletIndex) {
+          // Do nothing, this is unexpected and an error should have been logged already
+          continue;
+        }
```
glevco: As commented before, `getMaxIndicesForWallets` shouldn't return `null` for those, then you wouldn't need this check. Also, where would this error be logged?

packages/daemon/src/services/index.ts
```diff
@@ -376,6 +429,15 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
...
 
+    if (e instanceof Error) {
+      logger.error('Error handling vertex accepted', {
+        error: e.message,
+        stack: e.stack,
+      });
+    } else {
+      logger.error('Error handling vertex accepted', { error: e });
+    }
+
```
glevco: Should we remove this first `logger.error(e);`?

packages/daemon/src/services/index.ts
andreabadesso: We don't actually need it, but we already have this validation method, it was helpful during tests

I'll remove it, we can add it back later maybe with a flag to validate balances on each transaction

packages/daemon/src/db/index.ts
andreabadesso: Yes, done! 

Thanks

packages/daemon/src/db/index.ts
andreabadesso: No, there might be a brief period of time where it is in the database but not yet initialized, i.e. the addresses were not loaded yet, so it might be `null` in the database

packages/daemon/src/services/index.ts
andreabadesso: It should be! Thanks

packages/daemon/src/services/index.ts
andreabadesso: You're right

packages/daemon/src/services/index.ts
andreabadesso: Right

packages/daemon/src/services/index.ts
andreabadesso: Yes, thanks

packages/daemon/src/services/index.ts
andreabadesso: Fixed, thanks

packages/daemon/src/db/index.ts
andreabadesso: Done, can you please check again?

packages/daemon/src/services/index.ts
andreabadesso: They might return null, specially if the wallet is not yet ready.

Added a better log for that specific case, can you check again?

packages/daemon/src/db/index.ts
glevco: But is it possible that this function would be called in that period? If not, I think you could return non-null values and make the assertion inside this function

andreabadesso: It is possible, the wallet might be initialized at any time as it's initialized in the lambdas

pedroferreira1 (approved):

glevco (approved):

PR #209, opened 2025-01-24
fix: invalid parameter being sent to sendMessageSQS

andreabadesso:
### Motivation

We changed the format of the object being sent to the real-time transaction SQS queue by mistake in https://github.com/HathorNetwork/hathor-wallet-service/pull/199, causing validation to fail on every new transaction sent to the queue.

### Acceptance Criteria

- We should fix the message format
- We should include a helper method to send realtime transactions so we have types for the parameters

### Checklist

### DIFF
--- a/packages/daemon/src/services/index.ts
+++ b/packages/daemon/src/services/index.ts
@@ -388,7 +388,7 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
 
           await sendMessageSQS(JSON.stringify({
             wallets: Array.from(seenWallets),
-            txData,
+            tx: txData,
           }), queueUrl);
         }
       } catch (e) {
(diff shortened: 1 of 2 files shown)


tuliomir (dismissed):

packages/daemon/src/services/index.ts
```diff
@@ -381,15 +382,10 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
 
       try {
         if (seenWallets.length > 0) {
-          const queueUrl = NEW_TX_SQS;
```
pedroferreira1, line 384: You don't need the `NEW_TX_SQS` variable anymore here.

pedroferreira1 (dismissed):

tuliomir (approved):

pedroferreira1 (approved):

PR #216, opened 2025-02-26
chore: upgrade wallet-lib to v1.14.1

r4mmer:
### Motivation

Wallet-lib is a core dependency but is currently 2 major versions outdated.
This PR is the first part upgrading and refactoring the codebase to work with `@hathor/wallet-lib@1.14.1`

### Acceptance Criteria

- Upgrade the wallet-lib version to v1.14.1
- Refactor codebase to work after breaking changes

### Checklist

## Decisions

~~### Decision 1~~

~~Code: DEC-0001~~

~~Some constants are network configured and come from the `/v1a/version` response of the fullnode.~~

~~These constants were previously configured on the config module and updated by the `hathorLib.version.checkApiVersion`.~~
~~This was replaced by having an ApiConfig instance on the storage to avoid having 1 config for all wallets connected.~~

~~Currently we need to have a way to get these configured constants, we could go with one of the following:~~
~~1. Use pre-configured hardcoded constants.~~
~~2. Get from connected fullnode.~~
~~  - The method `maybeRefreshWalletConstants` in `packages/wallet-service/src/commons.ts` can be used to refresh constants based on a request to the fullnode.~~
~~3. Get from wallet-service config.~~

~~Current changes that may change after a decision is made:~~
~~- Uses of `hathorLib.transaction.getMaxOutputsConstant()`~~
~~  - Temporarily fixed by using `constants.MAX_OUTPUTS = 255`.~~
~~- Call to `hathorLib.version.checkApiVersion()` to update variables does not exist~~
~~  - Will use the version api directly so the data can be saved on the database.~~
~~  - Will not save the values until a decision is made.~~

~~### Decision 2~~

~~Code: DEC-0002~~

~~Some constants of the wallet-lib were deprecated, they are being used as sensible defaults but we need to check if we are still going to use these values.~~
~~An example would be the `DEFAULT_SERVERS` and `DEFAULT_SERVER` constants.~~

~~We can either:~~
~~- hardcode the values~~
~~  - includes creating a constants file to keep these values.~~
~~- create a way to ensure some settings are configured~~
~~  - This way we do not depend on sensible defaults~~
~~  - The deploy should fail if these are not configured~~

~~Currently the values are being hardcoded to avoid changing any behavior.~~

## Files

List of files that were changed:
- packages/daemon/src/config.ts
  - Use `String(process.env...)` only in requiredEnvs to ensure typing as `string` and not `string | undefined`.

Files that contain `@hathor/wallet-lib`

- [X] packages/common/src/types.ts
  - No changes, only uses common constants that have not changed.
- [X] packages/common/src/utils/nft.utils.ts
  - No changes, uses utils that exist on most recent version.
- [X] packages/common/src/utils/wallet.utils.ts
  - No changes, only uses common constants that have not changed.
- [X] packages/daemon/src/services/index.ts
  - No changes, but to ensure typing we need to make `getConfig().NETWORK` to be strictly a string.
- [X] packages/daemon/src/types/token.ts
  - Uses `HATHOR_TOKEN_CONFIG` which was deprecated, need to use a combination of `DEFAULT_NATIVE_TOKEN_CONFIG` and `NATIVE_TOKEN_UID`.
- [X] packages/daemon/src/utils/wallet.ts
  - Merged split imports of wallet-lib
  - XXX: Sets properties of Output that do NOT exist.
  - Some ts-ignore may be introducing bugs
- [X] packages/wallet-service/src/api/tokens.ts
  - Changed HATHOR_TOKEN_CONFIG.uid for NATIVE_TOKEN_UID
- [X] packages/wallet-service/src/api/totalSupply.ts
  - Changed HATHOR_TOKEN_CONFIG.uid for NATIVE_TOKEN_UID
- [X] packages/wallet-service/src/api/txOutputs.ts
- [x] packages/wallet-service/src/api/txProposalCreate.ts
- [X] packages/wallet-service/src/api/txProposalSend.ts
- [X] packages/wallet-service/src/api/txhistory.ts
  - Uses `HATHOR_TOKEN_CONFIG` which was deprecated, changed to `NATIVE_TOKEN_UID`.
- [x] packages/wallet-service/src/commons.ts
- [X] packages/wallet-service/src/db/index.ts
- [X] packages/wallet-service/src/types.ts
  - Uses `HATHOR_TOKEN_CONFIG` which was deprecated, need to use a combination of `DEFAULT_NATIVE_TOKEN_CONFIG` and `NATIVE_TOKEN_UID`.
- [x] packages/wallet-service/src/utils.ts
  - Uses constants that do not exist, like `DEFAULT_SERVER`.

Tests files will be changed last since we need to change the methods first.
Package manager files are deferred since they should be updated automatically when upgrading.

- [x] yarn.lock
- [x] package.json
- [x] packages/common/__tests__/utils/nft.utils.test.ts
- [x] packages/common/package.json
- [x] packages/daemon/__tests__/db/index.test.ts
- [x] packages/daemon/__tests__/services/services.test.ts
- [x] packages/daemon/package.json
- [x] packages/wallet-service/AUTHENTICATION.md
- [x] packages/wallet-service/package.json
- [x] packages/wallet-service/tests/api.test.ts
- [x] packages/wallet-service/tests/commons.test.ts
- [x] packages/wallet-service/tests/db.test.ts
- [x] packages/wallet-service/tests/txProposal.test.ts
- [x] packages/wallet-service/tests/utils.test.ts
- [x] packages/wallet-service/tests/utils.ts


packages/daemon/src/types/token.ts
```diff
@@ -23,9 +22,10 @@ export class TokenInfo {
     this.symbol = symbol;
     this.transactions = transactions || 0;
 
-    const hathorConfig = hathorLib.constants.HATHOR_TOKEN_CONFIG;
+    // XXX: get config from settings?
+    const hathorConfig = constants.DEFAULT_NATIVE_TOKEN_CONFIG;
```
r4mmer, line 26: This means that the daemon only supports Hathor as native token.
If we require support for side-dags we should add a nodeConfig module on the daemon (or common).

packages/wallet-service/tests/api.test.ts
```diff
@@ -1152,12 +1159,60 @@ test('changeAuthXpub should fail if signatures do not match', async () => {
   expect(returnBody.details[0].message).toBe('Signatures are not valid');
 });
 
+test('PUT /wallet/auth should fail if we cannot confirm the firstAddress', async () => {
```
r4mmer, line 1161: This test was created just to increase coverage to pass the CI

packages/wallet-service/tests/api.test.ts
```diff
@@ -1323,6 +1379,27 @@ test('loadWallet should update wallet status to ERROR if an error occurs', async
   expect(wallet.status).toStrictEqual(WalletStatus.ERROR);
 }, 30000);
 
+test('loadWalletFailed should create alert if xpubkey is missing', async () => {
```
r4mmer, line 1381: This test was created just to increase coverage to pass the CI

packages/wallet-service/tests/utils/pushnotification.utils.test.ts
```diff
@@ -57,6 +58,7 @@ describe('PushNotificationUtils', () => {
     });
 
     // reload module
+    jest.resetModules();
```
r4mmer, line 61: reset modules is required so that the `import('@src/utils/pushnotification.utils')` reimports the config and use the most recent values

packages/daemon/src/config.ts
```diff
@@ -49,7 +49,7 @@ export const STAGE = process.env.STAGE ?? 'local';
 export const FULLNODE_PEER_ID = process.env.FULLNODE_PEER_ID;
 export const FULLNODE_HOST = process.env.FULLNODE_HOST;
 export const STREAM_ID = process.env.STREAM_ID;
-export const NETWORK = process.env.NETWORK;
+export const NETWORK = String(process.env.NETWORK);
```
andreabadesso, line 52: Why? It should already be a string or undefined

packages/wallet-service/package.json
```diff
@@ -40,12 +40,13 @@
...
   },
   "peerDependencies": {
-    "@hathor/wallet-lib": "0.39.0",
+    "@hathor/wallet-lib": "1.14.1",
     "@wallet-service/common": "1.5.0"
   },
   "devDependencies": {
     "@types/aws-lambda": "8.10.95",
     "@types/jest": "^29.5.13",
+    "@types/joi": "^17.2.3",
```
andreabadesso:
```suggestion
    "@types/joi": "17.2.3",
```

Do the same for the above library

db/migrations/20250306170811-node_version_proxy.js
```diff
@@ -0,0 +1,83 @@
...
+      max_number_inputs: {
+        type: Sequelize.INTEGER.UNSIGNED,
+        allowNull: false,
+      },
+      max_number_outputs: {
+        type: Sequelize.INTEGER.UNSIGNED,
+        allowNull: false,
+      },
+    });
+  }
```
andreabadesso:
```suggestion
const { up } = require('./20210706175820-create-version-data');

await up();
```

packages/common/__tests__/utils/nft.utils.test.ts
```diff
@@ -265,7 +263,7 @@ describe('_updateMetadata', () => {
     });
 
     // eslint-disable-next-line jest/valid-expect
-    expect(NftUtils._updateMetadata('sampleUid', { sampleData: 'fake' }, network, logger))
+    expect(NftUtils._updateMetadata('sampleUid', { sampleData: 'fake' }, 3, logger))
```
andreabadesso, line 344: Was this test passing? Weird

packages/daemon/src/utils/wallet.ts
```diff
@@ -66,6 +63,7 @@ export const prepareOutputs = (outputs: EventTxOutput[], tokens: string[]): TxOu
 
   const preparedOutputs: [number, TxOutputWithIndex[]] = outputs.reduce(
     ([currIndex, newOutputs]: [number, TxOutputWithIndex[]], _output: EventTxOutput): [number, TxOutputWithIndex[]] => {
+      // XXX: Output typing makes no sense here, maybe we should convert from Output to the wallet-service's own TxOutput
```
andreabadesso, line 66: It's a hacky solution to use the `isTokenHTR` and `getTokenIndex` from the Output class

Btw, can you please refactor line 71 to use the constant?

packages/wallet-service/src/config.ts
```diff
@@ -0,0 +1,76 @@
...
+    serviceName: process.env.SERVICE_NAME,
+    maxAddressGap: Number.parseInt(process.env.MAX_ADDRESS_GAP, 10),
+    voidedTxOffset: Number.parseInt(process.env.VOIDED_TX_OFFSET, 10),
+    blockRewardLock: Number.parseInt(process.env.BLOCK_REWARD_LOCK, 10),
+    confirmFirstAddress: process.env.CONFIRM_FIRST_ADDRESS === 'true',
+    wsDomain: process.env.WS_DOMAIN,
+    dbEndpoint: process.env.DB_ENDPOINT,
+    dbName: process.env.DB_NAME,
+    dbUser: process.env.DB_USER,
+    dbPort: parseInt(process.env.DB_PORT, 10),
```
andreabadesso:
```suggestion
    dbPort: Number.parseInt(process.env.DB_PORT, 10),
```

packages/wallet-service/src/config.ts
```diff
@@ -0,0 +1,76 @@
...
+    dbPass: process.env.DB_PASS,
+    redisUrl: process.env.REDIS_URL,
+    redisPassword: process.env.REDIS_PASSWORD,
+    authSecret: process.env.AUTH_SECRET,
+    walletServiceLambdaEndpoint: process.env.WALLET_SERVICE_LAMBDA_ENDPOINT,
+    pushNotificationEnabled: process.env.PUSH_NOTIFICATION_ENABLED === 'true',
+    pushAllowedProviders: process.env.PUSH_ALLOWED_PROVIDERS,
+    isOffline: process.env.IS_OFFLINE === 'true',
+    txHistoryMaxCount: parseInt(process.env.TX_HISTORY_MAX_COUNT || '50', 10),
+    healthCheckMaximumHeightDifference: Number(process.env.HEALTHCHECK_MAXIMUM_HEIGHT_DIFFERENCE ?? 5),
```
andreabadesso: `parseInt` as well?

packages/wallet-service/src/config.ts
```diff
@@ -0,0 +1,76 @@
...
+    firebaseAuthUri: process.env.FIREBASE_AUTH_URI,
+    firebaseTokenUri: process.env.FIREBASE_TOKEN_URI,
+    firebaseAuthProviderX509CertUrl: process.env.FIREBASE_AUTH_PROVIDER_X509_CERT_URL,
+    firebaseClientX509CertUrl: process.env.FIREBASE_CLIENT_X509_CERT_URL,
+    firebasePrivateKey: process.env.FIREBASE_PRIVATE_KEY,
+
+    maxLoadWalletRetries: parseInt(process.env.MAX_LOAD_WALLET_RETRIES || '5', 10),
+    logLevel: process.env.LOG_LEVEL || 'info',
+    createNftMaxRetries: parseInt(process.env.CREATE_NFT_MAX_RETRIES || '3', 10),
+    warnMaxReorgSize: parseInt(process.env.WARN_MAX_REORG_SIZE || '100', 10),
```
andreabadesso:
```suggestion
    maxLoadWalletRetries: Number.parseInt(process.env.MAX_LOAD_WALLET_RETRIES || '5', 10),
    logLevel: process.env.LOG_LEVEL || 'info',
    createNftMaxRetries: Number.parseInt(process.env.CREATE_NFT_MAX_RETRIES || '3', 10),
    warnMaxReorgSize: Number.parseInt(process.env.WARN_MAX_REORG_SIZE || '100', 10),
```

db/migrations/20250306170811-node_version_proxy.js
```diff
@@ -0,0 +1,83 @@
...
+        allowNull: false,
+        primaryKey: true,
+        defaultValue: 1,
+      },
+      timestamp: {
+        type: Sequelize.BIGINT.UNSIGNED,
+        allowNull: false,
+      },
+      data: {
+        type: Sequelize.TEXT,
```
andreabadesso, line 21: Not very relevant since we have validation in the code, but what do you think about using the JSON type?

https://dev.mysql.com/doc/refman/8.4/en/json.html

packages/wallet-service/src/mempool.ts
```diff
@@ -36,15 +37,15 @@ const mysql = getDbConnection();
 export const onHandleOldVoidedTxs = async (): Promise<void> => {
```
andreabadesso, line 37: This code should no longer exist, I'll create a PR removing it

packages/wallet-service/src/nodeConfig.ts
```diff
@@ -0,0 +1,53 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { getUnixTimestamp } from '@src/utils';
+import { ServerlessMysql } from 'serverless-mysql';
+import { getVersionData, updateVersionData } from '@src/db';
+import { FullNodeVersionData, FullNodeApiVersionResponse } from '@src/types';
+import fullnode from '@src/fullnode';
+
+const VERSION_CHECK_MAX_DIFF = 60 * 60; // 1 hour
```
andreabadesso, line 14: Use an env variable for this? Was it hardcoded before?

packages/wallet-service/src/nodeConfig.ts
```diff
@@ -0,0 +1,53 @@
+/**
```
andreabadesso, line 1: Add docstrings to methods in this file

packages/wallet-service/src/types.ts
```diff
@@ -39,8 +39,53 @@ export enum TxProposalStatus {
...
+ * wallet-service environment config.
+ */
+export interface EnvironmentConfig {
+  defaultServer: string;
+  stage: string;
+  network: string;
+  serviceName: string;
+  maxAddressGap: number;
+  voidedTxOffset: number;
+  blockRewardLock: number; // *
```
andreabadesso:
```suggestion
  blockRewardLock: number;
```

packages/wallet-service/src/utils.ts
```diff
@@ -15,54 +15,16 @@ import * as bitcoin from 'bitcoinjs-lib';
...
-  }
-}
-
-hathorLib.network.setNetwork(process.env.NETWORK);
-hathorLib.storage.setStore(new CustomStorage());
+hathorLib.network.setNetwork(config.network);
 
 const libNetwork = hathorLib.network.getNetwork();
 const hathorNetwork = {
   messagePrefix: '\x18Hathor Signed Message:\n',
```
andreabadesso, line 26: Maybe this should also be a config

andreabadesso: We can do it in another PR though

packages/wallet-service/src/ws/utils.ts
```diff
@@ -24,15 +25,15 @@ export const connectionInfoFromEvent = (
 ): WsConnectionInfo => {
   const logger: Logger = createDefaultLogger();
   const connID = event.requestContext.connectionId;
-  if (process.env.IS_OFFLINE === 'true') {
+  if (config.isOffline) {
```
andreabadesso, line 28: Wow, I don't remember what this is for

Can't we use `config.stage === 'local'` here as well?

packages/wallet-service/tests/api.test.ts
```diff
@@ -793,7 +798,8 @@ test('POST /wallet', async () => {
 
   // get the first address
   const xpubChangeDerivation = walletUtils.xpubDeriveChild(XPUBKEY, 0);
-  const firstAddress = walletUtils.getAddressAtIndex(xpubChangeDerivation, 0, process.env.NETWORK);
+  const firstAddressData = addressUtils.deriveAddressFromXPubP2PKH(xpubChangeDerivation, 0, process.env.NETWORK);
```
andreabadesso, line 800: Use the config to get the network?

packages/wallet-service/tests/api.test.ts
```diff
@@ -926,7 +932,8 @@ test('POST /wallet should fail with ApiError.WALLET_MAX_RETRIES when max retries
...
```
andreabadesso, line 934: same here

andreabadesso: and for all below

packages/wallet-service/tests/db.test.ts
```diff
@@ -3246,7 +3256,7 @@ describe('getPushDeviceSettingsList', () => {
 
     // register wallets that will not be queried
     const loadWallet = (eachDevice) => createWallet(mysql, eachDevice.walletId, XPUBKEY, AUTH_XPUBKEY, 5);
-    await devicesToLoad.forEach(loadWallet);
```
andreabadesso, line 3249: 😳

andreabadesso: Was this test even working?

packages/wallet-service/tests/txProposal.test.ts
```diff
@@ -203,24 +206,29 @@ test('POST /txproposals with too many outputs should fail with ApiError.TOO_MANY
...
+    genesis_tx2_hash: 'cafecafecafecafecafecafecafecafecafecafecafecafecafecafecafecafe',
+    native_token: { name: 'Hathor', symbol: 'HTR'},
   });
+  jest.resetModules();
 
   await addToWalletTable(mysql, [{
     id: 'my-wallet',
     xpubkey: 'xpubkey',
-    authXpubkey: 'auth_xpubkey',
+   authXpubkey: 'auth_xpubkey',
```
andreabadesso:
```suggestion
    authXpubkey: 'auth_xpubkey',
```

andreabadesso (changes requested):

packages/daemon/src/config.ts
r4mmer: Yes, but since we added typing on the wallet-lib we cannot allow `undefined` network.
So I had to force the network to be a string.

packages/wallet-service/src/nodeConfig.ts
r4mmer: It was [hardcoded](https://github.com/HathorNetwork/hathor-wallet-service/blob/master/packages/wallet-service/src/commons.ts#L75) but I can move to a configuration.

Obs: the old value was 1000 hours since we are comparing unix timestamps and not js timestamps.

packages/wallet-service/tests/api.test.ts
r4mmer: Did not seem necessary to refactor the tests to also use the config. 
What do you think?

packages/wallet-service/tests/db.test.ts
r4mmer: It would load the wallets, but I don't think the await was doing much here.

packages/wallet-service/src/ws/utils.ts
r4mmer: Sure, but there seems to be 2 stages for running locally.
- dev: https://github.com/HathorNetwork/hathor-wallet-service/blob/master/packages/wallet-service/src/api/wallet.ts#L95
- local: https://github.com/HathorNetwork/hathor-wallet-service/blob/master/packages/wallet-service/src/utils.ts#L132

Should we change these as well?

packages/wallet-service/tests/txProposal.test.ts
r4mmer: Done

packages/wallet-service/src/types.ts
r4mmer: Done

db/migrations/20250306170811-node_version_proxy.js
r4mmer: Done

db/migrations/20250306170811-node_version_proxy.js
r4mmer: From what I understand the advantages for the JSON data type are:
- JSON validation, which we already do in the code.
- Storing in a way that allows querying the content of the JSON object, which we do not require.

But do you think we should change it anyway?

packages/wallet-service/src/nodeConfig.ts
r4mmer: Done

packages/wallet-service/src/config.ts
r4mmer: Done

packages/daemon/src/utils/wallet.ts
r4mmer: Done

packages/wallet-service/src/config.ts
r4mmer: Done

packages/wallet-service/src/config.ts
r4mmer: Done

packages/wallet-service/package.json
r4mmer: Done

db/migrations/20250306170811-node_version_proxy.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up (queryInterface, Sequelize) {
+    // Dropping the table and recreating it is easier than deleting the fields then addind the new
```
pedroferreira1:
```suggestion
    // Dropping the table and recreating it is easier than deleting the fields then adding the new
```

packages/wallet-service/src/schemas.ts
```diff
@@ -0,0 +1,71 @@
...
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import Joi from 'joi';
+import { FullNodeApiVersionResponse, EnvironmentConfig } from '@src/types';
+
+export const FullnodeVersionSchema = Joi.object<FullNodeApiVersionResponse>({
```
pedroferreira1, line 11: Will it break the code in case we change anything in the full node `/version` API?

packages/common/src/utils/nft.utils.ts
```diff
@@ -8,10 +8,9 @@
 import { LambdaClient, InvokeCommand, InvokeCommandOutput } from '@aws-sdk/client-lambda';
 import { addAlert } from './alerting.utils';
 import { Severity } from '../types';
-// @ts-ignore
 import { Network, constants, CreateTokenTransaction, helpersUtils } from '@hathor/wallet-lib';
-// @ts-ignore
-import type { HistoryTransaction } from '@hathor/wallet-lib';
+// FIXME: import from lib path on HathorLib
+import type { HistoryTransaction } from '@hathor/wallet-lib/lib/models/types';
```
pedroferreira1, line 13: Why can't we do it?

r4mmer: The `HistoryTransaction` is not exported on the lib index but it is used here so it needs to be imported from the lib internal path.
Ideally, later we will export this type from the lib index and move this to import from `@hathor/wallet-lib`.

packages/wallet-service/src/schemas.ts
r4mmer: The `}).unknown(true);` at the end should allow for new keys to be added without breaking the validation.

db/migrations/20250306170811-node_version_proxy.js
r4mmer: Done

packages/wallet-service/tests/api.test.ts
pedroferreira1: Maybe to guarantee that the config is being filled correctly?

pedroferreira1 (dismissed):

packages/wallet-service/src/ws/utils.ts
andreabadesso: Yea, I will investigate and normalize this, nothing to do in this current PR

db/migrations/20250306170811-node_version_proxy.js
andreabadesso: I think it's ok to go with TEXT since we are just passing it through on http requests

packages/wallet-service/tests/env.test.ts
```diff
@@ -0,0 +1,25 @@
+import config, { loadEnvConfig } from '@src/config';
+
+test('Configuration should load correctly during tests', () => {
+  expect.hasAssertions();
+
+  const loadedConfig = loadEnvConfig();
+  expect(loadedConfig).toStrictEqual(config);
```
andreabadesso: Since you're not mocking `confirmFirstAddress`, this will depend on the `.env`, right?

Maybe just set it

`process.env.CONFIRM_FIRST_ADDRESS = 'true'`

andreabadesso (dismissed):

andreabadesso (approved):

pedroferreira1 (approved):

PR #219, opened 2025-03-11
fix: NFTs were not being properly processed

andreabadesso:
### Motivation

There was a bug in the NFT handler invocation logic. The `shouldInvokeNftHandlerForTx` method was failing for valid NFT transactions because the event data from the full node contained null values in the decoded field for some outputs. When the output data wasn't properly transformed to handle these null values (replacing them with empty objects {}), the NFT detection would fail silently, causing NFTs to not be properly processed.

### Acceptance Criteria
- Fix the bug where NFT transactions with null decoded fields aren't properly detected
- Ensure proper transformation of transaction data from full node events
- Extract the transformation and invocation logic into a reusable method
- Add comprehensive tests with real production data that includes a valid NFT transaction
- Verify that NFT detection works correctly with the transformed data

### Checklist
[X] If you are requesting a merge into master, confirm this code is production-ready and can be included in future releases as soon as it gets merged
[X] Make sure either the unit tests and/or the QA tests are capable of testing the new features
[X] Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. More dependencies increase the possibility of one of them being hijacked and affecting us.

### DIFF
--- a/packages/common/__tests__/utils/nft.utils.test.ts
+++ b/packages/common/__tests__/utils/nft.utils.test.ts
@@ -764,25 +764,17 @@ describe('processNftEvent', () => {
     // Disable NFT auto review
     process.env.NFT_AUTO_REVIEW_ENABLED = 'false';
 
-    // Real event data from production (simplified)
-    const eventData = {
-      hash: "000041f860a327969fa03685ed05cf316fc941708c53801cf81f426ac4a55866",
-      version: 2,
-      inputs: [{ tx_id: "tx1", index: 0, spent_output: { value: 1, token_data: 0, script: "s" } }],
-      outputs: [{ value: 1, token_data: 0, script: "s" }],
-      tokens: ["token1"],
-      token_name: "Test",
-      token_symbol: "TST",
-    };
+    // Use the real event data constant
+    const eventData = { ...REAL_NFT_EVENT_DATA };
 
     // Mock network
     const mockNetwork = { name: 'testnet' };
-
+    
     // Call the method
     const result = await NftUtils.processNftEvent(
-      eventData,
-      'test-stage',
-      mockNetwork as any,
+      eventData, 
+      'test-stage', 
+      mockNetwork as any, 
       logger
     );
 
@@ -791,7 +783,7 @@ describe('processNftEvent', () => {
 
     // Verify shouldInvokeNftHandlerForTx was NOT called
     expect(shouldInvokeSpy).not.toHaveBeenCalled();
-
+    
     // Verify the lambda was NOT invoked
     expect(invokeNftLambdaSpy).not.toHaveBeenCalled();
   });
@@ -802,25 +794,17 @@ describe('processNftEvent', () => {
     // Make shouldInvokeNftHandlerForTx return false
     shouldInvokeSpy.mockReturnValue(false);
 
-    // Simplified event data
-    const eventData = {
-      hash: "000041f860a327969fa03685ed05cf316fc941708c53801cf81f426ac4a55866",
-      version: 2,
-      inputs: [{ tx_id: "tx1", index: 0, spent_output: { value: 1, token_data: 0, script: "s" } }],
-      outputs: [{ value: 1, token_data: 0, script: "s" }],
-      tokens: ["token1"],
-      token_name: "Test",
-      token_symbol: "TST",
-    };
+    // Use the real event data constant
+    const eventData = { ...REAL_NFT_EVENT_DATA };
 
     // Mock network
     const mockNetwork = { name: 'testnet' };
-
+    
     // Call the method
     const result = await NftUtils.processNftEvent(
-      eventData,
-      'test-stage',
-      mockNetwork as any,
+      eventData, 
+      'test-stage', 
+      mockNetwork as any, 
       logger
     );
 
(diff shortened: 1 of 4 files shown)


packages/common/src/utils/nft.utils.ts
```diff
@@ -173,4 +185,103 @@ export class NftUtils {
...
+    stage: string,
+    network: unknown,
+    logger: Logger
+  ): Promise<boolean> {
+    // Early return if NFT auto review is disabled
+    if (!isNftAutoReviewEnabled()) {
+      logger.debug('NFT auto review is disabled. Skipping NFT handler invocation.');
+      return false;
+    }
+
```
pedroferreira1, line 205: suggestion(non-blocking): we could add a condition here to skip if it's not a tx to create token.

```
if (eventData.version !== constants.CREATE_TOKEN_TX_VERSION) return false;
```

pedroferreira1 (dismissed):

packages/common/src/utils/nft.utils.ts
andreabadesso: Done! Thanks

pedroferreira1 (approved):

pedroferreira1 (dismissed):

packages/common/src/utils/nft.utils.ts
```diff
@@ -173,4 +186,89 @@ export class NftUtils {
...
+    // Create a new object with the required properties
+    const transformedTx: any = {
+      hash: fullNodeData.hash,
+      tx_id: fullNodeData.hash, // Add tx_id for compatibility
+      version: fullNodeData.version,
+      tokens: fullNodeData.tokens,
+      token_name: fullNodeData.token_name,
+      token_symbol: fullNodeData.token_symbol,
+      inputs: fullNodeData.inputs.map((input: FullNodeInput) => {
+        const tokenIndex = (input.spent_output.token_data & constants.TOKEN_INDEX_MASK) - 1;
```
tuliomir, line 250: suggestion(non-blocking): Low level operations like bit manipulations are always best accompanied by a comment explaining them, as they are harder to read than common javascript code.

packages/common/src/utils/nft.utils.ts
```diff
@@ -173,4 +186,89 @@ export class NftUtils {
...
+      outputs: fullNodeData.outputs.map((output: FullNodeOutput) => {
+        const tokenIndex = (output.token_data & constants.TOKEN_INDEX_MASK) - 1;
+
+        return {
+          value: output.value,
+          token_data: output.token_data,
+          script: output.script,
+          token: tokenIndex < 0 ? constants.HATHOR_TOKEN_CONFIG.uid : fullNodeData.tokens[tokenIndex],
+          decoded: output.decoded || {},
+          spent_by: null,
```
tuliomir, line 274: question: Is there no possibility that an event may arrive with a spent token creation tx? Not even an update?

packages/common/__tests__/utils/nft.utils.test.ts
```diff
@@ -320,3 +473,503 @@ describe('minor helpers', () => {
...
+          value: 100,
+          script: 'script2',
+          decoded: {
+            type: 'P2PKH',
+            address: 'addr2'
+          }
+        }]
+      };
+
+      const txFromEvent = {
```
tuliomir, line 530: thought(non-blocking): I'm wondering if we should add the transformed tx hardcoded to the test here, instead of calculating programatically at test runtime.

I don't have an objective answer for this, so I'm only leaving the idea on the PR for now.

packages/common/__tests__/utils/nft.utils.test.ts
```diff
@@ -320,3 +473,503 @@ describe('minor helpers', () => {
...
+      expect(txFromEvent.outputs[0].decoded).toStrictEqual({});
+
+      // Mock network for validation
+      const mockNetwork = {
+        name: 'testnet',
+      };
+
+      // Test that this transaction is detected as an NFT
+      expect(isNftTransactionSpy).toHaveBeenCalledTimes(0);
+      const shouldInvoke = NftUtils.shouldInvokeNftHandlerForTx(txFromEvent, mockNetwork as any, logger);
```
tuliomir: suggestion(non-blocking): Avoid using `any`, as it's prohibited by many standard typescript linting rules.
```suggestion
      const shouldInvoke = NftUtils.shouldInvokeNftHandlerForTx(txFromEvent, mockNetwork as unknown as Network, logger);
```

tuliomir (dismissed):

packages/common/src/utils/nft.utils.ts
andreabadesso: No, the fullnode sends events in the order they're processed

andreabadesso: I think it's irrelevant in this context, we're just transforming the tx here so it's processed properly by the wallet-lib's method

packages/common/src/utils/nft.utils.ts
andreabadesso: Done, thanks

packages/common/__tests__/utils/nft.utils.test.ts
andreabadesso: I don't like it, way too verbose for me with no performance gains

packages/common/__tests__/utils/nft.utils.test.ts
andreabadesso: Done, thanks!

tuliomir (approved):

pedroferreira1 (approved):

PR #222, opened 2025-03-25
chore: upgrade wallet-lib to v2.5.1

r4mmer:
### Acceptance Criteria

- Upgrade wallet-lib version to 2.5.1

### Checklist


packages/daemon/src/utils/cache.ts
```diff
@@ -43,7 +43,7 @@ export class LRU {
   }
 
   first(): string {
-    return this.cache.keys().next().value;
+    return this.cache.keys().next()?.value ?? '';
```
andreabadesso, line 46: Why?

packages/daemon/src/actors/WebSocketActor.ts
```diff
@@ -68,13 +69,13 @@ export default (callback: any, receive: any) => {
   };
 
   socket.onmessage = (socketEvent) => {
-    const event = JSON.parse(socketEvent.data.toString());
+    let event = bigIntUtils.JSONBigInt.parse(socketEvent.data.toString());
```
pedroferreira1: Why did you change to `let`?

packages/daemon/src/types/event.ts
```diff
@@ -120,7 +120,7 @@ export interface EventTxInput {
 }
 
 export interface EventTxOutput {
-  value: number;
+  value: bigint;
```
pedroferreira1, line 123: Is this a full node event? Does the `value` come as string?

packages/wallet-service/src/db/utils.ts
```diff
@@ -152,7 +152,10 @@ export class FromTokenBalanceMapToBalanceValueList {
 }
 
 export const sortBalanceValueByAbsTotal = (balanceA: TokenBalanceValue, balanceB: TokenBalanceValue): number => {
-  if (Math.abs(balanceA.total) - Math.abs(balanceB.total) >= 0) return -1;
+  function abs(num: bigint) {
+    return num >= 0n ? num : -num;
+  }
```
pedroferreira1, line 157: You had to create this method twice already. Maybe we could have a util `bigIntAbs` for this. What do you think?

packages/wallet-service/tests/utils/pushnotification.utils.test.ts
```diff
@@ -12,6 +12,8 @@ import { sendMock, lambdaInvokeCommandMock } from '@tests/utils/aws-sdk.mock';
 import { LambdaClient } from '@aws-sdk/client-lambda';
 import { buildWalletBalanceValueMap } from '@tests/utils';
```
andreabadesso:
```suggestion
```

packages/wallet-service/src/db/index.ts
```diff
@@ -2389,7 +2390,7 @@ export const filterTxOutputs = async (
     ignoreLocked: false,
     skipSpent: true,
     biggerThan: -1,
```
pedroferreira1: You changed the default in the other file in the Joi schema to `0n`. I think it doesn't affect anything but it would be good to have both the same.

packages/wallet-service/src/types.ts
```diff
@@ -446,7 +447,7 @@ export interface TxTokenBalance {
   txId: string;
   timestamp: number;
   voided: boolean;
-  balance: Balance;
+  balance: bigint;
```
pedroferreira1, line 450: Was it a bug before? `Balance` has some attributes.

packages/wallet-service/src/types.ts
```diff
@@ -745,7 +746,7 @@ export interface TxByIdToken {
   version: number;
   voided: boolean;
   weight: number;
-  balance: Balance;
+  balance: bigint;
```
pedroferreira1, line 749: Same as the other comment about the `Balance` type.

packages/daemon/src/utils/cache.ts
r4mmer: `this.cache.keys().next()` may be undefined if there is no next (cache is empty), so we need `?.value` to safely get the value (otherwise we would get an error at runtime trying to access `value` from `undefined`).

In the case of next actually being undefined, the `?? ''` ensures the return is a string so the type is consistent.

packages/daemon/src/types/event.ts
r4mmer: You are correct, I was changing instances of value to bigint and did not consider how the data is generated.
This comes from the reliable integration, so it is a json event from a fullnode websocket, maybe we should create zod schemas to parse and validate the event data.

packages/wallet-service/src/db/utils.ts
r4mmer: I did create 2 instances of this method, but both of these were inside a `sortBalanceValueByAbsTotal` which is a method that is duplicated on the wallet-service and daemon packages.

This issue is because some utilities should be moved to the `common` package to avoid duplicating code.
So I think we should move the `sortBalanceValueByAbsTotal` instead of the internal `abs` and this seems a little outside the scope of this PR which is already huge.

What do you think?

packages/wallet-service/src/types.ts
r4mmer: Yes, on `packages/wallet-service/src/db/index.ts` the method `getWalletTxHistory` is the only place that builds an instance of `TxTokenBalance`.

Previously this method would populate the balance with `SELECT wallet_tx_history.balance AS balance, ...` then type cast it as Balance with `<Balance>result.balance` but it was actually a number.
Since this would only be used by the api to `JSON.stringify` into the history response the number value would make it work and the type system would not break because the type cast to `Balance` would match the `TxTokenBalance` type.

Now however I changed the configuration of the db client to return BITINT columns as strings.
So all BIGINT columns, values, balances, etc should be treated as strings and then converted to bigint, which is why I changed the typing to bigint and the api now uses `JSONBigInt.stringify`

packages/wallet-service/src/types.ts
r4mmer: This is the exact same issue but with the `TxByIdToken` interface, used by `getTransactionById` method which is only used on 1 api and the result is passed to `JSON.stringify` without accessing any field.

Curiously there are other `getTransactionById` methods but these are not the same and do not use the `TxByIdToken` interface.

packages/wallet-service/src/db/index.ts
r4mmer: Changed here as well

packages/daemon/src/actors/WebSocketActor.ts
r4mmer: Changed back into const

packages/wallet-service/tests/utils/pushnotification.utils.test.ts
r4mmer: Done

packages/wallet-service/src/db/utils.ts
pedroferreira1: I agree. Can you create an issue in this repo with this discussion so we can improve this?

packages/daemon/src/types/event.ts
pedroferreira1: Makes sense to do it in another PR. Can you create an issue so we can track this?

pedroferreira1 (approved):

packages/wallet-service/src/db/utils.ts
r4mmer: Created [issue](https://github.com/HathorNetwork/hathor-wallet-service/issues/265)

packages/daemon/src/types/event.ts
r4mmer: Created the issue https://github.com/HathorNetwork/hathor-wallet-service/issues/266 to track this

andreabadesso (approved):

PR #228, opened 2025-04-02
chore: better eslint

andreabadesso:
## Motivation

The motivation for the changes in this PR is to fix ESLint issues in editors by ensuring each project gets the correct ESLint configuration. Currently, ESLint is failing because it is unable to correctly parse TypeScript files for each project.

## Acceptance Criteria
- Ensure ESLint correctly parses TypeScript files for the common, daemon, and wallet-service packages.
- Update the Nix flake configuration to properly inherit nodePackages and configure the development shell.
- Add TypeScript dependency to the project root
- Introduce a new tsconfig.json in the root importing packages tsconfigs

## Checklist
- [X] If you are requesting a merge into master, confirm this code is production-ready and can be included in future releases as soon as it gets merged.
- [X] Make sure either the unit tests and/or the QA tests are capable of testing the new features.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -12,7 +12,7 @@
       overlays.default = final: prev:
         let
           packages = self.packages.${final.system};
-          inherit (packages) node-packages;
+          inherit (packages) nodePackages;
         in
         {
           nodejs = final.nodejs_20;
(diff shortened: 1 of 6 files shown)


r4mmer (approved):

flake.nix
```diff
@@ -12,32 +12,40 @@
...
           packages = with pkgs; [
             nixpkgs-fmt
             nodejs_20
             yarn
             docker-compose
           ];
+          devshell = {
+            startup = {
+              setup.text = ''
+                export PATH="$PWD/node_modules/.bin:$PWD/packages/daemon/node_modules/.bin:$PWD/packages/wallet-service/node_modules/.bin:$PATH"
```
tuliomir, line 44: question(non-blocking): While I think adding the `.bin` folder to the path is a smart move, maybe it relies too much on the internal implementation of NodeJS.

What is our actual need for this export? Couldn't we just replace our usage of those binaries with `npx myDependency`, since `npx` is available [since npm 8](https://docs.npmjs.com/cli/v8/commands/npx)?

tuliomir (approved):

flake.nix
andreabadesso: Yea, the idea here is to have `typescript` available in the path, so the LSP from the dev editor can use it

This it good so we don't have to install `typescript` globally (using npm)

Also, we are using `yarn` in this repository, so we shouldn't use `npm` here

PR #229, opened 2025-04-02
fix: io decoded handling

andreabadesso:
### Motivation

After https://github.com/HathorNetwork/hathor-core/pull/1250, the `decoded` key of `output` and `spent_output` (inputs) is now returning an empty object when the script is unable to be parsed, causing our code to fail because we were expecting it to be null.

### Acceptance Criteria

- We should properly handle empty objects in `decoded`
- We should have a mechanism for checking if all required keys are available in the `decoded` object.

### Checklist
- [X ] Make sure either the unit tests and/or the QA tests are capable of testing the new features

### DIFF
--- a/packages/daemon/src/services/index.ts
+++ b/packages/daemon/src/services/index.ts
@@ -29,6 +29,7 @@ import {
   Transaction,
   TokenBalanceMap,
   TxOutputWithIndex,
+  isDecodedValid,
 } from '@wallet-service/common';
 import {
   prepareOutputs,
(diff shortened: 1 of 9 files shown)


r4mmer (dismissed):

packages/daemon/src/services/index.ts
```diff
@@ -238,7 +239,7 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
       const blockRewardOutput = outputs[0];
 
       // add miner to the miners table
-      if (blockRewardOutput.decoded) {
+      if (isDecodedValid(blockRewardOutput.decoded)) {
```
pedroferreira1: `decoded.address` is required here, right?

pedroferreira1 (dismissed):

packages/daemon/src/services/index.ts
andreabadesso: Done, thanks

r4mmer (approved):

pedroferreira1 (approved):

PR #235, opened 2025-04-04
fix: address invalid parameter

andreabadesso:
### Motivation

The motivation for these changes was to correct the handling of the `index` parameter in the `/addresses` API endpoint.

Previously, the `index` was incorrectly fetched from path parameters, causing all requests with `index` to be ignored and a list of addresses instead of the queried one

---

### Acceptance Criteria

- The `/addresses` endpoint should accept the `index` parameter as a query string parameter, not as a path parameter.
- If `index` is omitted, all addresses for the wallet should be returned.
- If a valid `index` is provided, only the address at that index should be returned.
- The serverless configuration should reflect the new query string parameter for `index`

---

### Checklist

### DIFF
--- a/packages/wallet-service/package.json
+++ b/packages/wallet-service/package.json
@@ -6,7 +6,7 @@
     "lint": "eslint src/ tests/ --ext .js,.jsx,.ts,.tsx src tests",
     "lint-fix": "eslint src/ tests/ --fix --ext .js,.jsx,.ts,.tsx src tests",
     "check-types": "tsc --noemit --skipLibCheck",
-    "test": "jest -u"
+    "test": "jest"
   },
   "author": "Hathor Labs",
   "license": "MIT",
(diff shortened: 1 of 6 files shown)


packages/wallet-service/serverless.yml
```diff
@@ -324,10 +324,6 @@ functions:
           method: get
           cors: true
           authorizer: ${self:custom.authorizer.walletBearer}
-          request:
-            parameters:
-              paths:
-                index: false
```
andreabadesso, line 330: We would only need this block if we wanted to enforce the presence of a specific query in api gateway, which is not the case

packages/wallet-service/package.json
```diff
@@ -6,7 +6,7 @@
     "lint": "eslint src/ tests/ --ext .js,.jsx,.ts,.tsx src tests",
     "lint-fix": "eslint src/ tests/ --fix --ext .js,.jsx,.ts,.tsx src tests",
     "check-types": "tsc --noemit --skipLibCheck",
-    "test": "jest"
+    "test": "jest -u"
```
r4mmer: We will always update the snapshots?

pedroferreira1 (dismissed):

packages/wallet-service/package.json
andreabadesso: No we shouldn't!

Fixed, thanks

r4mmer (approved):

pedroferreira1 (approved):

PR #245, opened 2025-04-11
fix: firebase env variables should be optional

andreabadesso:
### Motivation

The Firebase configuration fields were previously marked as required in the schema, which meant that all environments needed to have these fields configured, even if they didn't use push notifications. This was problematic because:

1. Not all environments need Firebase/push notifications
2. The validation was happening at module load time, which meant the application would fail to start even if push notifications were disabled
3. We were using a separate validation mechanism (`assertEnvVariablesExistence`) instead of reusing our Joi schema system

### Acceptance Criteria

- [x] Make Firebase configuration fields optional in the schema
- [x] Validate Firebase configuration only when push notifications are enabled
- [x] Use Joi schema validation instead of manual environment variable checking
- [x] Update tests to verify the new validation behavior
- [x] Ensure environments without Firebase can run without these fields
- [x] Maintain proper validation when push notifications are enabled

### Checklist
- [x] Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones


packages/wallet-service/src/utils/pushnotification.utils.ts
```diff
@@ -12,37 +12,23 @@ import fcmAdmin, { credential, messaging, ServiceAccount } from 'firebase-admin'
 import { MulticastMessage } from 'firebase-admin/messaging';
 import createDefaultLogger from '@src/logger';
 import config from '@src/config';
-import { assertEnvVariablesExistence } from '@wallet-service/common/src/utils/index.utils';
```
luislhl, line 15: This is not using anywhere else, apparently. Should we remove its code?

luislhl (approved):

r4mmer (approved):

PR #261, opened 2025-05-30
feat: token version

raul-oliveira:
### Motivation
This PR is part of the Dynamic transaction model implementation.

According to the [docs](https://github.com/HathorNetwork/rfcs/pull/94), tokens can be created with two versions:
> 
> # Token info version
> Given Hathor's [Anatomy of a Transaction RFC](https://github.com/HathorNetwork/rfcs/blob/master/text/0015-anatomy-of-tx.md#token-creation), it is reasonable to suggest that the byte used by the token creation transaction token_info_version will be used to determine fee-created tokens.
>
> Since each custom token id is the hash of the token creation transaction that created it, we can assume the enum values below can be assigned to the token_info_version byte in the token creation tx and then we can retrieve it.
>
> So, by adding a TokenInfoVersion enum we have:
> 
> DEPOSIT = 1 (as is)
> FEE = 2

The `wallet-service` is used by `wallet-lib`, a project that expect the `version` field to be available on the `getTokenDetails` API.

### Acceptance Criteria

- Add the version column in the database with a migration
- Update the `TokenInfo` entity to handle the `version`
  - Add to sync-daemon workspace
  - Add to wallet-service workspace
- Update some test utils functions to handle the version field
- Update the sql statements that brings the token details
- Add the version field in the `getTokenInformation` API method.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,14 +7,33 @@ Refer to https://github.com/HathorNetwork/rfcs/blob/master/projects/wallet-servi
 ### Local environment
 
 #### System dependencies
+```
+Node: 20x
+yarn: v4 (yarn-berry)
+```
 
-You need nodejs installed on your enviroment, we suggest the latest Active LTS version (v18.x.x).
+#### Install nix (preferred)
+
+For a better developer experience we suggest nix usage for mananing the enviroment. Visit this [link](https://nixos.org/download/#download-nix) to download it.
+
+To enable the commands `nix develop` and `nix build` using flakes, add the following to your `/etc/nix/nix.conf` file:
+
+```
+experimental-features = nix-command flakes
+```
 
 #### Clone the project and install dependencies
-
-`git clone https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon.git`
-
-`npm install`
+```sh
+$ git clone https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon.git
+```
+To initialize nix dev environment:
+```sh
+$ nix develop
+```
+then, install the depencies: 
+```sh
+yarn
+```
 
 #### Add env variables or an .env file to the repository:
 
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ Refer to https://github.com/HathorNetwork/rfcs/blob/master/projects/wallet-servi
 ### Local environment
 
 #### System dependencies
+
 ```
 Node: 20x
 yarn: v4 (yarn-berry)
@@ -23,14 +24,19 @@ experimental-features = nix-command flakes
 ```
 
 #### Clone the project and install dependencies
+
 ```sh
 $ git clone https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon.git
 ```
+
 To initialize nix dev environment:
+
 ```sh
 $ nix develop
 ```
-then, install the depencies: 
+
+then, install the depencies:
+
 ```sh
 yarn
 ```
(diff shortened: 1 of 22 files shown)


README.md
```diff
@@ -8,13 +8,38 @@ Refer to https://github.com/HathorNetwork/rfcs/blob/master/projects/wallet-servi
...
+
+```
+experimental-features = nix-command flakes
+```
 
 #### Clone the project and install dependencies
 
-`git clone https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon.git`
+```sh
+$ git clone https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon.git
```
andreabadesso, line 29: Can you replace this with

`git@github.com:HathorNetwork/hathor-wallet-service.git`?

README.md
```diff
@@ -54,6 +79,16 @@ AWS_SECRET_ACCESS_KEY="..."
 
 These are used for communicating with the alert SQS
 
+#### Docker images
+
+Some packages depends on some docker images. To build them you'll need to have Hathor VPN access configured, check this [link](https://github.com/HathorNetwork/ops-tools/blob/master/terraform/wireguard-vpn/SOP.md#adding-a-new-client-to-the-vpn) for it.
```
andreabadesso, line 84: Which ones?

packages/wallet-service/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "composite": true,
```
andreabadesso, line 3: What is this for?

packages/daemon/src/types/token.ts
```diff
@@ -5,9 +5,26 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { constants } from '@hathor/wallet-lib';
+import { constants } from "@hathor/wallet-lib";
```
andreabadesso, line 8: Why?

packages/wallet-service/src/types.ts
```diff
@@ -164,7 +164,23 @@ export interface TokenBalance {
...
+  name: string;
+  symbol: string;
+  version?: TokenInfoVersion | null;
+}
+
+export interface ITokenInfoOptions extends ITokenInfo {
+  transactions?: number
+}
+
+export class TokenInfo implements ITokenInfo {
```
andreabadesso, line 183: Do we have those types in the wallet-lib?

raul-oliveira: Yes, I'm introducing them in a new PR that should be merged before this.
The types we have there: TokenInfoVersion, ITokenInfo (With a different name)

package.json
```diff
@@ -7,7 +7,7 @@
     "packages/wallet-service"
   ],
   "engines": {
-    "node": ">=18"
+    "node": ">=20"
```
r4mmer, line 10: Maybe we should upgrade to 22, since the current lib version already requires node 22

packages/daemon/src/types/db.ts
```diff
@@ -119,6 +119,7 @@ export interface TokenInformationRow extends RowDataPacket {
   name: string;
   symbol: string;
   transactions: number;
+  version?: number | null;
```
r4mmer, line 122: Why null?

db/migrations/20250529233113-add-token-version.js
```diff
@@ -0,0 +1,15 @@
+"use strict";
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    queryInterface.addColumn("token", "version", {
+      type: Sequelize.INTEGER,
+      allowNull: true,
```
r4mmer, line 8: Why allow null?
We could create the column with a default value of 0 (or whatever value the `tokenVersion.DEPOSIT` is) so that all "current listed" tokens will be marked as DBTs (Deposit Based Token).

What do you think?

PR #263, opened 2025-07-14
feat: process nanocontract transactions and nc_addresses

r4mmer:
### Motivation

To maintain complete transaction history for wallets and addresses, we need to adapt our indexing service to handle these new transactions.

### Acceptance Criteria

- Receive nano header from fullnode events
- Add the transaction to the `nc_address` history
- Clean the transaction from the `nc_address` history when the tx is voided

### Checklist


packages/daemon/src/db/index.ts
```diff
@@ -776,7 +776,7 @@ export const getAddressWalletInfo = async (mysql: MysqlConnection, addresses: st
   }
 
   const addressWalletMap: StringMap<Wallet> = {};
-  const [results] = await mysql.query(
+  const [results, _]: [mysql.RowDataPacket[], mysql.FieldPacket[]] = await mysql.query(
```
andreabadesso: You can create a type for the rows and do:

```suggestion
  const [results, _]: [mysql.RowDataPacket[], mysql.FieldPacket[]] = await mysql.query<AddressWalletInfoRow>(
```

There are some examples in this file

packages/daemon/src/services/index.ts
```diff
@@ -164,6 +165,16 @@ export const metadataDiff = async (_context: Context, event: Event) => {
 export const isBlock = (version: number): boolean => version === hathorLib.constants.BLOCK_VERSION
   || version === hathorLib.constants.MERGED_MINED_BLOCK_VERSION;
 
+export function isNanoContract(headers: EventTxHeader[]) {
+  for (const header of headers) {
+    // NanoHeader id should always be 0x10
+    if (header.id === '10') {
```
andreabadesso: I don't get it, why `'10'`? Wouldn't it be `'16'`? Or `'0x10'`?

packages/daemon/src/utils/wallet.ts
```diff
@@ -161,6 +158,25 @@ export const getAddressBalanceMap = (
     addressBalanceMap[address] = TokenBalanceMap.merge(addressBalanceMap[address], tokenBalanceMap);
   }
 
+  for (const header of headers) {
+    if (header.id !== '10') {
+      // We currently only handle nano contract headers
+      continue;
```
andreabadesso: Transform the `isNanoContract` into a util and use it?

andreabadesso (changes requested):

packages/daemon/src/db/index.ts
```diff
@@ -1507,25 +1506,24 @@ export const cleanupVoidedTx = async (mysql: MysqlConnection, txId: string): Pro
...
+): Promise<StringMap<string>> => {
+  if (tokenIdList.length === 0) return {};
 
   const [results] = await mysql.query<TokenSymbolsRow[]>(
     'SELECT `id`, `symbol` FROM `token` WHERE `id` IN (?)',
     [tokenIdList],
   );
 
-  if (results.length === 0) return null;
+  if (results.length === 0) return {};
```
pedroferreira1, line 1526: Why did you change this return?

packages/daemon/src/services/index.ts
pedroferreira1: In the core is `NANO_HEADER = b'\x10'` and the events send in hex

```
>>> b'\x10'.hex()
'10'
```

I just thought we were exporting this type from the wallet-lib, but I checked and we are not. I think we should create an issue for that. We should export that from the lib, so we don't need to use the hardcoded value here.

pedroferreira1 (commented):
It's missing the seqnum handling

packages/daemon/src/db/index.ts
r4mmer: We use the return of this function on other methods ignoring the possible `null` value with a `ts-ignore` comment.

To solve this we could just treat the possible `null` everytime we call this function or make the return type consistent.
I chose to always return a `StringMap<string>` even if empty so the caller don't have to treat the `null` variant.

packages/daemon/src/utils/wallet.ts
r4mmer: Done

packages/daemon/src/db/index.ts

pedroferreira1 (approved):

andreabadesso (approved):

PR #268, opened 2025-07-23
Feat/nc seqnum tracking

r4mmer:
### Motivation

We need to track the seqnum of each address when calling a nano contract transaction.
We also need to expose this value to the client as an address metadata.

### Acceptance Criteria

- Update the address seqnum with each new nano contract transaction.
- Create the `/address/info` api to expose the address metadata.

### Checklist

### DIFF
--- a/packages/daemon/src/db/index.ts
+++ b/packages/daemon/src/db/index.ts
@@ -20,6 +20,7 @@ import {
   TokenSymbolsRow,
   MaxAddressIndexRow,
   AddressesWalletsRow,
+  AddressRow,
 } from '../types';
 import {
   TxInput,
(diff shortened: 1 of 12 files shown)


packages/daemon/src/db/index.ts
```diff
@@ -1584,3 +1585,36 @@ export const getMaxIndicesForWallets = async (
     }
   ]));
 };
+
+export async function getAddressInfo(mysql: MysqlConnection, address: string) {
```
andreabadesso: Docstrings for this and the next method

packages/daemon/src/services/index.ts
```diff
@@ -273,6 +275,19 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
         throw new Error('Processed a token creation event but it did not come with token name and symbol');
       }
       await storeTokenInformation(mysql, hash, token_name, token_symbol);
+    } else {
```
pedroferreira1: Why in the `else`? We might have a nano header in a `CREATE_TOKEN_TX_VERSION`

packages/daemon/src/db/index.ts

packages/daemon/src/services/index.ts
r4mmer: Done

packages/daemon/src/db/index.ts
```diff
@@ -1584,3 +1585,57 @@ export const getMaxIndicesForWallets = async (
...
+  return results[0];
+}
+
+/**
+ * Get an address seqnum.
+ *
+ * @param mysql - Database connection
+ * @param address - which address to fetch information from
+ */
+export async function getAddressSeqnum(mysql: MysqlConnection, address: string) {
```
pedroferreira1: Missing return type

pedroferreira1 (approved):

packages/daemon/src/services/index.ts
```diff
@@ -433,6 +435,18 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
         .catch((err: unknown) => logger.error('[ALERT] Error processing NFT event', err));
     }
 
+    // Need to check if there is a nano header and update the nc_address's seqnum if needed
```
pedroferreira1, line 438: Can we get here in a voided tx?

pedroferreira1 (dismissed):

andreabadesso (dismissed):

packages/daemon/src/db/index.ts
r4mmer: Added

packages/daemon/src/services/index.ts
r4mmer: Created a guard against voided transactions

andreabadesso (approved):

pedroferreira1 (approved):

PR #274, opened 2025-07-30
fix: support empty object for decoded output in full node events

pedroferreira1:
### Motivation

There's a case when the full node sends an event where the `decoded` value of the output is an empty object (`{}`) and we were supporting only `null` or the full object.

### Acceptance Criteria

- Support typing of empty object for the decoded output.

### Checklist

### DIFF
--- a/packages/common/src/utils/wallet.utils.ts
+++ b/packages/common/src/utils/wallet.utils.ts
@@ -6,6 +6,7 @@
  */
 
 import { constants } from '@hathor/wallet-lib';
+import { DecodedOutput } from '../types';
 
 /**
  * Checks if a given tokenData has any authority bit set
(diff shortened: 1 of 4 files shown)


packages/common/src/utils/wallet.utils.ts
```diff
@@ -25,7 +26,7 @@ export const isAuthority = (tokenData: number): boolean => (
  * @param requiredKeys - A list of keys to check
  * @returns true if the decoded object is valid, false otherwise
  */
-export const isDecodedValid = (decoded: any, requiredKeys: string[] = []): boolean => {
+export const isDecodedValid = (decoded: any, requiredKeys: string[] = []): decoded is DecodedOutput => {
```
pedroferreira1, line 29: I need this for the linter to understand that is this method returns `true`, the decoded value will have the keys in the `DecodedOutput` type.

andreabadesso (approved):

tuliomir (approved):

PR #281, opened 2025-08-21
fix: we should unspend transaction tx outputs when the tx spending it…

andreabadesso:
### Acceptance Criteria

- We should properly "unspend" utxos spent by voided transactions
- We should "unlock" utxos that are locked by a given `txproposal` in case this `txproposal` fails.
- We should properly recalculate `wallet` balances when a transaction is voided

IMPORTANT: In this PR we are (1) using a experimental docker image for the new tests and (2) skipping genesis address balance check

(1) This will be fixed when https://github.com/HathorNetwork/hathor-core/pull/1392 is merged
(2) This is a race condition, tests pass when running individually, we must debug and fix it in a later PR

### Checklist


packages/daemon/package.json
```diff
@@ -21,7 +21,7 @@
     "test_images_wait_for_db": "yarn dlx ts-node ./__tests__/integration/scripts/wait-for-db-up.ts",
     "test_images_wait_for_ws": "yarn dlx ts-node ./__tests__/integration/scripts/wait-for-ws-up.ts",
     "test_images_setup_database": "yarn dlx ts-node ./__tests__/integration/scripts/setup-database.ts",
-    "test": "jest --coverage",
+    "test": "jest --coverage --runInBand",
```
andreabadesso, line 24: Reduce flakiness

packages/wallet-service/src/api/txProposalSend.ts
```diff
@@ -140,6 +141,8 @@ export const send: APIGatewayProxyHandler = middy(walletIdProxyHandler(async (wa
       TxProposalStatus.SEND_ERROR,
     );
 
+    await releaseTxProposalUtxos(mysql, [txProposalId]);
```
andreabadesso, line 144: outputs would stay "stuck" until the next release routine (every 5 min)

packages/daemon/src/types/event.ts
```diff
@@ -103,7 +103,11 @@ export const EventTxOutputSchema = z.object({
       type: z.string(),
       address: z.string(),
       timelock: z.number().nullable(),
+      token_data: z.number().optional(),
     }).nullable(),
+    z.object({
+      token_data: z.number(),
+    }),
```
pedroferreira1: @r4mmer did this in a different way: https://github.com/HathorNetwork/hathor-wallet-service/pull/287/files

Maybe this one is better

packages/daemon/__tests__/integration/balances.test.ts
```diff
@@ -118,13 +124,14 @@ describe('unvoided transaction scenario', () => {
     // @ts-expect-error
     await transitionUntilEvent(mysql, machine, UNVOIDED_SCENARIO_LAST_EVENT);
     const addressBalances = await fetchAddressBalances(mysql);
-    expect(validateBalances(addressBalances, unvoidedScenarioBalances));
+    await validateBalances(addressBalances, unvoidedScenarioBalances);
```
tuliomir: chore: We should have at least one assertion per test.

In the case of the tests in this file, maybe we should just expect the `validateBalances()` method not to throw in each one of them.

packages/daemon/src/services/index.ts
```diff
@@ -537,6 +544,54 @@ export const voidTx = async (
   await voidTransaction(mysql, hash, addressBalanceMap);
   await markUtxosAsVoided(mysql, dbTxOutputs);
 
+
+  // CRITICAL: Unspent the inputs when voiding a transaction
```
tuliomir: chore: typo. This is also present in other points in this PR.
```suggestion
  // CRITICAL: Unspend the inputs when voiding a transaction
```

packages/wallet-service/tests/txProposalUtxoUnlock.test.ts
```diff
@@ -0,0 +1,305 @@
...
+
+    // Verify send failed and proposal status is SEND_ERROR
+    expect(sendResult.statusCode).toBe(400);
+    const sendBody = JSON.parse(sendResult.body as string);
+    expect(sendBody.success).toBe(false);
+
+    const txProposal = await getTxProposal(mysql, txProposalId);
+    expect(txProposal!.status).toBe(TxProposalStatus.SEND_ERROR);
+
+    // BUG: UTXO should be released when send fails, but currently it remains locked
```
tuliomir: question: Wasn't this bug fixed by https://github.com/HathorNetwork/hathor-wallet-service/pull/281/commits/865a58f3af4b7a5c596b7b2dcaf96dd82726c6dc , on the `api/txProposalSend.ts` file?

packages/daemon/__tests__/services/services_with_db.test.ts
```diff
@@ -62,3 +107,700 @@ describe('handleVoidedTx (db)', () => {
...
+      script: 'dqkUH70YjKeoKdFwMX2TOYvGVbXOrKaIrA==',
+    }];
+
+    await voidTx(mysql, txIdB, inputs, outputs, [tokenId], []);
+
+    // Check if the UTXO from transaction A is unspent again
+    utxo = await getTxOutput(mysql, txIdA, 0, false);
+    expect(utxo).not.toBeNull();
+    expect(utxo?.spentBy).toBeNull();
+  });
```
tuliomir, line 168: question: Should we check if utxo B was also removed, like the test below does?
```ts
// The output from transaction B should be voided (not accessible with getTxOutput)
const utxoB = await getTxOutput(mysql, txIdB, 0, false);
expect(utxoB).toBeNull(); // Should be null because it's voided
```

packages/daemon/__tests__/services/services_with_db.test.ts
```diff
@@ -62,3 +107,700 @@ describe('handleVoidedTx (db)', () => {
...
+    expect(utxoC).toBeNull(); // Should be null because it's voided
+  });
+
+  it('should handle voiding a transaction that spends already voided outputs', async () => {
+    expect.hasAssertions();
+
+    // Create transaction A that creates an output
+    const txIdA = 'test3-tx-a';
+    const txIdB = 'test3-tx-b'; // Will be voided first
+    const txIdC = 'test3-tx-c'; // Will be voided second
```
tuliomir: chore: Tx `C` is actually voided first in this test.

packages/daemon/__tests__/services/services_with_db.test.ts
```diff
@@ -62,3 +107,700 @@ describe('handleVoidedTx (db)', () => {
...
+
+    await voidTx(mysql, txIdB, inputs, outputs, [tokenId], []);
+
+    // CRITICAL BUG: Check wallet balance after voiding
+    walletBalance = await getWalletBalance(mysql, walletId, tokenId);
+    expect(walletBalance).not.toBeNull();
+
+    // This will FAIL because wallet balances are not updated during voiding
+    // The transaction count should decrease from 2 to 1
+    expect(walletBalance.transactions).toBe(1); // Should be back to 1 transaction
```
tuliomir: chore: This test is passing correctly. We should update these comments

tuliomir (commented):
The tests on the Daemon are currently not exiting properly. They all pass below 24s but the task ultimately fails due to timeout.

packages/daemon/src/db/index.ts
```diff
@@ -478,6 +486,98 @@ export const voidTransaction = async (
...
+  for (const [walletId, tokenMap] of Object.entries(walletBalanceMap)) {
+    for (const [token, tokenBalance] of tokenMap.iterator()) {
+      // Update wallet_balance table by reversing the transaction's impact
+      await mysql.query(
+        `UPDATE \`wallet_balance\`
+         SET total_received = total_received - ?,
+             unlocked_balance = unlocked_balance - ?,
+             locked_balance = locked_balance - ?,
+             transactions = transactions - 1,
+             unlocked_authorities = (unlocked_authorities | ?),
```
pedroferreira1, line 566: You are already updating the `unlocked_authorities` below. Why do you need to do this?

packages/daemon/src/db/index.ts
```diff
@@ -478,6 +486,98 @@ export const voidTransaction = async (
...
+      );
+
+      // If we're removing any of the authorities, we need to refresh the authority columns
+      // Similar to the address case, we need to recalculate authorities from all addresses in the wallet
+      if (tokenBalance.unlockedAuthorities.hasNegativeValue()) {
+        await mysql.query(
+          `UPDATE \`wallet_balance\`
+              SET \`unlocked_authorities\` = (
+                SELECT BIT_OR(\`unlocked_authorities\`)
+                  FROM \`address_balance\`
```
pedroferreira1, line 592: When do you update the `address_balance` table when a tx is voided?

packages/daemon/src/db/index.ts
```diff
@@ -478,6 +486,98 @@ export const voidTransaction = async (
...
+                    WHERE \`wallet_id\` = ?)
+                   AND \`token_id\` = ?)
+            WHERE \`wallet_id\` = ?
+              AND \`token_id\` = ?`,
+          [walletId, token, walletId, token],
+        );
+      }
+
+      // Handle locked authorities refresh if needed
+      if (tokenBalance.lockedAuthorities.hasNegativeValue && tokenBalance.lockedAuthorities.hasNegativeValue()) {
```
pedroferreira1: Why do you need both checks but not in the conditional above?

packages/daemon/__tests__/integration/balances.test.ts
andreabadesso: Done, thanks

packages/daemon/src/services/index.ts
andreabadesso: Fixed, thanks

packages/wallet-service/tests/txProposalUtxoUnlock.test.ts
andreabadesso: Yes, I forgot to remove the comment, thanks

packages/daemon/__tests__/services/services_with_db.test.ts
andreabadesso: Done, thanks

packages/daemon/__tests__/services/services_with_db.test.ts
andreabadesso: Fixed, thanks

packages/daemon/__tests__/services/services_with_db.test.ts
andreabadesso: Fixed, thanks

packages/daemon/src/db/index.ts
andreabadesso: Because we need to recalculate the `wallet_balance` state based on the entire wallet addresses authority balance, not only this single voided transaction

I improved the comment and also removed the lockedAuthorities recalculation as it is not needed

packages/daemon/src/db/index.ts
andreabadesso: In the [voidTransaction](https://github.com/HathorNetwork/hathor-wallet-service/blob/be594b82fa999d4f18c31d06384d14052e910925/packages/daemon/src/db/index.ts#L382) method

packages/daemon/src/db/index.ts
andreabadesso: This was clearly a bug. But I removed this entire locked authorities recalculation as it is not needed (locked authorities can't be spent)

packages/daemon/__tests__/integration/scenario_configs/voided_token_authority.balances.ts
```diff
@@ -0,0 +1,42 @@
...
+  },
+  walletBalances: {
+    // HTH balance: 6390 unlocked + 115200 locked (HRQe4CXj8AZXzSmuNztU8iQR74QTQMbnTs) + 0 unlocked + 100000000000 locked (HRXVDmLVdq8pgok1BCUKpiFWdAVAy4a5AJ)
+    'deafbeef:00': { unlockedBalance: 6390n, lockedBalance: 100000115200n },
+    'deafbeef:efb08b3e79e0ddaa6bc288183f66fe49a07ba0b7b2595861000478cc56447539': {
+      unlockedBalance: 1000n,
+      lockedBalance: 0n,
+      authorities: { unlocked: 2, locked: 0 }
+    },
+    'deadbeef:efb08b3e79e0ddaa6bc288183f66fe49a07ba0b7b2595861000478cc56447539': {
```
tuliomir, line 36: question: Is this a typo, or did the poor deaf thing really pass away?

packages/daemon/__tests__/integration/balances.test.ts
```diff
@@ -397,3 +410,116 @@ describe('transaction voiding chain scenario', () => {
...
+    jest.spyOn(Services, 'fetchMinRewardBlocks').mockImplementation(async () => 300);
+    await cleanDatabase(mysql);
+  });
+
+  afterAll(async () => {
+    // Clean up wallet data after this test to prevent affecting other tests
+    await cleanDatabase(mysql);
+  });
+
+  it.only('should do a full sync and the balances should match after voiding token authority', async () => {
```
tuliomir: issue: `only` clause will prevent other tests to run when added

packages/daemon/src/db/index.ts
```diff
@@ -426,37 +415,52 @@ export const voidTransaction = async (
...
+            tokenBalance.unlockedAuthorities.toUnsignedInteger(),
+            tokenBalance.lockedAuthorities.toUnsignedInteger(),
+            address,
+            token
+          ]
+        );
+      } else {
+        // Entry doesn't exist, this means the balance was never added in the first place
+        // This can happen in tests that don't simulate the complete transaction flow
+        console.log(`Warning: Trying to void transaction for address ${address} token ${token} but no balance entry exists`);
```
tuliomir: suggestion(non-blocking): Change `log` to `warn`
```suggestion
        console.warn(`Warning: Trying to void transaction for address ${address} token ${token} but no balance entry exists`);
```

packages/daemon/src/services/index.ts
```diff
@@ -485,6 +491,8 @@ export const handleVertexRemoved = async (context: Context, _event: Event) => {
     }
 
     logger.info(`[VertexRemoved] Voiding tx: ${hash}`);
+
+
```
r4mmer:
```suggestion
```

packages/daemon/src/services/index.ts
```diff
@@ -215,6 +220,7 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
 
     const isNano = isNanoContract(headers);
 
+
```
r4mmer:
```suggestion
```

packages/daemon/src/types/event.ts
andreabadesso: Done

packages/daemon/__tests__/integration/scenario_configs/voided_token_authority.balances.ts
andreabadesso: Not a typo, but I added the `cafecafe` wallet as well.

packages/daemon/__tests__/integration/balances.test.ts
andreabadesso: Removed it so it's ready for when the image is published.

packages/daemon/src/db/index.ts
andreabadesso: Done

packages/daemon/src/services/index.ts
andreabadesso: Done

packages/daemon/src/services/index.ts
andreabadesso: Done

r4mmer (dismissed):

tuliomir (dismissed):

tuliomir (dismissed):

tuliomir (dismissed):

r4mmer (dismissed):

andreabadesso:
e6458b8caea3e6ec326cf1de29decb63d00e1920 fixes an issue with the integration tests causing them to consistently fail when running all tests at once but to succeed when ran individually, here is a better explanation:

**Problem**: The `SyncMachine` context was being shared across all test instances, causing the LRU cache (`txCache`) to persist between tests. This led to the `unchanged` guard skipping genesis transactions in subsequent tests because they were already cached from previous tests.

**Root Cause**:

  1. Test 1 processes genesis transaction -> stores it in LRU cache
  2. Test 2 receives same genesis transaction -> `unchanged` guard finds it in cache -> skips processing
  3. Result: genesis address has 0 balance instead of expected balance

**Solution**: Changed from shared context to dynamically created LRU cache per machine instance:

  - Before: `txCache: new LRU(TX_CACHE_SIZE)` (shared)
  - After: `txCache: null` -> initialized fresh in `INITIALIZING` state
  
**More context on the `unchanged` guard**:

This guard is a performance optimization that prevents redundant database operations by checking if transaction metadata has actually changed. It compares the cached hash of a transaction's metadata against the current event's metadata hash. If they match, the event is ignored since we've already processed this exact state.

The guard is designed to skip repeated `VERTEX_METADATA_CHANGED` events for transactions we've already seen - for example, if only the accumulated weight changes (which wallet-service doesn't track), we can safely ignore the event. However, when the cache persisted across tests, it incorrectly skipped legitimate first-time processing of genesis transactions in subsequent test scenarios.

r4mmer (approved):

tuliomir (approved):

PR #283, opened 2025-08-27
chore: Container builds

tuliomir:
### Motivation

We need a way to run the Wallet Service inside a containerized environment containing a live and pristine blockchain privatenet. The first usage of this feature is being developed in https://github.com/HathorNetwork/hathor-wallet-lib/pull/909

### Acceptance Criteria

- Both the Daemon and the Wallet Service packages need independent Docker images
- There should be a way to run the migrations in the pristine MySQL instance before the Daemon container is started. For now, a third Docker image is created for the sole purpose of running the migrations.
- There should be solutions for the currently necessary connections to external services such as AWS, Firebase and other third parties. For now, they are being mocked.

### Notes

Although it's very important to automatically build all three images with every CI/CD deployment, this feature is not meant to be implemented now. The main objective of this PR is to have viable images for the integration tests on the Lib 909 mentioned above.

As soon as we have this infrastructure well tested, we'll begin work on the CD phase. For now, the Docker images will have to be published manually.

### Checklist


packages/wallet-service/package.json
```diff
@@ -64,7 +68,7 @@
     "serverless-api-gateway-throttling": "2.0.3",
     "serverless-better-credentials": "2.0.0",
     "serverless-iam-roles-per-function": "3.2.0",
-    "serverless-offline": "13.1.2",
+    "serverless-offline": "14.4.0",
```
tuliomir, line 71: The old dependency wasn't compatible with NodeJS 22 and had to be upgraded.

scripts/fetch-fullnode-ids.js
```diff
@@ -0,0 +1,131 @@
...
+
+        // Create .env file content
+        const envContent = `STREAM_ID=${streamId}\nFULLNODE_PEER_ID=${fullnodePeerId}\n`;
+
+        // Write to output file
+        fs.writeFileSync(outputFileName, envContent);
+        console.error(`Written identifiers to ${outputFileName}`);
+
+        // Output to stdout for shell script consumption
+        console.log(envContent);
```
luislhl: Who will consume this exactly?

scripts/fetch-fullnode-ids.js
```diff
@@ -0,0 +1,131 @@
...
+        if (!streamId || !fullnodePeerId) {
+          throw new Error(`Missing required fields in response. stream_id: ${streamId}, peer_id: ${fullnodePeerId}`);
+        }
+
+        // Create .env file content
+        const envContent = `STREAM_ID=${streamId}\nFULLNODE_PEER_ID=${fullnodePeerId}\n`;
+
+        // Write to output file
+        fs.writeFileSync(outputFileName, envContent);
+        console.error(`Written identifiers to ${outputFileName}`);
```
luislhl: Why are you using console.error on logs that are just info ones?

packages/daemon/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+# warranty. In no event will the authors be held liable for any damages
+# arising from the use of this software.
+# This software cannot be redistributed unless explicitly agreed in writing with the authors.
+
+# =========================================================================
+# This Dockerfile is used to build and run the Wallet Service Daemon container.
+# It requires:
+# - A MySQL instance, properly migrated ( see /db/Dockerfile )
+# - A Fullnode instance
+# - A Redis instance
```
andreabadesso, line 12: I don't think we need Redis in the daemon

packages/daemon/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+
+# Copy only the necessary files from the build phase
+COPY --from=builder /app .
+
+WORKDIR /app/packages/daemon/
+
+# The script should already be available from the builder stage copy
+RUN cp /app/scripts/fetch-fullnode-ids.js ./fetch-fullnode-ids.js
+RUN cp /app/scripts/merge-complementary-envs.sh ./merge-complementary-envs.sh
+RUN chmod +x ./merge-complementary-envs.sh
```
andreabadesso, line 68: What are these for?

packages/wallet-service/.aws/credentials
```diff
@@ -0,0 +1,8 @@
+# AWS Credentials for Local Development
```
andreabadesso, line 1: What is this file for? I don't like the idea of adding it to `.aws` in the project, can't you just add this inside the wallet-service's Dockerfile and only when an "DEV" arg is passed?

packages/wallet-service/src/api/wallet.ts
```diff
@@ -6,9 +6,10 @@
  */
 
 import { APIGatewayProxyHandler, Handler, SNSEvent } from 'aws-lambda';
-import { LambdaClient, InvokeCommand, InvokeCommandOutput } from '@aws-sdk/client-lambda';
+import { InvokeCommand, InvokeCommandOutput } from '@aws-sdk/client-lambda';
 import 'source-map-support/register';
 
+import { createLambdaClient } from '@src/utils/aws-offline-mock';
```
andreabadesso: Why are you adding mocks to the wallet-service code?

packages/wallet-service/src/utils/aws-offline-mock.ts
```diff
@@ -0,0 +1,135 @@
...
+import { LambdaClient, InvokeCommand } from '@aws-sdk/client-lambda';
+import { ApiGatewayManagementApiClient, PostToConnectionCommand } from '@aws-sdk/client-apigatewaymanagementapi';
+import createDefaultLogger from '@src/logger'
+
+/**
+ * Flag to determine if AWS services should be mocked
+ * It is set by default via the MOCK_AWS environment variable, but can be adjusted programmatically,
+ * especially for testing environments.
+ */
+export let shouldMockAWS = process.env.MOCK_AWS === 'true';
```
andreabadesso: Use `src/config.ts` instead of just getting from `process.env` here

packages/wallet-service/src/api/wallet.ts
andreabadesso: Oh ok, the `createLambdaClient` is deciding wether to mock it or not

Maybe improve the file name so it's not misleading

packages/wallet-service/src/utils/aws-offline-mock.ts
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * Set whether to mock AWS services
+ * @param value
+ */
+export function setShouldMockAWS(value: boolean) {
+  shouldMockAWS = value;
+}
```
andreabadesso: Why would we need this setter?

packages/wallet-service/Dockerfile
```diff
@@ -0,0 +1,68 @@
...
+
+# Copy root package files
+COPY . .
+
+# Enable corepack for yarn
+RUN corepack enable
+RUN yarn install
+
+WORKDIR /app/packages/wallet-service
+# RUN yarn install
```
andreabadesso: Remove

scripts/fetch-fullnode-ids.js
```diff
@@ -0,0 +1,131 @@
...
+
+const WebSocket = require('ws');
+const fs = require('fs');
+
+/**
+ * Please note that this URL assumes a Docker setup where the fullnode service
+ * is accessible via the hostname 'fullnode' on port 8080. Adjust as necessary.
+ * @type {string}
+ */
+const fullnodeBaseUrl = 'fullnode:8080';
```
andreabadesso: Use an env here

scripts/fetch-fullnode-ids.js
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * Output .env file name
+ * @type {string}
+ */
+const outputFileName = '.identifiers.env';
```
andreabadesso: Same, and default to `.identifiers.env`

scripts/fetch-fullnode-ids.js
```diff
@@ -0,0 +1,131 @@
+// Copyright 2025 Hathor Labs
```
andreabadesso, line 1: You can replace this entire file with a simple bash script using `wscat` and `jq`:

```
timeout 5s npx wscat -c ws://fullnode:8080/v1a/event_ws -x '{"type": "START_STREAM", "window_size": 1}' | head -1 | jq -r '"FULLNODE_PEER_ID=" + .peer_id, "STREAM_ID=" + .stream_id' > .identifiers.env
```

scripts/merge-complementary-envs.sh
```diff
@@ -0,0 +1,66 @@
...
+
+    # Skip if key is empty after trimming
+    if [ -z "$key" ]; then
+        continue
+    fi
+
+    # Export the variable
+    export "$key=$value"
+    echo "Exported: $key=$value"
+done < .identifiers.env
```
andreabadesso:
```suggestion
source .identifiers.env
```

andreabadesso (changes requested):

scripts/build-service.sh
```diff
@@ -0,0 +1,17 @@
+set -e
+set -o pipefail
+
+
+FINAL_TAG="hathornetwork/hathor-wallet-service-service:latest";
```
luislhl: This is just a thought for the future: We need to be careful with what names we're going to use for the images published to Docker Hub (if we're going to indeed publish there), to make it clear that they are meant for local testing only.

Even if we just publish to our internal registry, this would be a good thing to do as well.

For now, since we're not publishing them anywhere yet, the name is good enough.

packages/wallet-service/src/utils/aws-offline-mock.ts
luislhl: You could add an `ENV MOCK_AWS=true` instruction in the Dockerfiles, to make sure they will always use true as default.

I guess nobody using the images should disable this, right?

packages/wallet-service/Dockerfile.dev
luislhl, line 1: I'm thinking whether we should name the Dockerfiles that are specifically tailored for local usage as `Dockerfile.dev` or `Dockerfile.local`, to differentiate from the production ones. 

What do you think?

packages/daemon/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+
+# The daemon could need complementary environment variables dynamically.
+# The entrypoint script manages this before actually running the daemon.
+ENTRYPOINT ["./merge-complementary-envs.sh"]
```
luislhl, line 72: There is a way to avoid adding the scripts to the production image.

We could do something like this:

```
FROM node:22-alpine AS dev
WORKDIR /app

# Copy only the necessary files from the build phase
COPY --from=builder /app .

WORKDIR /app/packages/daemon/

# The script should already be available from the builder stage copy
RUN cp /app/scripts/fetch-fullnode-ids.js ./fetch-fullnode-ids.js
RUN cp /app/scripts/merge-complementary-envs.sh ./merge-complementary-envs.sh
RUN chmod +x ./merge-complementary-envs.sh

# The daemon could need complementary environment variables dynamically.
# The entrypoint script manages this before actually running the daemon.
ENTRYPOINT ["./merge-complementary-envs.sh"]

# ---- Prod Runner ----
FROM node:22-alpine AS prod
WORKDIR /app

# Copy only the necessary files from the build phase
COPY --from=builder /app .

WORKDIR /app/packages/daemon/

# Production runs directly, no helper scripts
CMD ["node", "dist/index.js"]
```

Then, when building the images, build them differently:

```
docker build --target dev -t my-daemon-dev .

docker build --target prod -t my-daemon-prod .
```

This avoids adding the scripts that are meant only for dev usage in the production image, and you'll be able to simplify the scripts I think, since they won't need to treat the production case (like having an env var to disable/enable).

What do you think?

scripts/fetch-fullnode-ids.js
tuliomir: This is obsolete code. In previous iterations of this script, it would be used to output the env variables to `stdout` instead of a file. I removed the old code and refactored/adapted the remaining on https://github.com/HathorNetwork/hathor-wallet-service/pull/283/commits/0d198b578902a1222752864974b80d217da26f5e

scripts/fetch-fullnode-ids.js
tuliomir: As explained above, this script was supposed to output to `stdout` the results of the env variables. So to avoid polluting the output, every informative log was sent to stderr. This complicated solution was removed on https://github.com/HathorNetwork/hathor-wallet-service/pull/283/commits/0d198b578902a1222752864974b80d217da26f5e

scripts/fetch-fullnode-ids.js
tuliomir: Good point. Implemented on https://github.com/HathorNetwork/hathor-wallet-service/pull/283/commits/0d198b578902a1222752864974b80d217da26f5e

scripts/fetch-fullnode-ids.js
tuliomir: Implemented on https://github.com/HathorNetwork/hathor-wallet-service/pull/283/commits/0d198b578902a1222752864974b80d217da26f5e

packages/wallet-service/Dockerfile
tuliomir: Removed on https://github.com/HathorNetwork/hathor-wallet-service/pull/283/commits/c8f5556845b530cf8eb0ad4bbc0125ea0834fea9 , thanks.

packages/wallet-service/src/utils/aws-offline-mock.ts
tuliomir: Good point. In my original intention, this `wallet-service` image would be used for development, but not necessarily for dockerized private networks. For the cases when the containerized services would connect to public networks ( `testnet-dev`, `testnet-staging`, for example ) , this mock would not be desirable.

@andreabadesso , since you interact more with this app in many dev conditions, do you have an opinion on this? Maybe the major (or sole) purpose of this image will be for containerized private networks. If so, it's nice to have this `MOCK_AWS` flag set as true by default here.

tuliomir: For now, I improved the documented comments on the Dockerfile on https://github.com/HathorNetwork/hathor-wallet-service/pull/283/commits/39768b844ed339b8da787859ea0ed4f84228209c

packages/wallet-service/src/api/wallet.ts
tuliomir: Thanks for the idea. Renamed to `utils/aws.utils.ts` on https://github.com/HathorNetwork/hathor-wallet-service/pull/283/commits/b581b4a73d4f1e9c9ef18bd8b7653204e4ebd3eb

packages/wallet-service/src/utils/aws-offline-mock.ts
tuliomir: Thanks, refactored to use the config on https://github.com/HathorNetwork/hathor-wallet-service/pull/283/commits/b581b4a73d4f1e9c9ef18bd8b7653204e4ebd3eb

packages/wallet-service/src/utils/aws-offline-mock.ts
tuliomir: This setter was needed because the unit tests needed to change this value dynamically. This was solved by https://github.com/HathorNetwork/hathor-wallet-service/pull/283/commits/b581b4a73d4f1e9c9ef18bd8b7653204e4ebd3eb when I refactored this file to use the config instead of env directly.

packages/daemon/Dockerfile
tuliomir: Removed on https://github.com/HathorNetwork/hathor-wallet-service/pull/283/commits/089014b1932eb845db6589ecd071dd112f2e0e67

packages/daemon/Dockerfile
tuliomir: These scripts add the mandatory fullnode identifiers to the container when it's used inside a containerized privated network. See the docs on the entrypoint script:

https://github.com/HathorNetwork/hathor-wallet-service/blob/089014b1932eb845db6589ecd071dd112f2e0e67/scripts/merge-complementary-envs.sh#L10-L16

scripts/fetch-fullnode-ids.js
tuliomir: Unfortunately I wasn't able to implement this solution in this environment. Even though I tried many variations of this suggestion of yours, `wscat` is never able to reach the fullnode.

For now, I'd prefer keeping this sub-optimal solution and improve it in a next PR. What do you think?

scripts/merge-complementary-envs.sh
tuliomir: I also wasn't able to make `source` work in this enviroment, despite trying many variations of this suggestion.

I'd also prefer keeping this sub-optimal approach and improve it later. What do you think?

scripts/build-service.sh
tuliomir: I'll ask for @andreabadesso opinion here as well.

I agree with you, Luis: this image will be used only for development purposes, but I tried keeping the name as close as possible to the current one on the `sync-daemon`. I confess the `...service-service` name didn't sound very nice to me, but I had no better suggestion to offer..

Maybe we could change only the tag of this image? `"hathornetwork/hathor-wallet-service-service:dev"`?

packages/daemon/Dockerfile
tuliomir: That's a nice suggestion, as it reverts the production code to the simpler form it had before. Refactored on https://github.com/HathorNetwork/hathor-wallet-service/pull/283/commits/5a5b2bda1822162fb17755f52f9458c913af4f2f , thanks!

packages/wallet-service/Dockerfile.dev
tuliomir: I tried renaming this file to `Dockerfile.dev`, but the IDE removed the `Dockerfile` syntax when I did so. This suggests to me that this change would break some convention.

I'll keep looking at it, but would you have some other suggestion? For now I tried adding as much information on the file comments as possible to avoid confusion.

tuliomir (commented):
GitHub didn't allow me to reply to the comments directly, and forced me to create my own review to do so.

packages/wallet-service/src/utils/aws-offline-mock.ts
luislhl: Ok, I see. It makes sense.

From my part you could resolve this conversation when you want, I left open just because Abadesso was questioned as well.

packages/wallet-service/.aws/credentials
tuliomir: Good catch, these files should have been refactored already. Fixed this on https://github.com/HathorNetwork/hathor-wallet-service/pull/283/commits/2bbd9f257388495ae8b4f9cd2ddfac6dcbd0d07e

scripts/build-service.sh
luislhl: I think changing the tag is enough

packages/wallet-service/Dockerfile.dev
luislhl: I think it's a pretty common practice: https://github.com/search?q=Dockerfile.dev&type=commits&p=1

Also, I tested on VSCode and it didn't remove the syntax when renamed to Dockerfile.dev.

Maybe this is more a question about your IDE. Which one is it?

But I don't consider this a blocker for the PR anyway.

scripts/merge-complementary-envs.sh
luislhl: Maybe you would have to add `export` to the .identifiers.env to make it work with source.

I made a simple test here to try to simulate your situation, with 3 files:

`.test.env`:

```
.test.env 
export STREAM_ID=123123
export FULLNODE_PEER_ID=345345
```

`test.sh`:

```
source .test.env

node test.js
```

`test.js`:

```
console.log(process.env.STREAM_ID);

console.log(process.env.FULLNODE_PEER_ID);
```

When running `bash test.sh`, I get:

```
❯ bash test.sh 
123123
345345
```

If I remove the `export` statements from the .env, which I believe is what happens in your case, then I get:

```
❯ bash test.sh 
undefined
undefined
```

Does this make any sense?

luislhl (dismissed):

luislhl (dismissed):

scripts/merge-complementary-envs.sh
andreabadesso: @tuliomir did you try @luislhl's suggestion?

scripts/build-service.sh
andreabadesso: Maybe use `hathornetwork/hathor-wallet-service-lambdas:dev`

packages/wallet-service/Dockerfile.dev
andreabadesso: I don't like the idea of changing the production `Dockerfile` as well

andreabadesso (changes requested):

packages/wallet-service/Dockerfile.dev
tuliomir: Thanks for the context, Luis! Changed the `Dockerfile` to `Dockerfile.dev` on https://github.com/HathorNetwork/hathor-wallet-service/pull/283/commits/3451eac36dbdeb6274244c49eccb1cebd1b9105d

scripts/build-service.sh
tuliomir: Thanks for the inputs! Changed the name and the tag of the development images on https://github.com/HathorNetwork/hathor-wallet-service/pull/283/commits/3451eac36dbdeb6274244c49eccb1cebd1b9105d

scripts/merge-complementary-envs.sh
tuliomir: Yes I did, and it worked well! My mistake was considering a plain `.env` file to work with `source`, when in reality it had to be a `.sh` script.

I refactored it on https://github.com/HathorNetwork/hathor-wallet-service/pull/283/commits/f6546e2c32d701fb36025be8755f5da7b462d7a5 and it's working properly now.

andreabadesso (approved):

luislhl (approved):

PR #286, opened 2025-09-04
chore: add logs to serverless mysql manager

r4mmer:
### Motivation

Add logs to investigate connection issues


r4mmer:
Since the investigation of this issue is over this branch can be deleted

PR #287, opened 2025-09-08
chore: warmup middleware for authorization lambdas

r4mmer:
### Motivation

Having both the auth token api and the authorizer lambda cold start at the same time increases the wallet start by more than 2 seconds.
Having them always be warmed would improve the UX.

### Acceptance Criteria

- Add the warmup middleware for both the auth token api and authorizer lambda

### Extra chores

- Added missing error treatment in auth token api
- Fixed fullnode event parser
- Prevent the get new addresses to fetch the same walet twice from the database.

### Checklist

### DIFF
--- a/packages/daemon/src/types/event.ts
+++ b/packages/daemon/src/types/event.ts
@@ -104,10 +104,10 @@ export const EventTxOutputSchema = z.object({
       address: z.string(),
       timelock: z.number().nullable(),
     }).passthrough().nullable(),
-    z.object({}).strict(),
     z.object({
-      token_data: z.number().nullish(),
-    }).strict()
+      token_data: z.number(),
+    }),
+    z.object({}).strict(),
   ]),
 });
 export type EventTxOutput = z.infer<typeof EventTxOutputSchema>;
--- a/packages/daemon/src/types/event.ts
+++ b/packages/daemon/src/types/event.ts
@@ -103,9 +103,11 @@ export const EventTxOutputSchema = z.object({
       type: z.string(),
       address: z.string(),
       timelock: z.number().nullable(),
-      token_data: z.number().nullish(),
     }).passthrough().nullable(),
-    z.object({}).strict()
+    z.object({}).strict(),
+    z.object({
+      token_data: z.number().nullish(),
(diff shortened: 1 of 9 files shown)


packages/daemon/src/types/event.ts
```diff
@@ -103,8 +103,11 @@ export const EventTxOutputSchema = z.object({
       type: z.string(),
       address: z.string(),
       timelock: z.number().nullable(),
-    }).nullable(),
-    z.object({}).strict()
+    }).passthrough().nullable(),
+    z.object({}).strict(),
+    z.object({
+      token_data: z.number().nullish(),
+    }).strict()
```
r4mmer: There was an event from the simulator with `{ token_data: 0 }` which would break the parser, so I added this option.

packages/wallet-service/src/api/auth.ts
```diff
@@ -83,6 +84,17 @@ export const tokenHandler: APIGatewayProxyHandler = middy(async (event) => {
...
+  if (!wallet) {
+    await closeDbConnection(mysql);
+    return {
+      statusCode: 400,
+      body: JSON.stringify({
+        success: false,
+        error: ApiError.WALLET_NOT_FOUND,
+      }),
+    };
+  }
```
r4mmer, line 96: Just including an error treatment in case we don't find the wallet.

pedroferreira1 (dismissed):

andreabadesso (dismissed):

packages/daemon/src/types/event.ts
andreabadesso: I solved it differently in https://github.com/HathorNetwork/hathor-wallet-service/pull/281/files#r2334854251, what do you think?

r4mmer: Changed to be the same as yours since it acomplishes the same thing and makes conflict resolution easier

pedroferreira1 (dismissed):

andreabadesso (dismissed):

andreabadesso (approved):

pedroferreira1 (approved):

PR #291, opened 2025-09-17
fix: voided token addr balance

andreabadesso:
### Motivation

When an address is involved in a token creation transaction that is later on voided and this transaction is the only one this address has received so far, the `initWalletBalance` method (that is called when initializing a new wallet) throws because of a mismatch in the rows from `address_balance` and `address_tx_history`

This happens because when a transaction is voided, we remove it from the `address_tx_history` but the address_balance is set to `0` and never removed again.

### Acceptance Criteria

- We should include a test that will fail if the number of rows in both tables for the voided token is different
- We should delete from the `address_balance` table in case the only row for that token is zero 
- We should delete from the `wallet_balance` in case the only row for that token is zero

### Checklist


packages/daemon/src/db/index.ts
```diff
@@ -396,28 +398,14 @@ export const voidAddressTransaction = async (
...
-        // if it's < 0, there must be an entry already, so it will execute "ON DUPLICATE KEY UPDATE" instead of setting it to 0
-        unlocked_balance: (tokenBalance.unlockedAmount < 0 ? 0 : tokenBalance.unlockedAmount),
-        // this is never less than 0, as locked balance only changes when a tx is unlocked
-        locked_balance: tokenBalance.lockedAmount,
-        unlocked_authorities: tokenBalance.unlockedAuthorities.toUnsignedInteger(),
-        locked_authorities: tokenBalance.lockedAuthorities.toUnsignedInteger(),
-        timelock_expires: tokenBalance.lockExpires,
-        transactions: 1,
-      };
-
```
andreabadesso, line 417: This should have been removed in the last PR

packages/daemon/src/services/index.ts
```diff
@@ -521,6 +523,7 @@ export const voidTx = async (
   outputs: EventTxOutput[],
   tokens: string[],
   headers: EventTxHeader[],
+  version?: number,
```
pedroferreira1: Why is it optional?

packages/daemon/src/db/index.ts
```diff
@@ -480,6 +468,31 @@ export const voidAddressTransaction = async (
       // for locked authorities, it doesn't make sense to perform the same operation. The authority needs to be
       // unlocked before it can be spent. In case we're just adding new locked authorities, this will be taken
       // care by the first sql query.
+
+      // If this is a token creation transaction, we must remove the address_balance
+      // row for that token as it has ceased to exist
+      if (isCreateTokenTx) {
```
pedroferreira1: As we discussed, we need to handle this case even for non token creation tx. The tx might be the only one that added balance to the address and must also be deleted.

packages/daemon/src/db/index.ts
```diff
@@ -600,6 +617,22 @@ export const voidWalletTransaction = async (
           [walletId, token, walletId, token],
         );
       }
+
+      // If this is a token creation transaction, clean up zeroed entries
+      if (isCreateTokenTx) {
```
pedroferreira1: Same as the other

packages/daemon/__tests__/integration/scenario_configs/single_voided_regular_transaction.balances.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+export default {
+  addressBalances: {
+    'HFtz2f59Lms4p3Jfgtsr73s97MbJHsRENh:00': { unlockedBalance: 0n, lockedBalance: 0n },
```
andreabadesso: I'll remove this line

packages/daemon/src/db/index.ts
andreabadesso: Done

packages/daemon/src/db/index.ts
andreabadesso: Done

packages/daemon/src/services/index.ts
andreabadesso: Fixed

pedroferreira1 (approved):

r4mmer (approved):

PR #298, opened 2025-10-06
feat(daemon): implemented delay on ACK + check for missed events

andreabadesso:
```mermaid
stateDiagram-v2
      [*] --> INITIALIZING
      INITIALIZING --> CONNECTING: Fetch initial state
      CONNECTING --> CONNECTED: WebSocket connected
      CONNECTING --> RECONNECTING: Connection failed

      state CONNECTED {
          [*] --> idle

          idle --> processingEvent: FULLNODE_EVENT
          processingEvent --> idle: Send ACK<br/>Start timeout ⏱️

          idle --> checkingForMissedEvents: ⏱️ ACK_TIMEOUT

          checkingForMissedEvents --> idle: No missed events ✅
      }

      checkingForMissedEvents --> RECONNECTING: Has missed events ⚠️

      CONNECTED --> RECONNECTING: WebSocket disconnected
      CONNECTED --> ERROR: Invalid stream/peer/network

      RECONNECTING --> CONNECTING: After backoff delay

      ERROR --> [*]

      note right of CONNECTED
          Event Processing Flow:
          1. Cancel ACK timeout
          2. Process event
          3. Send ACK
          4. Start ACK timeout

          Timeout Check:
          • HTTP GET /event
          • If events found → Reconnect
          • If no events → Back to idle
      end note

```

## Motivation

If we lost events from the fullnode due to network packet loss, the daemon would hang indefinitely waiting for an event that would never arrive. The WebSocket connection would remain open, but no new events would be received, causing the sync process to stall without any error or recovery mechanism.

## Acceptance Criteria
- Implement an ACK timeout mechanism that triggers when no events are received for a configured period
- When timeout expires, check the fullnode HTTP API (/event) to verify if events were missed
- If missed events are detected, reconnect to the WebSocket to resume event streaming
- If no missed events are found, return to idle state and continue waiting
- Ensure the timeout is cancelled and restarted appropriately on each event received
- Handle critical failures in the event verification process by transitioning to ERROR state
- Add appropriate logging for debugging (warn when missed events detected, debug when check passes)

### Checklist

### DIFF
--- a/packages/daemon/src/machines/SyncMachine.ts
+++ b/packages/daemon/src/machines/SyncMachine.ts
@@ -309,8 +309,8 @@ export const SyncMachine = Machine<Context, any, Event>({
               target: CONNECTED_STATES.idle,
             }],
             onError: {
-              // On error, return to idle and continue normal operation
-              target: CONNECTED_STATES.idle,
+              // Critical failure - we cannot verify event integrity
+              target: `#${SYNC_MACHINE_STATES.ERROR}`,
             },
           },
         },
--- a/packages/daemon/src/machines/SyncMachine.ts
+++ b/packages/daemon/src/machines/SyncMachine.ts
@@ -26,6 +26,7 @@ import {
   fetchInitialState,
   handleUnvoidedTx,
   handleReorgStarted,
+  checkForMissedEvents,
 } from '../services';
 import {
   metadataIgnore,
@@ -43,6 +44,7 @@ import {
   unchanged,
   vertexRemoved,
   reorgStarted,
+  hasNewEvents,
 } from '../guards';
 import {
   storeInitialState,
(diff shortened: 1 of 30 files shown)


pedroferreira1 (approved):

packages/daemon/src/machines/SyncMachine.ts
```diff
@@ -137,6 +140,11 @@ export const SyncMachine = Machine<Context, any, Event>({
       states: {
         [CONNECTED_STATES.idle]: {
           id: CONNECTED_STATES.idle,
+          after: {
+            ACK_TIMEOUT: {
```
r4mmer, line 144:
```suggestion
            [ACK_TIMEOUT]: {
```

Shouldn't this be the act timeout value?

andreabadesso: It's a weird syntax, but no, this is a "referenced delay":

https://stately.ai/docs/delayed-transitions#referenced-delays

r4mmer: Nice, naming it the same as the variable threw me off a little.

r4mmer (approved):

PR #299, opened 2025-10-08
feat(wallet-service): added read-only authorizer and api

andreabadesso:
Implements https://github.com/HathorNetwork/internal-rfcs/pull/39

### Acceptance Criteria

- We must be able to generate JWTs for read-only wallets (using xpubkeys)
- Wallets must be in the `READY` state in order to authenticate, we must error otherwise.
- APIs that mutate database state must be marked as write and we should reject requests made using a read-only JWT.

### Checklist

### DIFF
--- a/packages/wallet-service/src/api/auth.ts
+++ b/packages/wallet-service/src/api/auth.ts
@@ -342,12 +342,6 @@ export const bearerAuthorizer: APIGatewayTokenAuthorizerHandler = middy(async (e
 
   const walletId = data.wid;
   const mode = data.mode || 'full'; // Default to full for legacy tokens
-  const expirationTs = data.exp;
-
-  // Check token expiration
-  if (Math.floor(Date.now() / 1000) > expirationTs) {
-    throw new Error('Unauthorized');
-  }
 
   // For full-access tokens, verify wallet signature (existing logic)
   if (mode === 'full') {
--- a/packages/wallet-service/src/api/auth.ts
+++ b/packages/wallet-service/src/api/auth.ts
@@ -160,6 +160,7 @@ export const tokenHandler: APIGatewayProxyHandler = middy(async (event) => {
       ts: timestamp,
       addr: address.toString(),
       wid: walletId,
+      mode: 'full',
     },
     config.authSecret,
     {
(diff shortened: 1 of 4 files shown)


r4mmer (dismissed):

packages/wallet-service/src/api/auth.ts
```diff
@@ -233,20 +340,29 @@ export const bearerAuthorizer: APIGatewayTokenAuthorizerHandler = middy(async (e
...
+  const mode = data.mode || 'full'; // Default to full for legacy tokens
   const expirationTs = data.exp;
-  const verified = verifySignature(signature, timestamp, address, walletId);
 
-  if (verified && Math.floor(Date.now() / 1000) <= expirationTs) {
-    return _generatePolicy(walletId, 'Allow', event.methodArn, logger);
+  // Check token expiration
+  if (Math.floor(Date.now() / 1000) > expirationTs) {
+    throw new Error('Unauthorized');
+  }
```
pedroferreira1: We already have above this

```
    } else if (e.name === 'TokenExpiredError') {
      throw new Error('Unauthorized');
```

Isn't it checking expiration?

And in the old code we were returning a Deny policy for this if:

```
return _generatePolicy(walletId, 'Deny', event.methodArn, logger);
```

Is it ok to change this behavior?

pedroferreira1 (dismissed):

packages/wallet-service/src/api/auth.ts
andreabadesso: Removed the duplicate check, jwt.verify already checks for expiration.

> Is it ok to change this behavior?

Yes. Returning a policy with a "Deny" is equivalent to returning a 403, which is not what we want here, we want to return an "Unauthorized", i.e. 401 because the jwt is not valid

Different from the other case where the signature doesn't match:

```javascript
    if (!verified) {
      return _generatePolicy(walletId, 'Deny', event.methodArn, logger, mode);
    }
```

Where the jwt is valid, but the signature doesn't match so he's forbidden

pedroferreira1 (approved):

r4mmer (approved):

PR #301, opened 2025-10-09
feat: stop lib background tasks

r4mmer:
### Motivation

The wallet-lib start a background task that prevents jest from finishing properly.
The lib (after v2.8.X) exports a method to stop this task so we should use it.

### Acceptance Criteria

- Bump the lib to 2.8.3
- Call `stopGLLBackgroundTask` before all tests

### Checklist

### DIFF
--- a/packages/common/jestSetup.ts
+++ b/packages/common/jestSetup.ts
@@ -1,0 +1,3 @@
+import { stopGLLBackgroundTask } from '@hathor/wallet-lib';
+stopGLLBackgroundTask();
+
--- a/packages/common/jestSetup.ts
+++ b/packages/common/jestSetup.ts
@@ -1,3 +1,10 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
 import { stopGLLBackgroundTask } from '@hathor/wallet-lib';
 stopGLLBackgroundTask();
 
--- a/packages/daemon/jestSetup.ts
+++ b/packages/daemon/jestSetup.ts
@@ -1,3 +1,6 @@
 import { stopGLLBackgroundTask } from '@hathor/wallet-lib';
+
+Object.defineProperty(global, '_bitcore', { get() { return undefined; }, set() {} });
+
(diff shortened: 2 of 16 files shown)


packages/wallet-service/tests/pushSendNotificationToDevice.test.ts
```diff
@@ -253,6 +254,7 @@ describe('alert', () => {
     // allow android and desktop, while test for ios provider
     process.env.PUSH_ALLOWED_PROVIDERS = 'android,desktop';
     await import('@src/api/pushSendNotificationToDevice');
+    await stopWalletLibOpenHandles();
```
pedroferreira1, line 257: Why do you need this here and not in the `beforeAll`?

pedroferreira1 (dismissed):

packages/common/jestSetup.ts
```diff
@@ -0,0 +1,3 @@
+import { stopGLLBackgroundTask } from '@hathor/wallet-lib';
```
andreabadesso, line 8: Missing hathor header

packages/daemon/jestSetup.ts
```diff
@@ -0,0 +1,6 @@
+import { stopGLLBackgroundTask } from '@hathor/wallet-lib';
```
andreabadesso, line 8: Missing hathor header

packages/wallet-service/tests/pushSendNotificationToDevice.test.ts
r4mmer: In this file the `jest.resetModules` is called in the `beforeEach` which resets all imports for the test.
The dynamic import has somewhere down the line of imports imported the wallet-lib which resets the background task.

We can move this call to the before each but the only dynamic import here is in this test.

packages/daemon/jestSetup.ts
r4mmer: Done

packages/common/jestSetup.ts
r4mmer: Done

pedroferreira1 (approved):

andreabadesso (approved):

PR #307, opened 2025-11-03
fix: consider Serverless prefix when invoking functions

luislhl:
### Motivation

Some parts of the code have hardcoded function prefixes, and we recently started allowing customizable prefixes

### Acceptance Criteria

- The wallet-service and daemon packages should read the env var used to configure the prefix, with a default of `hathor-wallet-service`

### TODO

- [ ] We should redeploy testnet-india after this PR is merged (lambdas and daemon, also check https://github.com/HathorNetwork/ops-tools/pull/1102)

### Checklist

### DIFF
--- a/.codebuild/buildspec.yml
+++ b/.codebuild/buildspec.yml
@@ -59,6 +59,7 @@ env:
     testnetindia_PUSH_NOTIFICATION_ENABLED: "true"
     testnetindia_PUSH_ALLOWED_PROVIDERS: "android,ios"
     testnetindia_APPLICATION_NAME: "wallet-service-testnet-india"
+    testnetindia_SERVERLESS_DEPLOY_PREFIX: "wallet-service"
     mainnet_staging_DEFAULT_SERVER: "https://wallet-service.private-nodes.hathor.network/v1a/"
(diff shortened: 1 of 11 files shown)


andreabadesso (dismissed):

andreabadesso (approved):

.codebuild/buildspec.yml
```diff
@@ -59,6 +59,7 @@ env:
     testnetindia_PUSH_NOTIFICATION_ENABLED: "true"
     testnetindia_PUSH_ALLOWED_PROVIDERS: "android,ios"
     testnetindia_APPLICATION_NAME: "wallet-service-testnet-india"
+    testnetindia_SERVERLESS_DEPLOY_PREFIX: "wallet-service"
```
tuliomir, line 62: question(non-blocking): Just double-checking that we won't be using the default `'hathor-wallet-service'` prefix for `testnet-india`, right?

tuliomir (approved):

.codebuild/buildspec.yml
luislhl: Right, it was the first one where I tested a custom prefix

PR #311, opened 2025-11-17
fix: exp retry checkformissingevents

andreabadesso:
### Acceptance Criteria
- The `checkForMissedEvents` service implements a retry mechanism with exponential backoff
- Network errors and 5xx server errors trigger automatic retries (up to 10 attempts)
- Exponential backoff starts at 1 second and caps at 10 seconds maximum delay
- After exhausting all 10 retries (~75 seconds total), the service transitions to ERROR state with appropriate error logging
- Production errors like "Network error - Connection refused" will now trigger automatic retries instead of immediate failure

### Checklist

### DIFF
--- a/packages/daemon/src/utils/retry.ts
+++ b/packages/daemon/src/utils/retry.ts
@@ -1,0 +1,106 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import logger from '../logger';
+
+export interface RetryOptions {
+  maxRetries?: number;
+  initialDelayMs?: number;
+  maxDelayMs?: number;
+  backoffMultiplier?: number;
+  retryableErrors?: (error: any) => boolean;
+}
+
+const DEFAULT_OPTIONS: Required<RetryOptions> = {
+  maxRetries: 3,
+  initialDelayMs: 1000,
+  maxDelayMs: 10000,
+  backoffMultiplier: 2,
+  retryableErrors: (error: any) => {
+    // Retry on network errors and 5xx server errors
+    if (error.response) {
+      // HTTP error response received
+      return error.response.status >= 500 && error.response.status < 600;
+    }
+    // Network error (no response received)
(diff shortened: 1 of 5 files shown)


pedroferreira1 (approved):

packages/daemon/src/utils/retry.ts
```diff
@@ -0,0 +1,106 @@
...
+  maxDelayMs: 10000,
+  backoffMultiplier: 2,
+  retryableErrors: (error: any) => {
+    // Retry on network errors and 5xx server errors
+    if (error.response) {
+      // HTTP error response received
+      return error.response.status >= 500 && error.response.status < 600;
+    }
+    // Network error (no response received)
+    return true;
```
luislhl, line 30: I think there are errors that are useless to retry. I asked Sonnet for some examples (which I may not agree to all, but most of them):

```
  'ENOTFOUND',           // DNS resolution failed - invalid hostname
  'ERR_INVALID_URL',     // Malformed URL
  'ECONNABORTED',        // Request was explicitly cancelled
  'ERR_CANCELED',        // User/code cancelled the request
  'UNABLE_TO_VERIFY_LEAF_SIGNATURE', // SSL certificate validation failed
  'CERT_HAS_EXPIRED',    // Certificate expired
  'DEPTH_ZERO_SELF_SIGNED_CERT', // Self-signed certificate rejected
  'ERR_FR_MAX_BODY_LENGTH_EXCEEDED', // Payload too large
  'ERR_BAD_OPTION_VALUE', // Invalid configuration
  'ERR_BAD_OPTION',      // Invalid configuration
```

While these are the retryable ones it suggested:

```
// ECONNREFUSED - Server temporarily unavailable
// ECONNRESET - Connection dropped
// ETIMEDOUT - Network timeout
// ENETUNREACH - Network temporarily unreachable
// EHOSTUNREACH - Host temporarily unreachable
// EPIPE - Broken pipe
```

Maybe we should only retry a list of pre-defined ones (that we could discuss which). What do you think?

But I don't think this is a blocker, because the consequnce is just that we may send more requests until erroring out.

luislhl (approved):

PR #325, opened 2025-11-28
feat: token creation event

andreabadesso:
## Motivation

The fullnode now emits TOKEN_CREATED events when tokens are created, either via traditional CREATE_TOKEN_TX or nano contract syscalls. The daemon needs to handle these events to properly track token creation and manage token lifecycle during reorgs and voids.

## Acceptance Criteria

- Handle TOKEN_CREATED events and store token info in token table
- Track token-to-transaction mappings in new token_creation table (one tx can create multiple tokens)
- Delete nano-created tokens when `nc_execution` changes from success to `non-success`
- Delete ALL tokens (both types) when transaction that created it is voided
- Hybrid transactions (CREATE_TOKEN_TX + nano): preserve CREATE_TOKEN_TX token during reorg, delete only nano-created tokens

### Checklist


packages/daemon/src/services/index.ts
```diff
@@ -284,13 +330,6 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
       await unlockTimelockedUtxos(mysql, now);
     }
 
-    if (version === hathorLib.constants.CREATE_TOKEN_TX_VERSION) {
-      if (!token_name || !token_symbol) {
-        throw new Error('Processed a token creation event but it did not come with token name and symbol');
-      }
-      await storeTokenInformation(mysql, hash, token_name, token_symbol);
-    }
-
```
andreabadesso, line 298: This is now done in the `TOKEN_CREATED` event handling

packages/daemon/__tests__/services/services.test.ts
```diff
@@ -679,7 +683,7 @@ describe('handleVertexAccepted', () => {
 
     await handleVertexAccepted(context as any, {} as any);
 
-    expect(storeTokenInformation).toHaveBeenCalledWith(mockDb, hash, tokenName, tokenSymbol);
+    expect(storeTokenInformation).not.toHaveBeenCalled();
```
andreabadesso, line 688: We only insert to the token info table when the `TOKEN_CREATED` event is received

packages/daemon/src/services/index.ts
```diff
@@ -815,6 +975,52 @@ export const handleReorgStarted = async (context: Context): Promise<void> => {
...
+    await dbUpdateLastSyncedEvent(mysql, fullNodeEvent.event.id);
+
+    await mysql.commit();
+    logger.debug(`Successfully stored token ${token_uid} created by tx ${txId}`);
+  } catch (e) {
+    logger.error('Error handling TOKEN_CREATED event', e);
+    await mysql.rollback();
+    throw e;
+  } finally {
+    mysql.destroy();
```
r4mmer, line 1047: destroy the connection? shouldn't it be `quit` or `end`?

packages/daemon/src/services/index.ts
```diff
@@ -815,6 +975,52 @@ export const handleReorgStarted = async (context: Context): Promise<void> => {
...
+      token_uid,
+      token_name,
+      token_symbol,
+      nc_exec_info,
+    } = fullNodeEvent.event.data;
+
+    logger.debug(`Handling TOKEN_CREATED event for token ${token_uid}: ${token_name} (${token_symbol})`);
+
+    // Store the token information
+    await storeTokenInformation(mysql, token_uid, token_name, token_symbol);
```
r4mmer: Should we save token version?

packages/daemon/src/services/index.ts
```diff
@@ -601,6 +733,34 @@ export const voidTx = async (
...
+  /**
+   * Delete ALL tokens created by this voided transaction.
+   *
+   * This handles all three token creation scenarios:
+   *
+   * 1. Pure CREATE_TOKEN_TX (no nano):
+   *    - Deletes the single CREATE_TOKEN_TX token (token_id = tx_id)
+   *
+   * 2. Pure nano contract:
+   *    - Deletes all tokens created by nano syscalls (token_id ≠ tx_id)
```
r4mmer, line 707: How did you type a `≠`?

packages/daemon/src/services/index.ts
```diff
@@ -461,6 +500,62 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
...
+     * - Voiding deletes ALL tokens (handled by voidTx function)
+     * - This logic deletes ONLY nano-created tokens when execution state changes
+     * - A voided transaction might still have nc_execution = 'success'
+     */
+    const hasNanoHeaders = headers && headers.length > 0 && headers.some((h) => isNanoHeader(h));
+
+    if (hasNanoHeaders) {
+      const ncExecution = metadata?.nc_execution;
+
+      // If nc_execution is not 'success', delete any tokens created by this nano contract
```
r4mmer: Why? We may still execute the transaction if the state is pending or skipped, right?

packages/daemon/src/services/index.ts
```diff
@@ -601,6 +733,34 @@ export const voidTx = async (
...
+   * ENTIRE transaction is being voided.
+   *
+   * See handleVertexAccepted for nano execution state change logic, which
+   * ONLY deletes nano-created tokens when nc_execution becomes non-SUCCESS.
+   */
+  const tokensCreated = await getTokensCreatedByTx(mysql, hash);
+  if (tokensCreated.length > 0) {
+    logger.debug(`Voiding transaction ${hash} created ${tokensCreated.length} token(s), deleting them`);
+    await deleteTokens(mysql, tokensCreated);
+    await deleteTokenCreationMappings(mysql, tokensCreated);
```
r4mmer: Just checking, the `token_creation` table will delete the entry when the `token_id` entry on the token table is deleted (because of the onCascade).
So just deleting the token should delete the mappings as well, correct?
Maybe we should check how many entries the `deleteTokenCreationMappings` is affecting to see if the mappings are already deleted.

packages/daemon/src/services/index.ts
```diff
@@ -461,6 +500,62 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
...
+      const ncExecution = metadata?.nc_execution;
+
+      // If nc_execution is not 'success', delete any tokens created by this nano contract
+      if (ncExecution !== 'success') {
+        const tokensCreated = await getTokensCreatedByTx(mysql, hash);
+
+        if (tokensCreated.length > 0) {
+          logger.debug(`NC execution changed to ${ncExecution}, deleting ${tokensCreated.length} tokens created by tx ${hash}`);
+          await deleteTokens(mysql, tokensCreated);
+          await deleteTokenCreationMappings(mysql, tokensCreated);
```
r4mmer: same here

packages/daemon/src/services/index.ts
andreabadesso: Yes, but we can't consider the syscall token as created

If it goes from `success` -> `anything` -> `success` we will receive the `TOKEN_CREATED` event again and re-create the token we deleted

packages/daemon/src/services/index.ts
andreabadesso: It's unicode, claude did it

packages/daemon/src/services/index.ts
andreabadesso: If we reached this state, we already gracefully rolled back or commited, so the effect is the same, but I changed it to `end()` because there is no need to forcefully close the connection here

packages/daemon/src/services/index.ts
andreabadesso: Not in this PR

packages/daemon/src/services/index.ts
andreabadesso: You are correct, the `deleteTokenCreationMappings` is indeed redundant

Refactored it and added a test to make sure the cascade is working

Also removed the redudant `deleteTokenCreationMappins`

packages/daemon/src/services/index.ts
```diff
@@ -461,6 +499,61 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
...
+     * - A voided transaction might still have nc_execution = 'success'
+     */
+    const hasNanoHeaders = headers && headers.length > 0 && headers.some((h) => isNanoHeader(h));
+
+    if (hasNanoHeaders) {
+      const ncExecution = metadata?.nc_execution;
+
+      // If nc_execution is not 'success', delete any tokens created by this nano contract
+      if (ncExecution !== 'success') {
+        const tokensCreated = await getTokensCreatedByTx(mysql, hash);
```
pedroferreira1: For a token creation tx that has a nano execution, this would delete also the token created in the vertex tx. The tx might still be in the mempool after a reorg and only the nano execution that changed, so we should just get the tokens created by this tx in the nano execution.

packages/daemon/__tests__/integration/config.ts
```diff
@@ -16,7 +16,7 @@ export const UNVOIDED_SCENARIO_LAST_EVENT = 39;
 // reorg
 export const REORG_SCENARIO_PORT = 8082;
 // Same as the comment on the unvoided scenario last event
-export const REORG_SCENARIO_LAST_EVENT = 19;
+export const REORG_SCENARIO_LAST_EVENT = 18;
```
andreabadesso, line 19: This is unrelated to this PR, but the simulator events changed for the reorg scenario

packages/daemon/__tests__/machines/SyncMachine.test.ts
```diff
@@ -590,3 +590,83 @@ describe('Event handling', () => {
     expect(currentState.matches(`${SYNC_MACHINE_STATES.CONNECTED}.${CONNECTED_STATES.handlingReorgStarted}`)).toBeTruthy();
   });
 });
+
+describe('Error handling', () => {
```
andreabadesso, line 594: These tests come from https://github.com/HathorNetwork/hathor-wallet-service/pull/332, I cherry-picked the commit there

pedroferreira1 (approved):

r4mmer (approved):

PR #329, opened 2025-12-05
feat(wallet-service): added a new  param to the getUtxos endpoint to …

andreabadesso:
### Motivation

The totalAmount parameter returns UTXOs summing to at least the specified amount (for transaction building).

Users needed a way to query UTXOs with an upper bound, returning UTXOs summing to at most a specified amount.

Acceptance Criteria

- Add maxAmount query parameter to `/wallet/tx_outputs` and `/wallet/utxos` endpoints
- When maxAmount is provided, return UTXOs whose total value does not exceed the specified amount
- Return 400 error if both `totalAmount` and `maxAmount` are provided simultaneously

### Checklist


pedroferreira1 (dismissed):

packages/wallet-service/src/api/txOutputs.ts
```diff
@@ -221,6 +225,24 @@ const _getFilteredTxOutputs = async (walletId: string, filters: IFilterTxOutput)
     }
   }
 
+  // Apply maxAmount filter if specified (returns UTXOs summing to at most maxAmount)
+  if (filters.maxAmount) {
+    let accumulatedAmount = 0n;
+    const selectedTxOutputs: DbTxOutput[] = [];
+
+    // txOutputs are sorted by value DESC from the database, so we iterate
+    // from smallest to largest to maximize the number of UTXOs within the limit
```
r4mmer: Why do we want to "maximize" the number of utxos?
In the wallet-lib when an amount limit is given we usually try to select as few utxos as possible to avoid hitting the tx input limit (255)

andreabadesso: You are right, refactored it to invert the logic

packages/daemon/acceptance.md
r4mmer, line 1: I believe this should not be commited

packages/daemon/explanation.md
r4mmer, line 1: I believe this should not have been commited as well

pedroferreira1 (approved):

r4mmer (approved):

PR #338, opened 2026-01-07
feat: fee based tokens

raul-oliveira:
### Motivation
The fee-based token was added to the protocol and wallet-service should store, and provide, the token version to determine which type the token is.

### Acceptance Criteria

- Add the version column in the database with a migration
- Update the `TokenInfo` entity to handle the `version`
  - Add to sync-daemon workspace
  - Add to wallet-service workspace
- Update some test utils functions to handle the version field
- Update the sql statements that brings the token details
- Add the version field in the `getTokenInformation` API method.

### Checklist

### DIFF
--- a/packages/daemon/__tests__/services/services_with_db.test.ts
+++ b/packages/daemon/__tests__/services/services_with_db.test.ts
@@ -5,6 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import { TokenVersion } from '@hathor/wallet-lib';
 import * as db from '../../src/db';
 import { handleVoidedTx, voidTx, handleTokenCreated } from '../../src/services';
 import { LRU } from '../../src/utils';
@@ -961,9 +962,9 @@ describe('wallet balance voiding bug', () => {
     const tokenId3 = 'token003';
 
     // Add tokens to database
-    await db.storeTokenInformation(mysql, tokenId1, 'Token 1', 'TK1');
-    await db.storeTokenInformation(mysql, tokenId2, 'Token 2', 'TK2');
-    await db.storeTokenInformation(mysql, tokenId3, 'Token 3', 'TK3');
+    await db.storeTokenInformation(mysql, tokenId1, 'Token 1', 'TK1', TokenVersion.DEPOSIT);
+    await db.storeTokenInformation(mysql, tokenId2, 'Token 2', 'TK2', TokenVersion.DEPOSIT);
+    await db.storeTokenInformation(mysql, tokenId3, 'Token 3', 'TK3', TokenVersion.DEPOSIT);
 
     // Create mappings (simulate nano contract creating multiple tokens)
     await db.insertTokenCreation(mysql, tokenId1, txId, 'block-001');
@@ -1033,7 +1034,7 @@ describe('handleTokenCreated (db)', () => {
             },
             token_name: tokenName,
             token_symbol: tokenSymbol,
-            token_version: 'TOKEN_VERSION_1',
+            token_version: 1,
             initial_amount: 1000000,
           },
           group_id: null,
@@ -1092,7 +1093,7 @@ describe('handleTokenCreated (db)', () => {
             },
             token_name: 'Token 1',
             token_symbol: 'TK1',
-            token_version: 'TOKEN_VERSION_1',
+            token_version: 1,
             initial_amount: 1000000,
           },
           group_id: null,
@@ -1116,7 +1117,7 @@ describe('handleTokenCreated (db)', () => {
             },
             token_name: 'Token 2',
             token_symbol: 'TK2',
-            token_version: 'TOKEN_VERSION_1',
+            token_version: 1,
             initial_amount: 2000000,
           },
         },
@@ -1152,7 +1153,7 @@ describe('handleTokenCreated (db)', () => {
     const tokenSymbol = 'NCT';
 
     // First, create an existing token (simulating previous nano execution)
-    await db.storeTokenInformation(mysql, oldTokenId, tokenName, tokenSymbol);
+    await db.storeTokenInformation(mysql, oldTokenId, tokenName, tokenSymbol, TokenVersion.DEPOSIT);
     await db.insertTokenCreation(mysql, oldTokenId, txId, 'block-001');
 
     // Verify old token exists
@@ -1190,7 +1191,7 @@ describe('handleTokenCreated (db)', () => {
             },
             token_name: tokenName,
             token_symbol: tokenSymbol,
-            token_version: 'TOKEN_VERSION_1',
+            token_version: 1,
             initial_amount: 1000000,
           },
           group_id: 0,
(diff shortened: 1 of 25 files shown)


db/migrations/20251224000000-add-token-version.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('token', 'version', {
+      type: Sequelize.INTEGER,
+      allowNull: false,
+      defaultValue: 1, // Default to DEPOSIT (1) for existing tokens
```
raul-oliveira: only if this changes get applied before any fee-based token is created

db/migrations/20251224000000-add-token-version.js
```diff
@@ -0,0 +1,18 @@
...
+    });
+
+    // Set HTR token to NATIVE (version 0)
+    await queryInterface.sequelize.query(
+      "UPDATE token SET version = 0 WHERE id = '00'"
```
r4mmer: Nice

packages/common/__tests__/events/nftCreationTx.ts
```diff
@@ -22,7 +22,7 @@ import { HistoryTransaction } from '@hathor/wallet-lib/lib/models/types';
 export const nftCreationTx = {
   tx_id: '0025a6488045d7466639ead179a7f6beb188320f41cdb6df3a971db2ee86dbc3',
   signal_bits: 0,
-  version: 2,
+  version: 1,
```
r4mmer: This is not the token version, this is the transaction version.
Tx version 2 means a token creation version, which is the case for an NFT creation tx

packages/daemon/__tests__/services/services.test.ts
```diff
@@ -629,6 +630,7 @@ describe('handleVertexAccepted', () => {
     expect(mockDb.destroy).toHaveBeenCalled();
   });
 
+  // TODO-RAUL: add test for token creation tx with version 1 and version 2
```
r4mmer: Avoid commiting personal reminders.
Also, token creation is always version 2, if you mean token version 1 and 2 why don't we have this test?

packages/daemon/src/db/index.ts
```diff
@@ -1644,7 +1664,12 @@ export const getTokenInformation = async (
 
   if (results.length === 0) return null;
 
-  return new TokenInfo(tokenId, results[0].name as string, results[0].symbol as string);
+  return new TokenInfo(
+    tokenId,
+    results[0].name as string,
+    results[0].symbol as string,
+    results[0].version as TokenVersion,
```
r4mmer: `results[0].version` is a number and `TokenVersion` is an enum.
Apparently the best conversion from number to enum is a reverse lookup, since type casting to enum can be unsafe.

packages/daemon/src/services/index.ts
```diff
@@ -292,7 +294,7 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
       if (!token_name || !token_symbol) {
         throw new Error('Processed a token creation event but it did not come with token name and symbol');
       }
-      await storeTokenInformation(mysql, hash, token_name, token_symbol);
+      await storeTokenInformation(mysql, hash, token_name, token_symbol, TokenVersion.DEPOSIT);
```
r4mmer: Why not use the actual token version?

packages/wallet-service/tests/api.test.ts
```diff
@@ -425,18 +426,19 @@ test('GET /balances', async () => {
...
 
   // add the hathor token as it will be deleted by the beforeAll
-  const htrToken = { id: '00', name: 'Hathor', symbol: 'HTR' };
+  const htrToken = { id: '00', name: 'Hathor', symbol: 'HTR', version: TokenVersion.NATIVE };
   // add tokens
-  const token1 = { id: 'token1', name: 'MyToken1', symbol: 'MT1' };
-  const token2 = { id: 'token2', name: 'MyToken2', symbol: 'MT2' };
-  const token3 = { id: 'token3', name: 'MyToken3', symbol: 'MT3' };
-  const token4 = { id: 'token4', name: 'MyToken4', symbol: 'MT4' };
+  // TODO-RAUL: add test for token version 2
```
r4mmer: Avoid committing personal reminders

r4mmer (dismissed):
Did we not already have a handler for the token creation event?

Also, the handler only updates the token version, but if we use the token version from the tx we don't need this handler.

packages/common/__tests__/events/nftCreationTx.ts
tuliomir: This is a good point you bring up here. There can be confusion between token and tx version.
Maybe it's a good idea to clarify this in a _Troubleshooting_ section in our official docs.

raul-oliveira:
> Did we not already have a handler for the token creation event?
> 
> Also, the handler only updates the token version, but if we use the token version from the tx we don't need this handler.

It wasn't merged when I started this branch. Now it's up to date with master and have the `TOKEN_CREATED` handler as well.

packages/wallet-service/tests/api.test.ts
raul-oliveira: Deleted. tks

packages/daemon/src/services/index.ts
raul-oliveira: This is an outdated code. Currently we are using the `token_version`.

packages/common/__tests__/events/nftCreationTx.ts
raul-oliveira: Nice catch. I made a mistake. @tuliomir which docs are you referring to?

packages/daemon/__tests__/services/services.test.ts
raul-oliveira: Removed. This was a reminder for me to add tests for token creation tx with token version 1 and 2 (deposit and fee). I've added them.

packages/daemon/src/db/index.ts
raul-oliveira: Nice tip. It's done.

yarn.lock
pedroferreira1, line 1: Why do you have all these changes in this file? Which yarn are you using locally? You should use nix to run the same env.

raul-oliveira: Nice catch. It's fixed! now the wallet-lib was added from inside the nix develop environment.

pedroferreira1 (dismissed):

packages/daemon/src/types/db.ts
```diff
@@ -6,6 +6,7 @@
  */
 
 import { RowDataPacket } from 'mysql2/promise';
+import { TokenVersion } from '@hathor/wallet-lib';
```
tuliomir: suggestion(non-blocking): Remove unused variable?

packages/daemon/__tests__/services/services_with_db.test.ts
```diff
@@ -1033,7 +1034,7 @@ describe('handleTokenCreated (db)', () => {
             },
             token_name: tokenName,
             token_symbol: tokenSymbol,
-            token_version: 'TOKEN_VERSION_1',
+            token_version: 1,
```
tuliomir, line 1037: question(non-blocking): Shouldn't we use the enum here instead of a hardcoded number?
If so, we should also update the other `context` parameters along this file.

```suggestion
            token_version: TokenVersion.DEPOSIT,
```

If this breaks because of an incompatibility between `Number` and `TokenVersion`, maybe:
```ts
            token_version: Number(TokenVersion.DEPOSIT)
```

tuliomir (dismissed):

packages/daemon/__tests__/services/services_with_db.test.ts
raul-oliveira: In this case we are simulating a event from the fullnode with a flat value. The enum is dynamic, let's imagine some refactor happens in the future change the DEPOSIT value to 3, the test will pass but won't reflect the real event.

That's why I choose to use the number directly.

packages/daemon/src/types/db.ts
raul-oliveira: Done! Tks

tuliomir (approved):

pedroferreira1 (approved):

PR #339, opened 2026-01-08
feat(wallet-service): added an api to find out if a wallet is using m…

andreabadesso:
### Motivation

The idea here is to allow the web-wallet to reject user requests to use the "single address" feature if the user wallet has more than one address being used

### Acceptance Criteria

- We must have an API that returns whether a wallet has transactions in addresses other than the first one

### Checklist


packages/wallet-service/serverless.yml
```diff
@@ -352,6 +352,17 @@ functions:
     warmup:
       walletWarmer:
         enabled: true
+  hasNonFirstAddressTransactions:
+    handler: src/api/hasNonFirstAddressTransactions.get
+    events:
+      - http:
+          path: wallet/addresses/has-transactions
```
tuliomir: suggestion: The `has-transactions` endpoint name is misleading. I would expect it to return just if a wallet has been used at all, not if it is a single address wallet.

I would suggest something along the lines of:
- `has-many-addresses` or just `wallet/addresses/has-many`
- `has-transactions-on-multiple-addresses`
- `is-single-address-wallet`
- `is-candidate-single-address-wallet`
- `has-transactions-outside-first-address`

tuliomir (dismissed):

pedroferreira1 (dismissed):

tuliomir (approved):

pedroferreira1 (approved):

PR #350, opened 2026-01-23
feat: store first_block on transactions

andreabadesso:
This PR handles `first_block` changes (by storing them in the `transaction` table) and also handles multiple changes in a transaction metadata in the same event

### Motivation

After adding the `first_block` handling mechanism, we stopped handling `nc_execution` changes, upon investigation I've noticed that we were expecting one `METADATA_CHANGED` event for each change which is not what actually happens -- we receive a single metadata changed event with all changes to metadata

### Acceptance Criteria

- `metadataDiff` detects all independent metadata changes in a single event and returns them as an array
- A dispatch queue processes each change one-by-one before returning to idle
- `handleNcExecVoided` no longer needs to know about `first_block` — each handler has a single responsibility
- `first_block` is now a `transaction` column and gets handled by the new metadata diff 

### Checklist


packages/daemon/__tests__/utils.ts
```diff
@@ -336,22 +338,25 @@ export const checkTransactionTable = async (
...
+  [results] = await mysql.query<TransactionTableRow[]>(
+    `${baseQuery} AND \`first_block\` ${firstBlock !== null ? '= ?' : 'IS ?'}`,
+    [txId, timestamp, version, voided, height, firstBlock],
   );
 
   if (results.length !== 1) {
     return {
-      error: 'checkAddressTable query',
-      params: { txId, timestamp, version, voided, height },
+      error: 'checkTransactionTable query',
```
andreabadesso, line 358: Thanks Claude

r4mmer (dismissed):

packages/daemon/src/services/index.ts
```diff
@@ -128,75 +129,69 @@ export const metadataDiff = async (_context: Context, event: Event) => {
...
-              originalEvent: event,
-            };
-          }
+          // Independent changes: collect all into array
+          const types: string[] = [];
 
           // Check if nc_execution changed from 'success' to something else.
-          // If the tx has nano-created tokens in the database (tokens where token_id != tx_id),
-          // those tokens were created when nc_execution was 'success'.
-          // If nc_execution is now NOT 'success', we should delete those tokens.
```
pedroferreira1, line 185: Why did you remove this comment?

pedroferreira1 (dismissed):

pedroferreira1 (approved):

r4mmer (approved):

PR #369, opened 2026-02-23
feat(daemon): add monitoring layer to the sync state machine

andreabadesso:
## Summary

- Adds a **`MonitoringActor`** that centralises runtime health checks for the sync state machine.
- Four new optional env vars: `IDLE_EVENT_TIMEOUT_MS`, `STUCK_PROCESSING_TIMEOUT_MS`, `RECONNECTION_STORM_THRESHOLD`, `RECONNECTION_STORM_WINDOW_MS`.

| Condition | Response | Default threshold |
|---|---|---|
| No event received while WS connected | MAJOR alert + `process.exit(1)` (Kubernetes restarts) | 5 min |
| Stuck in any processing state | MAJOR alert only — machine keeps running | 1 hour |
| ≥ threshold reconnects within window | MAJOR alert | 10 reconnects / 5 min |

The stuck-processing check intentionally does **not** force a reconnection — a long-running handler (e.g. a large reorg) should be allowed to finish. Operators are alerted so they can investigate if needed.

## Acceptance criteria

- [ ] Idle timeout terminates the process; a MAJOR alert is sent before exit.
- [ ] Stuck-processing fires a MAJOR alert after 1 hour; machine continues running uninterrupted.
- [ ] Reconnection storm alert fires at threshold; stale entries outside the window are evicted.
- [ ] All four env vars are optional — daemon starts correctly when unset.
- [ ] All `MonitoringActor` unit tests pass; existing tests unaffected.

## Summary by CodeRabbit

* **New Features**
  * Runtime health monitoring for the sync process: idle detection, stuck-processing alerts, and reconnection-storm detection; sync lifecycle now emits monitoring signals.
  * New configurable monitoring settings (timeouts, thresholds, windows) with sensible defaults.

* **Tests**
  * Comprehensive tests covering idle/stuck/reconnection behaviors, timer lifecycles, alerting, cooldowns, and logging.
  * Test setup updated with monitoring config values and mocks to prevent external side effects.


packages/daemon/src/actions/index.ts
```diff
@@ -197,3 +197,66 @@ export const stopHealthcheckPing = sendTo(
...
  */
 export const logEventError = (_context: Context, event: Event) => logger.error(bigIntUtils.JSONBigInt.stringify(event));
+
+/*
+ * This is a helper to get the monitoring ref from the context and throw if it's not found.
+ */
+export const getMonitoringRefFromContext = (context: Context) => {
+  if (!context.monitoring) {
+    throw new Error('No monitoring actor in context');
+  }
```
r4mmer, line 207: issue (blocking): Throwing in `getMonitoringRefFromContext` can crash the machine before `monitoring` is spawned

Do we know the effects of the raised exception on the machine?

Also, is it possible to do something like "no monitoring -> start monitor -> resume machine"?

packages/daemon/src/actors/MonitoringActor.ts
```diff
@@ -0,0 +1,193 @@
...
+          `[monitoring] No fullnode events received for ${idleMinutes} minutes while WebSocket is connected — terminating`,
+        );
+        addAlert(
+          'Daemon Idle — No Events Received',
+          `No fullnode events received for ${idleMinutes} minute(s) while the WebSocket is connected. ` +
+            'Terminating the process so Kubernetes can restart it.',
+          Severity.MAJOR,
+          { idleMs: String(idleMs) },
+          logger,
+        ).finally(() => process.exit(1));
```
r4mmer: issue (blocking): Idle detector terminates the process from inside the actor

Having a hard shutdown can be very hard to determine the cause of the issue, can we send an event like `MONITORING_IDLE_TIMEOUT` that logs/add alert and proceeds to the normal daemon shutdown?

packages/daemon/src/actors/MonitoringActor.ts
```diff
@@ -0,0 +1,193 @@
...
+      }
+    }, idleTimeoutMs);
```
r4mmer: suggestion: Idle detection interval being equal to the desired idle time may be too gracious.

If interval is the same as the timeout we may wait up to 2 times the timeout before triggering the timeout.
To mitigate this we could have the setInteval timeout be `idleTimeoutMs/2` so we will hang for at most "timeout * 1.5".

packages/daemon/src/actors/MonitoringActor.ts
```diff
@@ -0,0 +1,193 @@
...
+ * back to the machine to trigger a soft reconnect.
+ *
+ * Responsibilities:
+ *
+ * 1. Idle-stream detection — no EVENT_RECEIVED for >IDLE_EVENT_TIMEOUT_MS while
+ *    connected fires a MAJOR alert so operators know the fullnode stream may have
+ *    stalled without triggering a WebSocket disconnect.
+ *
+ * 2. Stuck-processing detection — PROCESSING_STARTED begins a one-shot timer;
+ *    PROCESSING_COMPLETED cancels it.  If the timer fires the actor fires a
```
r4mmer, line 28: question: is it the timer or the actor that will fire a `CRITICAL` alert?

packages/daemon/src/actors/MonitoringActor.ts
```diff
@@ -0,0 +1,193 @@
...
+    if (idleCheckTimer) {
+      clearInterval(idleCheckTimer);
+      idleCheckTimer = null;
+    }
+  };
+
+  // ── Stuck-processing detection ───────────────────────────────────────────────
+  let stuckTimer: ReturnType<typeof setTimeout> | null = null;
+
+  const startStuckTimer = () => {
```
r4mmer, line 91: issue (blocking): Stuck-processing logic does not implement the actor documentation comment because it does not notify the machine

The header comment says it “fires a CRITICAL alert and sends MONITORING_STUCK_PROCESSING back to the machine”, but the implementation only sends an alert (Severity.MAJOR).

packages/daemon/src/actors/MonitoringActor.ts
```diff
@@ -0,0 +1,193 @@
...
+        `${reconnectionTimestamps.length} reconnections occurred in the last ${windowMinutes} minute(s). ` +
+          'The daemon may be stuck in a reconnection loop.',
+        Severity.MAJOR,
+        {
+          reconnectionCount: String(reconnectionTimestamps.length),
+          windowMinutes: String(windowMinutes),
+        },
+        logger,
+      ).catch((err: Error) =>
+        logger.error(`[monitoring] Failed to send reconnection storm alert: ${err}`),
```
r4mmer: question: Will the `err` string interpolation use `err.message` or `[object Object]`?

packages/daemon/src/actors/MonitoringActor.ts
```diff
@@ -0,0 +1,193 @@
...
+  let reconnectionTimestamps: number[] = [];
+
+  const trackReconnection = () => {
+    const now = Date.now();
+    reconnectionTimestamps.push(now);
+
+    const windowStart = now - stormWindowMs;
+    reconnectionTimestamps = reconnectionTimestamps.filter(t => t >= windowStart);
+
+    if (reconnectionTimestamps.length >= stormThreshold) {
```
r4mmer, line 127: issue(blocking): Reconnection storm will spam the alerting system.

If we trigger a reconection storm, each new "RECONNECTING" will trigger a new alert, which will spam the alerting system.

We could create a `stormAlertLastFiredAt` and only fire a new alert if we have not fired for a minute.

packages/daemon/src/delays/index.ts
```diff
@@ -24,3 +24,4 @@ export const ACK_TIMEOUT = () => {
   const { ACK_TIMEOUT_MS } = getConfig();
   return ACK_TIMEOUT_MS;
 };
+
```
r4mmer: nitpick: clean empty line

packages/daemon/src/machines/SyncMachine.ts
```diff
@@ -130,15 +138,21 @@ export const SyncMachine = Machine<Context, any, Event>({
       },
     },
     [SYNC_MACHINE_STATES.RECONNECTING]: {
-      onEntry: ['clearSocket', 'increaseRetry', 'stopHealthcheckPing'],
+      onEntry: [
+        'clearSocket',
+        'increaseRetry',
+        'stopHealthcheckPing',
+        'sendMonitoringReconnecting',
+        'sendMonitoringDisconnected',
```
r4mmer, line 156: question: Are we sure these 2 actions should be triggered on the same state?

packages/daemon/src/machines/SyncMachine.ts
```diff
@@ -158,44 +172,46 @@ export const SyncMachine = Machine<Context, any, Event>({
               cond: 'invalidNetwork',
               target: `#${SYNC_MACHINE_STATES.ERROR}`,
             }, {
-              actions: ['storeEvent', 'sendAck'],
+              actions: ['storeEvent', 'sendAck', 'sendMonitoringEventReceived'],
```
r4mmer, line 185: question: Do we have a better way to trigger `sendMonitoringEventReceived` on required transitions?

Marking on every transition may become an issue if we have more events in the future that we forget to add this.

packages/daemon/src/machines/SyncMachine.ts
andreabadesso: Yes, it's disconnected and reconnecting

packages/daemon/src/actions/index.ts
andreabadesso: I refactored it to use `choose` and fail silently instead of throwing

This is more defensive than anything as there is no case where monitoring will be null here

I don't think we should do (at least in this PR) a more elaborate mechanism of pausing the machine and resuming the monitor if it's null

packages/daemon/src/actors/MonitoringActor.ts
andreabadesso: I like the idea, done in 597a0da

Moved the `process.exit(1)` to the index.js 

Basically what happens now is that the "final" ERROR state is reached, that sends the alert through SQS, logs and when it resolves, the `machine.onDone()` callback is called, terminating the process

packages/daemon/src/actors/MonitoringActor.ts
andreabadesso: Done in 597a0da

packages/daemon/src/actors/MonitoringActor.ts
andreabadesso: Updated the comment, thanks

packages/daemon/src/actors/MonitoringActor.ts
andreabadesso: It would call Error.prototype.toString() which would produce `Error: <message>`, but I've changed it to use `err.message` directly

packages/daemon/src/actors/MonitoringActor.ts
andreabadesso: Nice suggestion, thanks

Done

packages/daemon/src/delays/index.ts
andreabadesso: Done

packages/daemon/src/machines/SyncMachine.ts
andreabadesso: Unfortunately not! In XState v4, there is no clean mechanism to intercept all transitions through a particular state without adding the action explicitly per-transition

packages/daemon/src/actors/MonitoringActor.ts
```diff
@@ -0,0 +1,203 @@
...
+ *
+ * 3. Reconnection storm detection — fires a MAJOR alert when the daemon
+ *    reconnects more than RECONNECTION_STORM_THRESHOLD times within
+ *    RECONNECTION_STORM_WINDOW_MS.  Duplicate alerts are suppressed for
+ *    STORM_ALERT_COOLDOWN_MS (1 min) to avoid spamming the alerting system.
+ */
+const DEFAULT_IDLE_EVENT_TIMEOUT_MS = 5 * 60 * 1000;
+const DEFAULT_STUCK_PROCESSING_TIMEOUT_MS = 60 * 60 * 1000; // 1 hour
+const DEFAULT_RECONNECTION_STORM_THRESHOLD = 10;
+const DEFAULT_RECONNECTION_STORM_WINDOW_MS = 5 * 60 * 1000;
```
pedroferreira1: You already have these default values in the config file. Do you need it here as well?

packages/daemon/src/machines/SyncMachine.ts
```diff
@@ -335,6 +369,8 @@ export const SyncMachine = Machine<Context, any, Event>({
         },
         [CONNECTED_STATES.checkingForMissedEvents]: {
           id: CONNECTED_STATES.checkingForMissedEvents,
+          entry: ['sendMonitoringProcessingStarted'],
+          exit: ['sendMonitoringProcessingCompleted'],
```
pedroferreira1: Would be great to have "categories", so you can say that all states in a category X will have entry Y and exit Z, so you don't need to repeat these lines in all of them

pedroferreira1 (dismissed):

packages/daemon/src/actors/MonitoringActor.ts
andreabadesso: Removed the redundant defaults, config already handles them.

packages/daemon/src/machines/SyncMachine.ts
andreabadesso: Added a `monitoredState()` helper that wraps state definitions with the monitoring entry/exit actions. New processing states just need to be wrapped with it.

pedroferreira1 (approved):

r4mmer (approved):

PR #376, opened 2026-03-05
perf(daemon): batch DB queries in handlingVoidedTx

andreabadesso:
### Motivation

The `handlingVoidedTx` handler is slow due to N+1 query patterns. A transaction with many inputs and (address, token) pairs triggers hundreds of sequential DB round-trips. This PR batches the heaviest queries.

### Acceptance Criteria

- Batch input lookups in `voidTx` using `getTxOutputs` instead of per-input `getTxOutput` (N queries → 1)
- Batch `voidAddressTransaction`: single UPDATE with CASE WHEN via `buildBatchCaseUpdate` helper, single authority recalc, single zero-balance cleanup (~4*M → 3)
- Batch `voidWalletTransaction`: same pattern (~3*M → 3)
- Add `VALIDATE_ADDRESS_BALANCES` env var (default `true`) to optionally skip post-void validation

### Benchmark — local A/B (this branch vs `master`)

Reproducible harness committed in this PR: `packages/daemon/src/scripts/bench-void-tx.ts` (seeds a synthetic voidable tx, runs `voidTx` with OTel spans captured, re-seeds per run) + `compare-bench.ts` (side-by-side report). Same host, same MySQL 8.0 on `localhost:3306` (from `docker-compose.dev.yml`), same DB state per iteration, traces exported to Jaeger for inspection.

#### 255 inputs / 255 address-token pairs — 10 runs + 2 warmup (matches our target prod scenario)

**Total `voidTx` wall-clock**

| | master | branch | delta | speedup |
|---|---:|---:|---:|---:|
| mean | 708.2 ms | 106.5 ms | −85.0% | **6.65×** |
| p50  | 715.3 ms | 104.4 ms | −85.4% | 6.85× |
| p95  | 759.8 ms | 118.7 ms | −84.4% | 6.40× |
| min  | 664.1 ms | 100.2 ms | −84.9% | 6.63× |
| max  | 759.8 ms | 118.7 ms | −84.4% | 6.40× |

**Per-span breakdown** (mean summed within a run)

| span | master | branch | speedup |
|---|---:|---:|---:|
| `voidAddressTransaction` | 530.0 ms | 53.8 ms | **9.85×** |
| `unspendInputs` | 147.8 ms | 21.5 ms | **6.88×** |
| `clearTxProposalForVoidedTx` | 6.8 ms | 6.9 ms | 0.98× (flat — not touched by this PR) |
| `markUtxosAsVoided` | 2.3 ms | 2.2 ms | 1.02× (not touched) |
| `voidTransaction` | 1.7 ms | 1.6 ms | 1.06× (not touched) |
| `getTxOutputsFromTx` | 1.1 ms | 0.9 ms | 1.28× (noise) |
| `voidWalletTransaction` | 2.5 ms | 2.8 ms | 0.90× *(bench seeds no wallets — this path is effectively a no-op here; same `buildBatchCaseUpdate` pattern as `voidAddressTransaction` so a real wallet workload should track that span's ~10× win)* |

Untouched spans correctly hover at ~1× — confirms the measurement isolates the batched paths.

#### 50 inputs / 50 pairs — scale-sensitivity check (5 runs + 1 warmup)

| | master | branch | speedup |
|---|---:|---:|---:|
| Total mean | 154.1 ms | 26.0 ms | **5.92×** |
| `voidAddressTransaction` | 113.4 ms | 10.5 ms | **10.85×** |
| `unspendInputs` | 29.2 ms | 4.9 ms | **5.92×** |

Speedup ratio on batched paths is stable across scales; absolute wins grow with N as expected from N+1 → 1 batching.

#### Caveats

- Local numbers are smaller than the mainnet-staging benchmarks previously reported (622ms → 430ms input lookups, 2612ms → 1280ms `voidAddressTransaction`) because mainnet-staging has production-sized per-row data and network latency; the **ratios** are what transfer. The earlier mainnet-staging benchmark also measured the `VALIDATE_ADDRESS_BALANCES=false` win (≈2081 ms saved per void), which is not exercised here since the harness calls `voidTx` directly.
- `voidWalletTransaction` is not exercised by the harness (no wallets seeded). The code change there mirrors `voidAddressTransaction` 1:1, so its speedup should track.

#### Tests

- Daemon unit suite: **214/214 pass** (10 suites)
- Daemon integration suite: **20/20 pass** (3 suites) against fullnode event simulators + MySQL on :3380

### Checklist

## Summary by CodeRabbit

* **Refactor**
  * Transaction voiding and balance reconciliation now use batched DB updates and bulk cleanup for fewer operations and improved performance.
* **New Features**
  * Added a runtime flag to enable or disable address balance validation.
  * Added a batched fetch endpoint for transaction outputs to speed multi-input processing.
* **Chores / Tests**
  * Updated tests and mocks to align with the new public API and batched flows.


packages/daemon/src/db/index.ts
```diff
@@ -401,105 +402,117 @@ export const voidAddressTransaction = async (
...
+  for (const { address, token, balance } of pairs) {
+    updateSql += ` WHEN \`address\` = ? AND \`token_id\` = ? THEN ?`;
+    updateParams.push(address, token, balance.unlockedAuthorities.toUnsignedInteger());
+  }
+  updateSql += ` ELSE 0 END,
+    \`locked_authorities\` = \`locked_authorities\` | CASE`;
+  for (const { address, token, balance } of pairs) {
+    updateSql += ` WHEN \`address\` = ? AND \`token_id\` = ? THEN ?`;
+    updateParams.push(address, token, balance.lockedAuthorities.toUnsignedInteger());
+  }
```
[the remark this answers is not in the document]
andreabadesso: This is not a bug introduced by this PR. The original code had the same behavior — `locked_authorities` was only ORed and never recalculated from UTXOs. The original code had explicit comments explaining the rationale:

> for locked authorities, it doesn't make sense to perform the same operation. The authority needs to be unlocked before it can be spent.

> NOTE: No need to do the same for locked authorities as they can't be spent before being unlocked and we trust the fullnode

These comments were removed during the refactor since the logic is now in the `buildBatchCaseUpdate` helper, but the behavior is preserved exactly as before.

andreabadesso:
Follow-up: #380 proposes a scheduled balance validation job so we can safely disable per-void validation (`VALIDATE_ADDRESS_BALANCES=false`) without losing the safety net.

packages/daemon/src/config.ts
```diff
@@ -100,6 +100,9 @@ export const RECONNECTION_STORM_WINDOW_MS = parseInt(process.env.RECONNECTION_ST
 // Other
 export const USE_SSL = process.env.USE_SSL === 'true';
 
+// When false, skips the address balance validation after voiding transactions
+export const VALIDATE_ADDRESS_BALANCES = process.env.VALIDATE_ADDRESS_BALANCES !== 'false';
```
pedroferreira1, line 104: Should we add this to the default export below, as the others?

packages/daemon/src/services/index.ts
```diff
@@ -723,34 +723,16 @@ export const voidTx = async (
...
-          timelock: null, // Not needed for unspending
-          heightlock: null, // Not needed for unspending
-          locked: false, // Not needed for unspending
-          spentBy: hash, // This is what we're unsetting
-          voided: false, // Not needed for unspending
-        });
-      }
-    }
+      // Batch-fetch all input outputs in a single query, then filter by spentBy
+      const allInputOutputs = await getTxOutputs(
```
pedroferreira1, line 727: The previous used method `getTxOutput` had another db where clause `voided = FALSE`. Should we add an optional parameter in this method to have the same behavior as before?

packages/daemon/src/db/index.ts
```diff
@@ -42,6 +43,78 @@ import {
...
+      return `${colPrefix}${col.column} = CASE${caseExpr} ELSE ${colPrefix}${col.column} END`;
+    }
+  });
+
+  const keyPairs = pairs.map(getKeys);
+  params.push(keyPairs);
+
+  const sql = `UPDATE ${table}${alias} SET
+    ${setClauses.join(',\n    ')},
+    ${colPrefix}\`transactions\` = ${colPrefix}\`transactions\` - 1
```
luislhl, line 112: This helper is named like a generic batch-update builder, but it currently hardcodes `transactions = transactions - 1`, which makes it specific to voiding logic and creates a hidden side effect.

Proposed solution:
- keep `buildBatchCaseUpdate` generic by removing the fixed `transactions` decrement
- add a flag or extra parameter so callers can opt in to additional fixed SET clauses
- then apply `transactions = transactions - 1` explicitly in the voiding caller only, not inside the generic helper

packages/daemon/src/db/index.ts
```diff
@@ -401,105 +474,87 @@ export const voidAddressTransaction = async (
...
-      if (isCreateTokenTx) {
-        // The transaction that created the token was voided, so we can remove
-        // it from the tokens table as well.
-        await mysql.query(
-          `DELETE FROM token
-            WHERE id = ?`,
-          [token]
-        );
-      }
-    }
```
luislhl, line 500: question: What happened to this token deletion logic? Is it not needed anymore?

I thought this PR was trying to optimize performance without changing behavior. This seems like a behavior change.

packages/daemon/src/db/index.ts
```diff
@@ -401,105 +474,87 @@ export const voidAddressTransaction = async (
...
+  if (updateResult.affectedRows < pairs.length) {
+    console.warn(`warning: Voiding transaction ${txId}: updated ${updateResult.affectedRows} address_balance rows but expected ${pairs.length}`);
```
luislhl, line 521: This is here to replace this old warning, right?

```
console.warn(`warning: Trying to void transaction for address ${address} token ${token} but no balance entry exists`);
```

Could you update the message (or add comments) to explain a little better what is the meaning of having less rows than expected? Is it a possible previous error during sync?

andreabadesso:
Addressed the review feedback in f14e0b1.

**@pedroferreira1 + Copilot — `voidTx` / `getTxOutputs` missing `voided = FALSE`.**
Real correctness bug confirmed and fixed in `db/index.ts`. The old `getTxOutput(..., skipSpent=false)` always filtered voided rows; the batched replacement now does too, so `unspendUtxos` can't clear `spent_by` on already-voided rows.

**@pedroferreira1 — `VALIDATE_ADDRESS_BALANCES` missing from the default config export.**
Added to `config.ts`.

**@luislhl — `buildBatchCaseUpdate` hardcoding `transactions = transactions - 1`.**
Fixed — the decrement is now an explicit opt-in column def in both callers:
```ts
{ column: '`transactions`', op: 'subtract', getValue: () => 1 }
```
The helper generates the same `transactions = transactions - CASE WHEN ... THEN 1 ELSE 0 END` shape, which evaluates to `-1` for every matched row (one WHEN per pair). Semantically identical to the old literal; no regression expected on the benchmark.

**@luislhl — the old per-pair \"no balance entry exists\" warning replaced by a single count.**
Yes, that was the replacement, and you're right that the new message was too terse. Rewrote it to explain *why* a mismatch happens (sync inconsistency vs. already-partially-voided) and to include a truncated sample of the (address, token) pairs that were checked — capped at 10 to avoid flooding logs on large txs.

**@luislhl — \"what happened to this token deletion logic?\"**
It moved and was generalized. The old code in `voidAddressTransaction` matched on `version === CREATE_TOKEN_TX_VERSION` and deleted by `token = tx_id`, which only covers classic CREATE_TOKEN_TX transactions. The new code in `voidTx` (services/index.ts:767) calls `getTokensCreatedByTx()` which reads the `token_creation` mapping table and deletes *every* token this tx created — covering both CREATE_TOKEN_TX and nano-created tokens, including hybrid transactions. That's a strict superset of the old behavior. I added the rationale in the inline docblock above the new call, but happy to pull it out into a migration note if that'd be clearer.

**Copilot — `bench-void-tx.ts` `cleanDb()` safety.**
Added a hard guard: requires `--dangerously-reset-db` to opt in, AND rejects any `DB_ENDPOINT` that isn't `localhost` / `127.0.0.1` / `::1` / `host.docker.internal`. A misconfigured env can no longer silently wipe a real DB.

Type check clean, 135/135 non-DB unit tests pass. DB integration tests weren't run (no local MySQL), but the changes are conservative — matching the old semantics for the voided filter and preserving the SQL shape for the batched update.

packages/daemon/src/services/index.ts
```diff
@@ -723,34 +723,16 @@ export const voidTx = async (
...
+        mysql,
+        inputs.map((input) => ({ txId: input.tx_id, index: input.index })),
+      );
+      const inputsSpentByThisTx = allInputOutputs.filter((output) => output.spentBy === hash);
```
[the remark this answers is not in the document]
andreabadesso: Fixed at the source in f14e0b1 — added `AND `voided` = FALSE` to the `getTxOutputs` SQL itself (db/index.ts:333-337), which matches the old `getTxOutput(..., skipSpent=false)` semantics and keeps the voided filter applied to every caller automatically. The inline filter Copilot suggested would also work, but filtering at the DB avoids fetching rows we're going to discard and makes the caller site the same shape as the old code.

packages/daemon/src/config.ts
andreabadesso: Done in f14e0b1 — `VALIDATE_ADDRESS_BALANCES` added to the default export.

packages/daemon/src/services/index.ts
andreabadesso: Fixed in f14e0b1 — added `AND `voided` = FALSE` to `getTxOutputs` itself (db/index.ts:333-337), which restores the old `getTxOutput(..., skipSpent=false)` semantics without needing a new optional parameter. Every caller gets it automatically.

packages/daemon/src/scripts/bench-void-tx.ts
```diff
@@ -0,0 +1,399 @@
...
+    'address_tx_history', 'address_balance', 'address',
+    'tx_output', 'transaction', 'token',
+    'token_creation', 'tx_proposal',
+  ];
+  await conn.query('SET FOREIGN_KEY_CHECKS = 0');
+  for (const t of tables) {
+    await conn.query(`DELETE FROM \`${t}\``);
+  }
+  await conn.query('SET FOREIGN_KEY_CHECKS = 1');
+}
```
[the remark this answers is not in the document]
andreabadesso: Guarded in f14e0b1. `cleanDb()` now requires `--dangerously-reset-db` to opt in AND rejects any `DB_ENDPOINT` that isn't a recognized local host (localhost / 127.0.0.1 / ::1 / host.docker.internal). A misconfigured env can't silently wipe a real DB anymore.

packages/daemon/src/db/index.ts
andreabadesso: Fixed in f14e0b1. `buildBatchCaseUpdate` no longer hardcodes `transactions = transactions - 1` — the helper is generic again. Both voiding callers now pass the decrement explicitly via a column def:
```ts
{ column: '`transactions`', op: 'subtract', getValue: () => 1 }
```
Generated SQL is semantically identical to the old literal — one `WHEN ... THEN 1` per matched pair, `ELSE 0`.

packages/daemon/src/db/index.ts
andreabadesso: Not a behavior change — it moved and generalized. The old code here matched `version === CREATE_TOKEN_TX_VERSION` and deleted by `token = tx_id`, which only covers classic CREATE_TOKEN_TX. The new code lives in `voidTx` at [services/index.ts:767](https://github.com/HathorNetwork/hathor-wallet-service/blob/perf/batch-void-tx-queries/packages/daemon/src/services/index.ts#L767) and calls `getTokensCreatedByTx()` against the `token_creation` mapping table, so it deletes **every** token this tx created — both CREATE_TOKEN_TX and nano-created tokens, including hybrid transactions. Strict superset of the old behavior; the inline docblock above the new call explains the three scenarios.

packages/daemon/src/db/index.ts
andreabadesso: Yes, that's the replacement. Reworked the message in f14e0b1 to explain what an `affectedRows < pairs.length` result actually means:

> This usually indicates a data inconsistency — the balance row was never added, or the tx was already (partially) voided.

The warning now also includes a truncated sample of the pairs checked (first 10, with a `+N more` suffix) so operators can grep for the tx in logs without flooding on large txs. One caveat: the sample is the first 10 pairs we *tried*, not the 10 that failed — MySQL doesn't tell us which entries in the `IN(?)` set didn't match. If that's worth pursuing, happy to add a follow-up `SELECT` to pinpoint the missing ones at the cost of an extra round trip.

packages/daemon/src/db/index.ts
```diff
@@ -328,10 +327,14 @@ export const getTxOutputs = async (
...
   const txIdIndexPair = inputs.map((utxo) => [utxo.txId, utxo.index]);
+  // `voided = FALSE` matches the semantics of the per-input `getTxOutput` this
+  // function replaced on the unspend-inputs path. Without it, `voidTx` could
+  // clear `spent_by` on rows that are already voided.
   const [results] = await mysql.query(
     `SELECT *
        FROM \`tx_output\`
-      WHERE (\`tx_id\`, \`index\`) IN (?)`,
+      WHERE (\`tx_id\`, \`index\`) IN (?)
+        AND \`voided\` = FALSE`,
```
pedroferreira1: It's better to create a parameter to add this filter, so we don't change the behavior of this method, that is used in other places

andreabadesso: Good call, fixed in 8ca4470. Added a `skipVoided` boolean parameter on `getTxOutputs` (default `false`, preserves the original signature's behavior for any future caller) and `voidTx` now passes `true` explicitly. Mirrors the `skipSpent` pattern already used by `getTxOutput`.

pedroferreira1 (approved):

luislhl (approved):

PR #383, opened 2026-03-14
feat: add OpenTelemetry distributed tracing to daemon and Lambdas

andreabadesso:
Implements [RFC 0001-opentelemetry-observability](https://github.com/HathorNetwork/rfcs/pull/105) for both the daemon (K8s) and wallet-service (Lambda) packages.

### Daemon (`packages/daemon`)
- OTel SDK init in `tracing.ts` with `BatchSpanProcessor` + OTLP HTTP exporter
- Auto-instrumentation for mysql2, http, aws-sdk (fs/dns disabled)
- Manual spans on all critical event handlers (`handleVertexAccepted`, `handleVoidedTx`, `handleReorgStarted`, etc.) via `withSpan()` helper
- Root `event.processing.*` spans via `tracedService()` wrapper in SyncMachine — service-level spans become children automatically
- Winston log correlation: `trace_id` and `span_id` injected into log output
- Graceful shutdown on SIGTERM with span flush
- Local dev setup: `docker-compose.dev.yml` with MySQL + Jaeger

### Wallet-service Lambdas (`packages/wallet-service`)
- Manual SDK init (RFC Option B) — ADOT layer was too large for the 250MB Lambda size limit
- Cherry-picked instrumentations (aws-lambda, aws-sdk, http, mysql, redis, winston) adding ~39MB vs ~120MB for the full meta-package
- Webpack entries prepended with `tracing.ts` so auto-instrumentation patches libraries before import
- `OTEL_SDK_DISABLED=true` kill switch to skip all OTel module loading
- `OTEL_CONSOLE_DEBUG=true` to print spans to CloudWatch logs for debugging

### Bug fix (included)
- Fix `lockExpires` Joi validation in push notification handler (`.valid(null)` → `.allow(null)`) — fixes Opsgenie alert #50062

## Cold Start Benchmark

A/B test on `getLatestBlock` (n=7 each, forced cold starts):

| Metric | WITHOUT OTel | WITH OTel | Delta |
|--------|------------:|----------:|------:|
| **Avg Init** | 1164 ms | 1870 ms | **+706 ms (+60.7%)** |

Warm handler overhead (n=10, console debug OFF):

| Metric | OTel OFF | OTel ON | Delta |
|--------|--------:|-------:|------:|
| **Median** | 9.6 ms | 15.9 ms | **+6.3 ms** |
| p95 | 16.1 ms | 199.9 ms | +184 ms (BatchSpanProcessor spikes) |

Full report: `packages/wallet-service/docs/otel-cold-start-benchmark.md`

## Configuration

All tracing is opt-in per environment:

| Env var | Purpose |
|---------|---------|
| `OTEL_EXPORTER_OTLP_ENDPOINT` | Set to collector URL to enable trace export |
| `OTEL_SDK_DISABLED=true` | Kill switch — skips all OTel initialization |
| `OTEL_CONSOLE_DEBUG=true` | Print spans to stdout/CloudWatch for debugging |
| `OTEL_SERVICE_NAME` | Override service name (defaults provided) |

## Test plan
- [x] Daemon: TypeScript compiles, ESLint passes
- [x] Wallet-service: `serverless package` succeeds under 250MB limit (243MB)
- [x] Lambda invoke returns 200 with OTel enabled (no runtime errors)
- [x] Spans visible in CloudWatch with `OTEL_CONSOLE_DEBUG=true`
- [x] MySQL auto-instrumentation captures `db.statement` attributes
- [x] Cold start benchmark validates overhead within RFC estimates (200-800ms)
- [x] Existing tests pass (CI green)


packages/daemon/src/tracing.ts
```diff
@@ -0,0 +1,40 @@
...
+  spanProcessor: new BatchSpanProcessor(
+    new OTLPTraceExporter({
+      url: process.env.OTEL_EXPORTER_OTLP_ENDPOINT,
+    }),
+    {
+      maxQueueSize: 2048,
+      maxExportBatchSize: 512,
+      scheduledDelayMillis: 5000,
+    },
+  ),
```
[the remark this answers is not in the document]
andreabadesso: Fixed in 613413c — span processor is now only created when `OTEL_EXPORTER_OTLP_ENDPOINT` is set.

packages/daemon/src/tracing.ts
```diff
@@ -0,0 +1,40 @@
...
+    getNodeAutoInstrumentations({
+      '@opentelemetry/instrumentation-fs': { enabled: false },
+      '@opentelemetry/instrumentation-dns': { enabled: false },
+    }),
+  ],
+});
+
+sdk.start();
+
+process.on('SIGTERM', () => sdk.shutdown());
```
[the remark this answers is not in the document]
andreabadesso: Fixed in 613413c — `sdk.shutdown()` is now awaited with try/catch.

packages/daemon/src/tracing.ts
```diff
@@ -0,0 +1,50 @@
...
+
+sdk.start();
+
+process.on('SIGTERM', async () => {
+  try {
+    await sdk.shutdown();
+  } catch (err) {
+    console.error('OTel SDK shutdown error:', err);
+  }
+});
```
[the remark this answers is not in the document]
andreabadesso: Fixed in 5109216 — using `process.once` and calling `process.exit()` after shutdown.

andreabadesso:
All CodeRabbit feedback addressed in 5109216:
- Added `exportTimeoutMillis: 5000` to `BatchSpanProcessor` for predictable shutdown
- Changed `process.on` to `process.once` and added `process.exit()` after shutdown
- Fixed span leak on early returns in `handleVertexAccepted`

packages/daemon/package.json
```diff
@@ -55,6 +60,12 @@
   "dependencies": {
     "@aws-sdk/client-lambda": "3.540.0",
     "@aws-sdk/client-sqs": "3.540.0",
+    "@opentelemetry/api": "^1.9.0",
+    "@opentelemetry/auto-instrumentations-node": "^0.56.0",
+    "@opentelemetry/exporter-trace-otlp-http": "^0.57.0",
+    "@opentelemetry/resources": "^1.30.0",
+    "@opentelemetry/sdk-node": "^0.57.0",
+    "@opentelemetry/sdk-trace-node": "^1.30.0",
```
andreabadesso: Please use fixed versions

packages/wallet-service/package.json
```diff
@@ -21,6 +21,17 @@
...
+    "@opentelemetry/exporter-trace-otlp-http": "^0.57.0",
+    "@opentelemetry/instrumentation-aws-lambda": "^0.50.0",
+    "@opentelemetry/instrumentation-aws-sdk": "^0.48.0",
+    "@opentelemetry/instrumentation-http": "^0.57.0",
+    "@opentelemetry/instrumentation-mysql": "^0.43.0",
+    "@opentelemetry/instrumentation-redis": "^0.44.0",
+    "@opentelemetry/instrumentation-winston": "^0.43.0",
+    "@opentelemetry/resources": "^1.30.0",
+    "@opentelemetry/sdk-node": "^0.57.0",
+    "@opentelemetry/sdk-trace-node": "^1.30.0",
```
andreabadesso: Same here, use fixed versions

packages/wallet-service/src/tracing.ts
```diff
@@ -0,0 +1,65 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+// Skip all OTel initialization when disabled — avoids module loading cost.
+// eslint-disable-next-line no-console
+console.log(`[tracing] OTEL_SDK_DISABLED=${process.env.OTEL_SDK_DISABLED || 'false'}`);
```
andreabadesso: Remove useless log

pedroferreira1 (dismissed):

pedroferreira1 (dismissed):

packages/wallet-service/src/tracing.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  const sdk = new NodeSDK({
+    resource: new Resource({
+      'service.name': process.env.OTEL_SERVICE_NAME || 'wallet-service-lambda',
+      'deployment.environment': process.env.STAGE || 'local',
+    }),
+    ...(spanProcessors.length && { spanProcessors }),
+    instrumentations: [
+      new AwsLambdaInstrumentation({
+        disableAwsContextPropagation: true,
```
luislhl: Why are you disabling this?

Won't this make it harder to trace situations where Lambdas call each other, or the Daemon calls Lambdas?

From what I could understand, this means the Lambda being called will ignore any trace information passed by the caller.

Usually in distributed tracing systems the chain of invocations between the microservices (which in this case are the daemon and the Lambdas) is important to debug the features involving multiple of them.

But I don't know if we will be able to trace them together in other ways even with this disabled.

packages/daemon/src/machines/SyncMachine.ts
```diff
@@ -392,16 +420,16 @@ export const SyncMachine = Machine<Context, any, Event>({
...
-    handleReorgStarted,
-    handleTokenCreated,
+    handleVertexAccepted: tracedService('handleVertexAccepted', handleVertexAccepted),
+    handleVertexRemoved: tracedService('handleVertexRemoved', handleVertexRemoved),
+    handleVoidedTx: tracedService('handleVoidedTx', handleVoidedTx),
+    handleTxFirstBlock: tracedService('handleTxFirstBlock', handleTxFirstBlock),
+    handleNcExecVoided: tracedService('handleNcExecVoided', handleNcExecVoided),
+    handleUnvoidedTx: tracedService('handleUnvoidedTx', handleUnvoidedTx),
+    handleReorgStarted: tracedService('handleReorgStarted', handleReorgStarted),
+    handleTokenCreated: tracedService('handleTokenCreated', handleTokenCreated),
```
luislhl: You're wrapping here all these services to be traced, but inside many of them (maybe all), the code is also wrapped in another `tracer.startActiveSpan` call.

Won't this generate duplicate spans for every service?

packages/daemon/src/services/index.ts
```diff
@@ -91,6 +92,22 @@ import { invokeOnTxPushNotificationRequestedLambda, getDaemonUptime, retryWithBa
 import { addAlert, Severity } from '@wallet-service/common';
 import { JSONBigInt } from '@hathor/wallet-lib/lib/utils/bigint';
 
+const tracer = trace.getTracer('wallet-service-daemon');
+
+async function withSpan<T>(name: string, fn: () => Promise<T>): Promise<T> {
```
luislhl, line 97: This is almost identical to the wrapper in SyncMachine.ts. Do we need both?

packages/daemon/src/tracing.ts
```diff
@@ -0,0 +1,53 @@
...
+    'service.name': process.env.OTEL_SERVICE_NAME || 'wallet-service-daemon',
+    'service.version': process.env.SERVICE_VERSION || 'unknown',
+    'deployment.environment': process.env.STAGE || 'local',
+  }),
+  ...(spanProcessor && { spanProcessor }),
+  instrumentations: [
+    getNodeAutoInstrumentations({
+      '@opentelemetry/instrumentation-fs': { enabled: false },
+      '@opentelemetry/instrumentation-dns': { enabled: false },
+    }),
```
luislhl: It seems this will load a bunch of instrumentations for tools we don't use.

Shouldn't we handpick just the needed ones like we did for the Lambdas? This could impact memory consumption in the daemon (I don't know by how much)

packages/daemon/src/services/index.ts
```diff
@@ -708,14 +751,15 @@ export const voidTx = async (
     if (inputsSpentByThisTx.length > 0) {
       await unspendUtxos(mysql, inputsSpentByThisTx);
     }
```
luislhl, line 753: Would it be better to put this into its own "withSpan"? It seems we will make many sql queries together. Having this one separated could help to debug which one is the slow one.

packages/daemon/src/tracing.ts
```diff
@@ -0,0 +1,53 @@
...
+        exportTimeoutMillis: 5000,
+      },
+    )
+  : undefined;
+
+const sdk = new NodeSDK({
+  resource: new Resource({
+    'service.name': process.env.OTEL_SERVICE_NAME || 'wallet-service-daemon',
+    'service.version': process.env.SERVICE_VERSION || 'unknown',
+    'deployment.environment': process.env.STAGE || 'local',
```
luislhl: We will need to update our manifests in Kubernetes to set this, right? Or maybe find another way to define the version inside the application itself.

In any case, at least a reminder would be good to not forget this.

packages/daemon/src/tracing.ts
```diff
@@ -0,0 +1,53 @@
...
+      '@opentelemetry/instrumentation-fs': { enabled: false },
+      '@opentelemetry/instrumentation-dns': { enabled: false },
+    }),
+  ],
+});
+
+sdk.start();
+
+process.once('SIGTERM', async () => {
+  try {
```
luislhl: Should we handle SIGINT here as well?

packages/wallet-service/src/tracing.ts
```diff
@@ -0,0 +1,63 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+// Skip all OTel initialization when disabled — avoids module loading cost.
+if (process.env.OTEL_SDK_DISABLED !== 'true') {
```
luislhl, line 9: Should have a feature flag like this for the daemon as well?

packages/wallet-service/src/tracing.ts
andreabadesso: Fixed in 76c2607 — removed `disableAwsContextPropagation: true` so Lambdas now pick up parent trace context from callers.

packages/daemon/src/machines/SyncMachine.ts
andreabadesso: Good catch! Fixed in 2f2d8ef — removed the `tracedService` wrapper entirely. Each service function already creates its own span via `startActiveSpan` internally, so the wrapper was producing duplicate spans.

packages/daemon/src/services/index.ts
andreabadesso: Resolved by 2f2d8ef — the `tracedService` wrapper in SyncMachine was removed, so there's no more duplication. The `withSpan` helper in services/index.ts remains as the single place for tracing service internals.

packages/daemon/src/tracing.ts
andreabadesso: Fixed in 70ea8b4 — replaced `getNodeAutoInstrumentations` with only the instrumentations the daemon actually uses: HTTP, MySQL, and Winston.

packages/daemon/src/services/index.ts
andreabadesso: The MySQL instrumentation already creates spans per query automatically, so individual SQL calls should be visible in traces without additional manual wrapping. I think the current granularity is good enough — we already wrap logical units like `voidTransaction`, `markUtxosAsVoided`, etc. with `withSpan`. Let me know if you'd still prefer a separate span here.

packages/daemon/src/tracing.ts
andreabadesso: Fixed in 17fe66f — now reads the version from `package.json` as a fallback, so it works out of the box without K8s manifest changes. The env var `SERVICE_VERSION` can still override it when needed.

packages/daemon/src/tracing.ts
andreabadesso: Fixed in 30f9c63 — added `SIGINT` handler alongside `SIGTERM`.

packages/wallet-service/src/tracing.ts
andreabadesso: Fixed in 3b019e4 — added the same `OTEL_SDK_DISABLED` guard for the daemon, matching the Lambda pattern. Also switched to `require()` so all OTel modules are fully skipped when disabled.

packages/daemon/src/services/index.ts
luislhl: No, that 's ok for me. I'll close.

luislhl (approved):

pedroferreira1 (approved):

PR #386, opened 2026-03-19
chore: Docker image build on CI

tuliomir:
### Motivation

The Wallet Lib has, since https://github.com/HathorNetwork/hathor-wallet-lib/pull/909 , used a Wallet Service image to run its integration tests, but those images are being built manually.

Having them built with each new tag ensures the integration tests are executed on an updated version of the Wallet Service. 

### Acceptance Criteria

- A new Wallet Service image should be published to DockerHub on every release tag
- It should consist of the `sync-daemon`, `ws-migrator` and `lambas` images.

### Notes

This PR proposes two tags: one for the latest `dev` and other for a pinned version like `dev-v1.12.0`.

It will also be necessary to manually verify if the DockerHub credentials are available for this repository before the script can actually work.

### Checklist

## Summary by CodeRabbit

* **Chores**
  * Added automated Docker image builds and publishing triggered by release tags: builds multiple service images and publishes them to Docker Hub with both static and versioned tags.
  * Uses build caching to speed up image creation and ensures reproducible builds for one image.
  * Updated packaging ignore rules for a service to streamline build context.


.github/workflows/main.yml
```diff
@@ -122,3 +122,87 @@ jobs:
...
         export NODE_ENV=test
         nix develop . -c yarn workspace sync-daemon run test_integration
+
+  docker:
+    runs-on: ubuntu-latest
+    needs: [test]
+    if: startsWith(github.ref, 'refs/tags/v')
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
```
luislhl: Maybe we should use v6, which is the newest

.github/workflows/main.yml
```diff
@@ -122,3 +122,87 @@ jobs:
...
+      run: |
+        echo "sequelize=$(jq -er '.devDependencies["sequelize"]' package.json)" >> "$GITHUB_OUTPUT"
+        echo "sequelize_cli=$(jq -er '.devDependencies["sequelize-cli"]' package.json)" >> "$GITHUB_OUTPUT"
+        echo "mysql2=$(jq -er '.devDependencies["mysql2"]' package.json)" >> "$GITHUB_OUTPUT"
+
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v3
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
```
luislhl: There a v4 for these from 2 weeks ago, in case you think we should upgrade

.github/workflows/main.yml
```diff
@@ -122,3 +122,87 @@ jobs:
...
+
+    - name: Login to DockerHub
+      uses: docker/login-action@v3
```
luislhl: There is v4 for this one too

.github/workflows/main.yml
tuliomir: Thanks, I took the opportunity to upgrade all action versions on https://github.com/HathorNetwork/hathor-wallet-service/pull/386/commits/82f57d0cc4e59a18db34865686351e3a3cc59aeb . The functionality most probably didn't change very much, but the NodeJS support is really important to make those workflows stable.

.github/workflows/main.yml
tuliomir: I agree we should upgrade. Done on https://github.com/HathorNetwork/hathor-wallet-service/pull/386/commits/82f57d0cc4e59a18db34865686351e3a3cc59aeb

.github/workflows/main.yml
tuliomir: This one too, upgraded on https://github.com/HathorNetwork/hathor-wallet-service/pull/386/commits/82f57d0cc4e59a18db34865686351e3a3cc59aeb

.github/workflows/main.yml
```diff
@@ -122,3 +122,87 @@ jobs:
...
+
+    - name: Extract dependency versions for migrator
+      id: deps
+      run: |
+        echo "sequelize=$(jq -er '.devDependencies["sequelize"]' package.json)" >> "$GITHUB_OUTPUT"
+        echo "sequelize_cli=$(jq -er '.devDependencies["sequelize-cli"]' package.json)" >> "$GITHUB_OUTPUT"
+        echo "mysql2=$(jq -er '.devDependencies["mysql2"]' package.json)" >> "$GITHUB_OUTPUT"
+
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v3
```
luislhl: Why do we need this? It seems you're building only for amd64.

Usually we need it when building cross-platform.

.github/workflows/main.yml
```diff
@@ -122,3 +122,87 @@ jobs:
...
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+    - name: Copy daemon .dockerignore to root
+      run: cp packages/daemon/.dockerignore .dockerignore
+
+    - name: Build and push daemon dev image
+      uses: docker/build-push-action@v7
+      with:
+        context: .
```
luislhl, line 161: Should the context really be `.` here? Or should it be restricted to `packages/daemon`?

I'm just asking because I spotted this asymmetry, I'm not really sure and didn't test to check.

tuliomir: Yes, this is the explanation on why:

`─────────────────────────────────────────────────`
**Why `context: .` is required for a Yarn workspaces monorepo:**

Line 41 does `COPY . .` — it copies the entire build context into the container. Then line 48 runs `yarn install` and line 50 runs `yarn workspace sync-daemon run build`. Yarn workspaces resolves dependencies across the monorepo: the daemon depends on `@wallet-service/common` (a sibling package). If the context were `packages/daemon/`, Yarn wouldn't find the root `package.json`, the workspace config, or the `packages/common/` directory — `yarn install` would fail.

The `.dockerignore` (copied from `packages/daemon/.dockerignore`) is what keeps the context lean — it excludes `packages/wallet-service/`, `node_modules/`, `.git/`, etc. So the context is the repo root, but filtered down to just what the daemon build needs.

The same applies to the lambdas image — its Dockerfile also does `COPY . .` and `yarn install` from the root. The migrator is the only one that doesn't need this, since it just `COPY ./db ./db` and installs its own minimal deps.
`─────────────────────────────────────────────────`

So the `context: .` is correct for all three images, and the asymmetry you spotted is actually consistent — each Dockerfile was written to expect the repo root as its build context.

.github/workflows/main.yml
```diff
@@ -122,3 +122,87 @@ jobs:
...
+        cache-to: type=gha,mode=max
+
+    - name: Remove stale .dockerignore before migrator build
+      run: rm -f .dockerignore
+
+    - name: Build and push migrator dev image
+      uses: docker/build-push-action@d08e5c354a6adb9ed34480a06d141179aa583294 # v7.0.0
+      with:
+        context: .
+        file: db/Dockerfile.dev
```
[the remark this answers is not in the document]
tuliomir: Created `db/.dockerignore` (excludes everything except `db/`) and updated the step to `cp db/.dockerignore .dockerignore` before the migrator build — same copy-to-root pattern used by the daemon and lambdas builds above.

Note: The `!db/` pattern may look wrong inside `db/.dockerignore`. This is expected — the file is copied to the repo root before the build (`context: .`), so Docker evaluates the patterns from the root where `!db/` is correct.

Do you agree?

.github/workflows/main.yml
tuliomir: Thanks for the heads up. Removed Qemu on https://github.com/HathorNetwork/hathor-wallet-service/pull/386/commits/36258dd82bbf7be572b33261a3238c1a88b53052

.github/workflows/main.yml
luislhl: Great :+1:

luislhl (approved):

andreabadesso (approved):

PR #389, opened 2026-04-09
fix: correct package.json path in daemon tracing

andreabadesso:
### Motivation

The daemon crashed on startup in the `dev` environment with:

```
Error: Cannot find module '../../package.json'
```

This was triggered by the OTel tracing module added in #383. At `packages/daemon/src/tracing.ts:38`, `require('../../package.json')` resolves to `packages/package.json` (nonexistent) from both the source and the compiled `dist/tracing.js`. The correct relative path is `../package.json`, which resolves to `packages/daemon/package.json`.

The fallback path always runs because `SERVICE_VERSION` is not set in any environment. Dev fired first because it was deployed with `OTEL_EXPORTER_OTLP_ENDPOINT` set (OTel enabled); prod would hit the same crash as soon as #383 is rolled out.

The sibling `packages/wallet-service/src/tracing.ts` does not set `service.version` at all, so it is not affected.

### Acceptance Criteria

- [x] `packages/daemon/src/tracing.ts` uses `require('../package.json')` so `service.version` resolves correctly at runtime
- [ ] Daemon starts successfully in the `dev` environment with OTel enabled (no `Cannot find module` error)
- [ ] `service.version` reported in OTel spans matches the version in `packages/daemon/package.json`

### Checklist

Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Telemetry now reports the correct service version and will fall back to "unknown" if a version cannot be determined. Existing telemetry enable/disable behavior and startup/shutdown remain unchanged.

### DIFF
--- a/packages/daemon/src/tracing.ts
+++ b/packages/daemon/src/tracing.ts
@@ -35,7 +35,9 @@ if (process.env.OTEL_SDK_DISABLED !== 'true') {
     resource: new Resource({
       'service.name': process.env.OTEL_SERVICE_NAME || 'wallet-service-daemon',
       // eslint-disable-next-line @typescript-eslint/no-var-requires
-      'service.version': process.env.SERVICE_VERSION || require('../package.json').version,
+      // The daemon package.json has no "version" field — the monorepo root
+      // package.json is the single source of truth (kept in sync with git tags).
+      'service.version': process.env.SERVICE_VERSION || require('../../../package.json').version || 'unknown',
       'deployment.environment': process.env.STAGE || 'local',
     }),
     ...(spanProcessor && { spanProcessor }),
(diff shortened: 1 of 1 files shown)


andreabadesso:
Good catch @coderabbitai — you're right that `packages/daemon/package.json` has no `"version"` field, so the original fix would have left `service.version` as `undefined` in every span.

Pushed 8102f35 to address this. Rather than adding a `"version"` field to the daemon's `package.json`, I pointed the require at the **monorepo root** `package.json` (`../../../package.json`), which is the single source of truth for the version in this repo (currently `1.12.0`, kept in sync with git tags via the release branch flow). Adding a version to the daemon package would duplicate that and drift from the release process, which only bumps the root.

Also added `|| 'unknown'` as a safe fallback per your suggestion.

Verified empirically against the actual Docker build:

1. `docker build --target prod -f packages/daemon/Dockerfile .` succeeds.
2. `/app/package.json` (monorepo root) is present in the final image — not excluded by `packages/daemon/.dockerignore`.
3. From inside the image, a probe script at `/app/packages/daemon/dist/` resolving `require('../../../package.json')` returns `name: hathor-wallet-service, version: 1.12.0`.

pedroferreira1 (approved):

luislhl (approved):

PR #390, opened 2026-04-09
fix: do not fail daemon shutdown on OTel span flush errors

andreabadesso:
### Motivation

When the OTLP endpoint is unreachable at shutdown time (e.g. Jaeger outage, network blip, pod eviction during a collector restart), `sdk.shutdown()` throws an `AggregateError [ECONNREFUSED]` from the exporter's final flush. The handler added in #383 logged it at `error` level and called `process.exit(1)`, which has two bad consequences:

1. The stderr log line containing the word `error` trips the existing `errors-in-logs` alert pattern, paging on-call for a transient collector outage.
2. Exiting with a non-zero code marks the pod termination as unhealthy in k8s.

Losing a handful of buffered telemetry spans during shutdown is not a service failure. This was surfaced in `dev` during an OTLP endpoint session where the collector briefly went away; the same class of alert will fire against `prod` the first time Jaeger has any hiccup.

### Acceptance Criteria

- [x] `sdk.shutdown()` failures during SIGTERM/SIGINT no longer cause the daemon to exit with code 1
- [x] The shutdown flush failure is still observable in logs, but logged at `warn` level with a message that does not contain the word `error` (so the `errors-in-logs` alert pattern is not triggered)
- [ ] A daemon pod with `OTEL_EXPORTER_OTLP_ENDPOINT` pointed at an unreachable host terminates cleanly on SIGTERM (exit 0, no error-level log)

### Checklist

Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Telemetry/exporter flush failures during shutdown are now treated as non-fatal: the shutdown sequence attempts to flush, logs a single warning if flush fails, and the process exits cleanly with a zero exit code to ensure graceful termination.

### DIFF
--- a/packages/daemon/src/tracing.ts
+++ b/packages/daemon/src/tracing.ts
@@ -55,10 +55,15 @@ if (process.env.OTEL_SDK_DISABLED !== 'true') {
     // Swallow exporter flush failures (e.g. OTLP endpoint unreachable) so
     // they don't cause a non-zero exit and don't trip the errors-in-logs
     // alert pattern.
+    //
+    // Intentionally do NOT log the raw error object: its stack trace
+    // commonly contains strings like "AggregateError" / "Error:" which
+    // would be picked up by the log-based alert regex, defeating the
+    // purpose of this handler.
     try {
       await sdk.shutdown();
-    } catch (err) {
-      console.warn('OTel SDK: failed to flush spans during shutdown (non-fatal):', err);
+    } catch {
+      console.warn('OTel flush skipped during shutdown (non-fatal)');
     }
     process.exit(0);
   };
--- a/packages/daemon/src/tracing.ts
+++ b/packages/daemon/src/tracing.ts
@@ -51,13 +51,16 @@ if (process.env.OTEL_SDK_DISABLED !== 'true') {
   sdk.start();
 
   const shutdown = async () => {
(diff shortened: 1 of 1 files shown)


andreabadesso:
Good catch @coderabbitai — this is the exact failure mode that fired the original alert: the log line containing the stringified error object (`AggregateError [ECONNREFUSED]: ...`) is what the log-based alert regex matched on in the first place. Logging the raw `err` via `console.warn` would defeat the whole purpose of this fix.

Pushed d9102a9:

- Dropped the catch binding entirely so there is no raw error object to accidentally format.
- Log a fixed-text, bland message: `OTel flush skipped during shutdown (non-fatal)` — no stack, no error class name, no reason string.
- Added a comment explaining why we intentionally do not include `err.message` / `err.name` either (AggregateError's `name` literally contains "Error", which would also match the regex).

Trade-off: we lose diagnostic detail at shutdown time. If an operator ever needs to debug *why* a shutdown flush is failing, the right path is to check the OTLP endpoint health rather than grep the daemon logs — telemetry-loss-on-shutdown is expected behaviour when the collector is down.

pedroferreira1 (approved):

luislhl (approved):

Issue #395, opened 2026-04-17
perf(daemon): per-event sync optimizations identified from Tempo traces


andreabadesso:
## Context

While investigating steady-state sync performance via Grafana Tempo traces (env `testnet-india`), we found that per-event processing is healthy (~25ms for `handleVertexAccepted`, all sub-DB calls <5ms) but there are several micro-inefficiencies that compound significantly during catch-up.

At 10M events during a full sync, even 10ms savings per event = ~28 hours of wall time. The findings below are ordered by expected impact.

**Trace data reference:** in a ~29 min `testnet-india` window we observed ~152 events processed (71 `handleVertexAccepted`, 76 `metadataDiff`, 4 `handleTxFirstBlock`, 1 `handleTokenCreated`, 0 voids/reorgs). Per-event handler work was ~25ms, sub-spans summed to ~8ms, leaving ~17ms of un-instrumented overhead per event (likely: connection acquire, transaction wrapping, un-wrapped DB calls).

---

### 1. `mysql.destroy()` kills pooled connections — should be `connection.release()`

**File:** `packages/daemon/src/services/index.ts` (11 occurrences across all handlers)

`getDbConnection()` in `packages/daemon/src/db/index.ts:52` returns a pool connection via `pool.getConnection()`. Every handler then calls `mysql.destroy()` at the end, which closes the underlying TCP socket permanently. This defeats the pool — each handler forces a TCP + MySQL auth handshake (plus TLS if enabled) on the next use. Effectively the pool size is 0.

**Fix:** Replace all 11 `mysql.destroy()` calls with `mysql.release()`.

**Estimated savings:** ~10–50ms per event (depends on RDS latency and TLS).

### 2. Use prepared statements (`mysql.execute`) instead of `mysql.query`

**File:** `packages/daemon/src/db/index.ts` — 68 `mysql.query` calls, 0 `mysql.execute` calls.

`query()` re-parses and re-plans SQL on every call. `execute()` (mysql2 prepared statements) caches the plan per connection. With ~10–15 queries per event, this compounds.

**Fix:** Migrate to `mysql.execute()` where parameters are passed. Note: statements using `VALUES ?` with batched arrays may need rework — mysql2 prepared statements handle array expansion differently.

**Estimated savings:** ~1–2ms per query × 10–15 queries = 10–30ms per event.

### 3. Batch N+1 in `updateAddressTablesWithTx`

**File:** `packages/daemon/src/db/index.ts:687-753`

[...]

PR #400, opened 2026-04-17
perf(daemon): use prepared statements for simple scalar queries

### LINKED ISSUE #2: Stop using /v1a/transaction?type=block&count=1 to query the best block
"I don't really like using this API to get the network height but we still don't have an endpoint for that. Maybe we should add one, it's a simple and important API.

In the meantime maybe we could use `/v1a/get_block_template`, this returns the template of the block to be mined.

`response.metadata.height` is the block height to be mined, so the network height is this value -1. I still don't like it, so not sure if we should change.

One of the reasons I don't like using this API is because it was developed to be used on the explorer. With the new explorer service, this API might become deprecated."

_Originally posted by @pedroferreira1 in https://github.com/HathorNetwork/hathor-wallet-service-sync_daemon/pull/1#discussion_r612602567_

andreabadesso:
### Motivation

Tier 1 fix #2 from #395. mysql2's `execute()` caches the query plan per connection; `query()` re-parses on every call. At ~10–15 queries per event on mainnet catch-up, per-query plan-cache reuse adds up.

### Acceptance Criteria

- All `mysql.query()` sites with only scalar `?` placeholders migrated to `mysql.execute()` in `packages/daemon/src/db/index.ts`
- Sites using `VALUES ?`, `SET ?`, or `IN (?)` stay on `query()` (prepared statements don't support those expansion forms)
- Integration and unit tests pass
- No crashes, no behavioral regressions in reachable integration scenarios

### Scope

- **42 of 68** `mysql.query()` call sites converted
- **26 remain on `query()`** because mysql2's prepared-statement protocol does not support:
  - `VALUES ?` with nested arrays (multi-row INSERT)
  - `SET ?` with an object (column expansion)
  - `IN (?)` with an array placeholder (list expansion)
- Those 26 would need to be rewritten with explicit placeholder expansion (e.g. `IN (?, ?, ?)` generated at runtime) before they can move — tracked as a Tier 1.5 follow-up

### Checklist
- [x] Make sure either the unit tests and/or the QA tests are capable of testing the new features (all 214 unit tests pass; integration tests cover the converted paths)
- [x] Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. (no new deps)

### DIFF
--- a/packages/daemon/__tests__/db/index.test.ts
+++ b/packages/daemon/__tests__/db/index.test.ts
@@ -105,7 +105,9 @@ describe('transaction methods', () => {
     await addOrUpdateTx(mysql, 'txId1', null, 1, 1, 65.4321);
     const tx = await getTransactionById(mysql, 'txId1');
 
-    expect(tx?.weight).toStrictEqual(65.4321);
+    // `weight` is stored as FLOAT (32-bit); prepared statements return the
+    // true binary value rather than the text-protocol's rounded display.
+    expect(tx?.weight).toBeCloseTo(65.4321, 4);
   });
 
   test('db which is not on our database should return null', async () => {
--- a/packages/daemon/src/db/index.ts
+++ b/packages/daemon/src/db/index.ts
@@ -213,7 +213,7 @@ export const getTxOutputsFromTx = async (
   mysql: any,
   txId: string,
 ): Promise<DbTxOutput[]> => {
-  const [results] = await mysql.query(
+  const [results] = await mysql.execute(
     `SELECT *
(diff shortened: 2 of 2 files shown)


pedroferreira1 (approved):

luislhl:
> tracked as a Tier 1.5 follow-up

Where are you going to leave this improvement registered?

luislhl (approved):

andreabadesso:
@luislhl tracked in #403 — lists the 26 remaining sites and the acceptance criteria for the Tier 1.5 migration.

PR #428, opened 2026-05-11
test(wallet-service): harden /version contract

tuliomir:
### Motivation

Completes the follow-up work after HathorNetwork/hathor-wallet-service#420, which was a one-line emergency bugfix (allowing `native_token.version` on the fullnode `/version` payload). The fix shipped narrow on purpose; both CodeRabbit and Copilot flagged that it left loose ends — TypeScript typing and tests. This PR closes those out, scoped tightly to `native_token.version`.

### Changes

- **Type alignment** — `FullNodeApiVersionResponse.native_token` gains `version?: number` so the TS type matches the Joi schema. The `as FullNodeApiVersionResponse` cast in `fullnode.ts` is dropped (it was masking the drift; `Joi.object<T>()` already types the validator's output).
- **Schema unit tests** (new `tests/schemas.test.ts`, 4 cases) — first direct coverage of `FullnodeVersionSchema`, all scoped to the `native_token.version` field:
  - **PR 420 regression guard:** the modern testnet payload (with `native_token.version: 0`) validates.
  - Legacy payload (no `native_token.version`) validates.
  - Unknown fields under `native_token` fail validation (design intent — keep nested object strict).
  - Non-integer `native_token.version` (e.g. `1.5`) fails validation.
- **`fullnode.version()` tests** (extend `tests/fullnode.test.ts`) — happy-path and schema-rejection cases that verify the schema wiring inside `fullnode.ts` actually runs and throws. Rejection case uses a `native_token.version` violation.

### Out of scope (intentionally)

A live-fullnode contract test (that would fail CI when the `/version` response drifts) was considered and removed from this PR, since there is no existing precedent for tests making real HTTP calls in the wallet-service workspace — PR HathorNetwork/hathor-wallet-service#394 went the other direction. That can be a separate PR if we want to introduce that pattern.

### Acceptance Criteria

- `FullnodeVersionSchema` accepts the post-PR 420 `native_token` payload and rejects misuse of the new `version` field; covered by unit tests.
- `FullNodeApiVersionResponse` reflects the schema, removing the type cast that was hiding the drift.

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Native token responses now optionally include a version property for richer metadata.

* **Refactor**
  * Version handling updated to return the validated result directly for clearer, safer responses.

* **Tests**
  * Added tests for modern and legacy native-token payloads, strict schema validation, and rejection on invalid version values.

[Review Change Stack](https://app.coderabbit.ai/change-stack/HathorNetwork/hathor-wallet-service/pull/428?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/wallet-service/src/types.ts
+++ b/packages/wallet-service/src/types.ts
@@ -126,7 +126,7 @@ export interface FullNodeApiVersionResponse {
   genesis_block_hash?: string,
   genesis_tx1_hash?: string,
   genesis_tx2_hash?: string,
-  native_token?: { name: string, symbol: string };
+  native_token?: { name: string, symbol: string, version?: number };
 }
 
 export interface TxProposal {
--- a/packages/wallet-service/tests/schemas.test.ts
+++ b/packages/wallet-service/tests/schemas.test.ts
@@ -1,10 +1,8 @@
 import { FullnodeVersionSchema } from '@src/schemas';
 import { defaultTestVersionData } from '@tests/utils';
 
-// Regression guard for the testnet outage that motivated PR 420: the
-// fullnode added `native_token.version` and the lambda crashed because the
-// nested object did not allow it. Locking the modern payload in as a
-// passing case prevents that fix from being reverted by accident.
+// Regression guard: the `native_token.version` is not available in all live fullnode instances.
+// The Wallet Service must be able to handle both cases.
 test('FullnodeVersionSchema regression: native_token.version is accepted', () => {
   const payload = {
     ...defaultTestVersionData(),
--- a/packages/wallet-service/tests/schemas.test.ts
+++ b/packages/wallet-service/tests/schemas.test.ts
@@ -1,43 +1,23 @@
 import { FullnodeVersionSchema } from '@src/schemas';
-import { FullNodeApiVersionResponse } from '@src/types';
-
-const VALID_PAYLOAD: FullNodeApiVersionResponse = {
-  version: '0.70.0-rc.1',
-  network: 'testnet-india',
-  nano_contracts_enabled: true,
-  min_weight: 8,
-  min_tx_weight: 8,
-  min_tx_weight_coefficient: 0,
-  min_tx_weight_k: 0,
-  token_deposit_percentage: 0.01,
-  reward_spend_min_blocks: 300,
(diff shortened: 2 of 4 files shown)


packages/wallet-service/tests/schemas.test.ts
```diff
@@ -0,0 +1,45 @@
+import { FullnodeVersionSchema } from '@src/schemas';
+import { defaultTestVersionData } from '@tests/utils';
+
+// Regression guard for the testnet outage that motivated PR 420: the
```
r4mmer: We should not reference the PR solving the outage if possible, the comment should document the core issue this addresses instead.

I think a link to an issue describing the problem would be good if the core issue was too complex to describe in the comment.

r4mmer (dismissed):

packages/wallet-service/src/types.ts
```diff
@@ -126,7 +126,7 @@ export interface FullNodeApiVersionResponse {
   genesis_block_hash?: string,
   genesis_tx1_hash?: string,
   genesis_tx2_hash?: string,
-  native_token?: { name: string, symbol: string };
+  native_token?: { name: string, symbol: string, version?: number };
```
raul-oliveira, line 129: I believe we should change the validation schema to allow more fields than the required ones.

So if the fullnode sent the right fields with a couple of more ones like "abc": "asdasd", and "anotherTokenProp": "random" 

It shoudn't break the lambda, only if one of the required is missing.

packages/wallet-service/tests/schemas.test.ts
tuliomir: Good point. Reworded on https://github.com/HathorNetwork/hathor-wallet-service/pull/428/commits/7e61b390d9c405b545b47f8c975b03c8569826cc

packages/wallet-service/src/types.ts
tuliomir: I think this is a good subject for debate, but if I took this decision only in this `native_token` field it would leave the rest of the validations following a different, stricter pattern.

This should be done in another PR in my opinion.

raul-oliveira (approved):

r4mmer (approved):

PR #442, opened 2026-05-28
feat(db/migrations): migrations (2/6)

r4mmer:
## Motivation

Add the database schema for shielded outputs. This layer is additive-only — no existing column is renamed or dropped, and every new column carries a safe default — so master-era code keeps running against the migrated schema unchanged. No code reads the new columns yet.

## Summary

Nine additive Sequelize migrations:

- `…100000` alter `tx_output` — add shielded mode columns
- `…100001` create `shielded_tx_output_data`
- `…100002` alter `address` — `bip32_account` nullable, `ct_address`, `idx_address_ct_long`, backfill
- `…100003` alter `address_balance` — shielded balance columns
- `…100004` alter `address_tx_history` — shielded delta
- `…100005` alter `wallet_balance` — shielded balance columns
- `…100006` alter `wallet_tx_history` — shielded delta
- `…100007` alter `wallet` — add `ct_status ENUM('none','creating','ready','error')`
- `…100008` alter `token` — add `total_supply`

## Acceptance Criteria

- [ ] All 9 migrations apply cleanly on a fresh DB (`sequelize-cli db:migrate`)
- [ ] No existing column renamed or dropped; all new columns have defaults
- [ ] `bip32_account` is nullable (MySQL NULL-distinct unique index lets observation-only rows coexist)
- [ ] wallet-service suite green against the migrated schema (schema-touching suites pass; the pre-existing env-failing suites are unaffected)
- [ ] daemon suite green (master code never writes the new columns → defaults apply)

## Stack

1. #439 — foundations
2. **#442 — migrations**
3. #443 — event schema
4. #444 — DB helpers
5. #446 — ingestion
6. #447 — wallet-service read-path

Built clean from the net state of #434 (the reference branch), which stays open as the reference until this stack lands.

PR #439, opened 2026-05-28
feat(shielded-outputs): foundations (1/6)

r4mmer:
## Motivation

First layer of the shielded-outputs stack. PR1 lands the pure additive foundations — shared balance primitives and a typed crypto-rewind wrapper — with no schema, no ingestion, and no native dependency, so every later layer builds on a stable base.

## Summary

**`@wallet-service/common`**
- `src/types.ts` — `Balance` extended with `unlockedShieldedAmount`, `lockedShieldedAmount`, `totalShieldedReceived`, all defaulted to `0n` so existing callers still compile.
- `src/shielded.ts` — shielded-output primitives/helpers; exported via `index.ts`.
- Unit tests for both (`__tests__/shielded.test.ts`, `__tests__/types.test.ts`).

**`packages/daemon/src/crypto/ctRewind.ts`** — the typed rewind wrapper surface: `RewindError`, the `AmountRewind`/`FullyRewind` arg+result interfaces, and `rewindAmount`/`rewindFully`. Bodies are **stubs that throw `RewindError`** — no `@hathor/ct-crypto-node` import. The real native binding and its round-trip test are deferred to a later integration PR.

**`__tests__/crypto/ctRewind.test.ts`** — asserts both functions throw `RewindError`.

**`.github/workflows/main.yml`** — daemon tests run with `NETWORK=mainnet`.

## Acceptance Criteria

- [ ] `@wallet-service/common` builds; `Balance` has the 3 new shielded fields, all defaulted to `0n`
- [ ] Existing daemon + wallet-service callers that construct `Balance` still compile
- [ ] `ctRewind.ts` exports `RewindError`, the arg/result interfaces, and `rewindAmount`/`rewindFully`
- [ ] `rewindAmount`/`rewindFully` throw `RewindError` (stub) — no `@hathor/ct-crypto-node` import anywhere
- [ ] `ctRewind.test.ts` asserts both throw `RewindError`
- [ ] common unit tests green (shielded + types)

## Stack

1. **#439 — foundations**
2. #442 — migrations
3. #443 — event schema
4. #444 — DB helpers
5. #446 — ingestion
6. #447 — wallet-service read-path

Built clean from the net state of #434 (the reference branch), which stays open as the reference until this stack lands.

## Summary by CodeRabbit

* **New Features**
  * Added shielded transaction support across daemon and wallet, including new shielded balance fields and extended shielded output metadata.
  * Realtime payloads now include lightweight `shielded_outputs` and derived `addresses` for involved-sets.
[...]

PR #443, opened 2026-05-28
feat(daemon): Daemon event schema (3/6)

r4mmer:
## Motivation

Teach the daemon to parse shielded outputs from the fullnode wire format. This layer only extends the Zod event schemas; `handleVertexAccepted` still doesn't read `shielded_outputs[]` — that wiring is PR5.

## Summary

**`packages/daemon/src/types/event.ts`**
- `AmountShieldedOutputSchema` (mode 1, has `token_data`)
- `FullyShieldedOutputSchema` (mode 2, has `asset_commitment` + `surjection_proof`)
- `ShieldedOutputSchema` — discriminated union of the two
- `SpentOutputSchema` — discriminated union `mode: 0 | 1 | 2` with a preprocessor that defaults missing `mode` to 0 for transparent wire shapes
- `EventTxInputSchema.spent_output` widened to `SpentOutput`
- `TxEventDataWithoutMetaSchema.shielded_outputs` added as `ShieldedOutput[].default([])`

**Other files**
- `__tests__/__fixtures__/events.ts` — shielded vertex fixtures
- Two new test files: `event-shielded.test.ts`, `vertex-with-shielded-fixture.test.ts`
- `bench-void-tx.ts`, `__tests__/utils.ts` — `mode: 0` added to transparent output literals required by the union
- `services/index.ts`, `utils/wallet.ts` — `@ts-ignore` at call sites that read transparent-only fields on the now-wider `SpentOutput` union; removed when those files are rewritten in PR5

## Acceptance Criteria

- [ ] `tsc --noEmit` passes clean in packages/daemon
- [ ] `event.ts` exports the new shielded/spent schemas; `spent_output` is `SpentOutput`; `shielded_outputs` defaults to `[]`
- [ ] preprocessor defaults missing `mode` to 0 for transparent wire shapes
- [ ] new tests pass: `event-shielded.test.ts`, `vertex-with-shielded-fixture.test.ts`
- [ ] full daemon suite green (13/13 suites, 236 tests at this layer)
- [ ] ingestion unchanged — `handleVertexAccepted` does not yet read `shielded_outputs[]`

## Stack

1. #439 — foundations
2. #442 — migrations
3. **#443 — event schema**
4. #444 — DB helpers
5. #446 — ingestion

### DIFF
--- a/packages/daemon/__tests__/types/event-shielded.test.ts
+++ b/packages/daemon/__tests__/types/event-shielded.test.ts
@@ -1,0 +1,170 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import {
+  ShieldedOutputSchema,
+  SpentOutputSchema,
+  TxEventDataWithoutMetaSchema,
+} from '../../src/types/event';
+
+describe('shielded event schemas', () => {
+  describe('ShieldedOutputSchema', () => {
+    it('accepts a mode=1 AmountShielded entry with token_data', () => {
+      const v = ShieldedOutputSchema.parse({
+        mode: 1,
+        commitment: 'aa'.repeat(33),
+        range_proof: 'bb'.repeat(64),
+        script: 'cc'.repeat(20),
+        ephemeral_pubkey: 'dd'.repeat(33),
+        token_data: 1,
+        decoded: { address: 'WT4nABC' },
+      });
+      expect(v.mode).toBe(1);
+      if (v.mode === 1) {
+        expect(v.token_data).toBe(1);
+      }
+    });
+
+    it('accepts a mode=2 FullyShielded entry with asset_commitment and surjection_proof', () => {
+      const v = ShieldedOutputSchema.parse({
+        mode: 2,
+        commitment: 'aa'.repeat(33),
+        range_proof: 'bb'.repeat(64),
+        script: 'cc'.repeat(20),
+        ephemeral_pubkey: 'dd'.repeat(33),
+        asset_commitment: 'ee'.repeat(33),
+        surjection_proof: 'ff'.repeat(64),
+        decoded: { address: 'WT4nXYZ' },
+      });
+      expect(v.mode).toBe(2);
+      if (v.mode === 2) {
+        expect(v.asset_commitment).toBe('ee'.repeat(33));
+        expect(v.surjection_proof).toBe('ff'.repeat(64));
+      }
+    });
+
+    it('rejects an unknown mode (mode=9)', () => {
+      expect(() =>
+        ShieldedOutputSchema.parse({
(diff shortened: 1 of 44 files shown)


packages/daemon/src/types/event.ts
```diff
@@ -120,10 +120,67 @@ export const EventTxOutputSchema = z.object({
...
+  // plus `vertex.heightlock` (shielded outputs don't carry an on-wire
+  // `locked` flag).
+  timelock: z.number().int().nullish(),
+}).passthrough();
+
+const BaseShieldedFieldsSchema = z.object({
+  commitment: HexStringSchema,
+  range_proof: HexStringSchema,
+  script: HexStringSchema,
+  ephemeral_pubkey: HexStringSchema,
```
pedroferreira1: `commitment` and `ephemeral_pubkey` have fixed length. Does it make sense to add this validation to the scheme? The same for `asset_commitment` below

pedroferreira1 (approved):

packages/daemon/__tests__/types/event-shielded.test.ts
```diff
@@ -0,0 +1,170 @@
...
+        ephemeral_pubkey: 'dd'.repeat(33),
+        asset_commitment: 'ee'.repeat(33),
+        surjection_proof: 'ff'.repeat(64),
+        decoded: { address: 'WT4nXYZ' },
+      });
+      expect(v.mode).toBe(2);
+      if (v.mode === 2) {
+        expect(v.asset_commitment).toBe('ee'.repeat(33));
+        expect(v.surjection_proof).toBe('ff'.repeat(64));
+      }
```
tuliomir, line 47: suggestion(non-blocking): It's not a good practice to make assertions inside conditionals on `jest`.

We've already asserted `v.mode` is `2`, so we can remove this `if` here.

One more suggestion on the topic: add `jest/no-conditional-expect` as an `eslint` rule to this project, ensuring this is avoided in the future.

tuliomir (approved):

packages/daemon/__tests__/types/event-shielded.test.ts
r4mmer: Tried removing the if, it is there to deduplicate the typing since `ShieldedOutputSchema` is a union of the possible shielded outputs.

packages/daemon/src/types/event.ts
r4mmer: Done

packages/daemon/__tests__/types/event-shielded.test.ts
tuliomir: I'd suggest then using `if (v.mode !== 2) throw new Error('Incorrect mode')`.

The important part here is that if v.mode was actually different from `2` the test would incorrectly pass, since the `if` would prevent the assertions from running.

But like I said, it's good practice but not mandatory.,

pedroferreira1 (approved):

PR #444, opened 2026-05-28
feat(daemon): shielded DB helpers (PR 4/6)

r4mmer:
## Motivation

Shielded outputs require new database operations that don't exist in the current schema: storing encrypted output data, tracking CT address observations, and recording the outcome of amount-rewind attempts. Without these helpers, the ingestion layer (PR5) has no way to persist shielded state.

This PR adds the write/query layer over the columns introduced in PR2, keeping it isolated from ingestion so the DB logic can be reviewed independently.

## Summary

**`packages/daemon/src/db/index.ts`** — 5 new helpers:
- `insertShieldedTxOutputData` — inserts a row into `shielded_tx_output_data` for AmountShielded and FullyShielded outputs
- `upsertShieldedAddressObservation` — records that a CT address appeared in a tx; `bip32_account = NULL` marks observation-only (not yet resolved to a wallet key)
- `findShieldedAddressOwnership` — looks up which wallets own a CT address, returning the `bip32_account` index needed for rewind dispatch
- `markTxOutputRecovered` — sets `recovery_state = 'recovered'` and writes the recovered `value` after a successful rewind
- `markTxOutputRecoveryFailed` — sets `recovery_state = 'recovery_failed'` when rewind is impossible or the output belongs to another wallet

**`packages/daemon/src/types/db.ts`** — `DbTxOutput` extended with `mode: number` and `recoveryState: string | null`

**`packages/daemon/src/types/transaction.ts`** — adds `RecoveryState` union type and `ShieldedOutputMode` enum

**`@ts-ignore` annotations** — `unlockUtxos` in `utils/wallet.ts` maps `DbTxOutput` to `TxOutput[]`; `value` and `tokenId` are now nullable on shielded rows, but the function itself is rewritten in PR5. Three `@ts-ignore`s mark the call sites.

**6 new test suites** covering every new helper (insert, upsert, find, mark-recovered, mark-failed, shielded spent-by update). All 19/19 suites pass (253 tests).

## Acceptance Criteria

- [ ] `packages/daemon/src/db/index.ts` exports all 5 new helpers
- [ ] `DbTxOutput` has `mode: number` and `recoveryState: string | null`
- [ ] `RecoveryState` and `ShieldedOutputMode` are exported from `types/transaction.ts`
- [ ] 6 new DB test suites exist, one per helper group
- [ ] Full daemon suite passes: 19/19 suites, 253/253 tests
- [ ] `tsc` compiles clean (no errors beyond the 3 annotated `@ts-ignore`s in `unlockUtxos`)
- [ ] No calls to `@hathor/ct-crypto-node` anywhere in this PR
- [ ] New helpers are not yet called from `services/index.ts` (wired in PR5)

## Stack

1. #439 — foundations
2. #442 — migrations
3. #443 — event schema
4. **#444 — DB helpers**
5. #446 — ingestion

### DIFF
--- a/packages/daemon/src/db/index.ts
+++ b/packages/daemon/src/db/index.ts
@@ -254,7 +254,7 @@ export interface InsertTxOutputArgs {
   locked: boolean;
   voided: boolean;
   spent_by?: string | null;
-  recovery_state: 'unowned' | 'recovered' | 'recovery_failed' | null;
+  recovery_state: RecoveryState | null;
 }
 
 /**
--- a/packages/daemon/src/db/index.ts
+++ b/packages/daemon/src/db/index.ts
@@ -28,7 +28,7 @@ import {
   TxOutputWithIndex,
   Balance,
 } from '@wallet-service/common';
-import { isAuthority, toTokenVersion } from '@wallet-service/common';
+import { isAuthority, toTokenVersion, RecoveryState, Bip32Account } from '@wallet-service/common';
 import { getWalletBalanceMap } from '../utils/wallet';
 import {
   AddressBalanceRow,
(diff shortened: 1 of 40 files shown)


packages/daemon/src/db/index.ts
```diff
@@ -740,28 +1084,24 @@ export const updateAddressTablesWithTx = async (
...
-   */
-  const addressEntries = Object.keys(addressBalanceMap).map((address) => [address, 1]);
-  if (addressEntries.length > 0) {
-    await mysql.query(
-      `INSERT INTO \`address\`(\`address\`, \`transactions\`)
-            VALUES ?
-                ON DUPLICATE KEY UPDATE transactions = transactions + 1`,
-      [addressEntries],
-    );
-  }
```
r4mmer, line 759: This bump on transactions is removed because later (during "handle vertex accepted") we will create a list of all addresses involved on the tx and bump the transaction count there.

This is the source of the error on the integration test, if we miss the bump here the void will try to remove and will go into negative (which creates an SQL error since the column is `UNSIGNED`).

packages/daemon/src/types/transaction.ts
```diff
@@ -19,6 +21,8 @@ export interface DbTxOutput {
   txProposalId?: string | null;
   txProposalIndex?: number | null;
   voided?: boolean | null;
+  mode: number;
+  recoveryState: RecoveryState | null;
```
tuliomir, line 25: question(non-blocking): I'm thinking about consistency here. The columns `voided`, `txProposalIndex`, `txProposalId`, `spendBy` can accept `undefined` besides their own valid type and `null`.

It doesn't look like there is actual meaning to `undefined` for those fields, so in my opinion they should be simplified by removing the `undefined`. But in the case I'm not aware of some implementation detail: should `recoveryState` also accept `undefined`?

packages/daemon/src/db/index.ts
```diff
@@ -235,13 +235,351 @@ export const addUtxos = async (
...
+  address: string;
+  value: bigint | null;
+  token_id: string | null;
+  authorities: number;
+  timelock: number | null;
+  heightlock: number | null;
+  locked: boolean;
+  voided: boolean;
+  spent_by?: string | null;
+  recovery_state: 'unowned' | 'recovered' | 'recovery_failed' | null;
```
tuliomir: chore: I believe we already have an `enum` for that, don't we?

If would make it easier for callers not to need a hardcoded literal, and instead use a constant as parameter.

packages/daemon/src/db/index.ts
```diff
@@ -235,13 +235,351 @@ export const addUtxos = async (
...
+ * inputs uniformly. The query only consumes the primary-key columns.
+ *
+ * @param mysql - Database connection
+ * @param inputs - The spent input references (tx_id + index per entry)
  * @param txId - The transaction that spent these utxos
  */
-export const updateTxOutputSpentBy = async (mysql: any, inputs: TxInput[], txId: string): Promise<void> => {
+export const updateTxOutputSpentBy = async (
+  mysql: any,
+  inputs: Array<{ tx_id: string; index: number }>,
```
tuliomir, line 649: question(non-blocking): Why stop using the `TxInput` type?

packages/daemon/src/db/index.ts
```diff
@@ -235,13 +235,351 @@ export const addUtxos = async (
...
+ */
+export async function markTxOutputRecovered(
+  conn: any,
+  txId: string,
+  index: number,
+  recovered: { value: bigint; token_id: string },
+): Promise<{ affectedRows: number }> {
+  const [r] = await conn.query(
+    `UPDATE tx_output
+        SET value = ?, token_id = ?, recovery_state = 'recovered'
```
pedroferreira1, line 504: In case it's an amount shielded only, it also sends the token_id?

packages/daemon/src/db/index.ts
```diff
@@ -235,13 +235,351 @@ export const addUtxos = async (
...
+ * column.
+ */
+export async function incrementTokenTotalSupply(
+  conn: any,
+  tokenId: string,
+  delta: bigint,
+): Promise<void> {
+  await conn.execute(
+    'UPDATE token SET total_supply = total_supply + ? WHERE id = ?',
+    [delta.toString(), tokenId],
```
pedroferreira1, line 571: Does it work well the sql sum of bigint values as string?

packages/daemon/src/db/index.ts
```diff
@@ -740,28 +1084,24 @@ export const updateAddressTablesWithTx = async (
...
-                ON DUPLICATE KEY UPDATE transactions = transactions + 1`,
-      [addressEntries],
-    );
-  }
+  // The per-address `address.transactions` bump is handled by the central
+  // `bumpAddressInvolvement` helper earlier in the vertex pipeline, which
+  // is driven by the wire-level involvement set so it covers
+  // shielded-credit-only, shielded-spend-only, and unowned shielded
+  // vertices alongside the transparent ones. This function is now only
+  // responsible for the per-(address, token) row writes.
```
pedroferreira1, line 1167: I saw you created the helper but I don't see it in use yet. Is it used in another PR?

packages/daemon/src/db/index.ts
```diff
@@ -821,18 +1177,30 @@ export const updateAddressTablesWithTx = async (
...
       // care by the first sql query.
 
-      // update address_tx_history with one entry for each pair (address, token)
-      entries.push([address, txId, token, tokenBalance.total(), timestamp]);
+      // update address_tx_history with one entry for each pair (address, token).
+      // `balance` is the transparent delta (signed unlocked + locked) and
+      // `shielded_balance_delta` is the shielded delta (signed
+      // unlockedShielded + lockedShielded); the two columns live on the
+      // same row keyed by (address, tx_id, token_id).
+      const transparentDelta = tokenBalance.unlockedAmount + tokenBalance.lockedAmount;
```
pedroferreira1, line 1260: Is this the same as the previous `tokenBalance.total()`?

packages/daemon/src/db/index.ts
r4mmer: An amount shielded utxo will be saved with value null and token_id known, then if a known wallet owns the address we rewind the value and use this method to set the value+token_id again.

If it was not owned it will be left with mode=1, value=null and token_id=cafecafecafe....

packages/daemon/src/db/index.ts
r4mmer: Yes, when we added support for bigint values we changed the database connection config (./db/config.js) to always use the options: 

```js
{
// ....
    dialect: 'mysql',
    dialectOptions: {
      supportBigNumbers: true,
      bigNumberStrings: true,
    }
}
```

So interacting with a "big number" on the database will always return as strings and allow receiving as string as well.  

The reason behind this is that without these options a query returning a big number would be auto converted to Javascript's `number` and this would cause errors, but returning as strings and manually converting to bigint makes it work no matter how high the number is.

packages/daemon/src/db/index.ts
r4mmer: Yes, the next PR will only have the changes on ingestion logic (which use the helpers in this PR)

packages/daemon/src/db/index.ts
r4mmer:
```ts
  total(): bigint {
    return this.unlockedAmount + this.lockedAmount;
  }
```

Total is a helper that indicates "all balance changes regardless of locked state" but now means "all **transparent** balance changes ...".

Since the total method is used in 2 places on the entire codebase i thought that just using the added values would create less confusion as to what is being added.

packages/daemon/src/db/index.ts
pedroferreira1: Can we use the type you created?

packages/daemon/src/db/index.ts
```diff
@@ -235,13 +235,351 @@ export const addUtxos = async (
...
+ * @param conn - Database connection
+ * @param address - The P2PKH spend address (CTSpend-derived) observed as the
+ *   destination of a shielded `tx_output`
+ */
+export async function upsertShieldedAddressObservation(
+  conn: any,
+  address: string,
+): Promise<void> {
+  await conn.query(
+    `INSERT IGNORE INTO \`address\` (\`address\`, \`transactions\`)
```
pedroferreira1: This is a comment by claude:

[minor] Don't use INSERT IGNORE. It downgrades NOT NULL / range / FK errors to warnings, not justa" semantics. Also the name says upsert* but the behavior is
  duplicate-key. Use INSERT INTO address (address, transactions) VALUES (?, 0) ON DUPLICATE KEY
  UPDATE address = address for the same "row exists, preserve registration data" semantics. Also the  elling the 'unowned' | 'recovered' | 'recovery_failed' literal union
  name says upsert* but the behavior is insert-if-absent — consider ensureShieldedAddressObserved.

pedroferreira1 (approved):

packages/daemon/src/db/index.ts
r4mmer: The name was due to this insert originally adding transaction=1, but the transaction bump was moved to a single place (before it was spread across different places depending on the origin of the address).
Now, having a `on duplicate key set address=address` when address is the PK is the same as `INSERT IGNORE`.

packages/daemon/src/db/index.ts
```diff
@@ -297,16 +636,18 @@ export const getTxOutputsFromTx = async (
     const utxo: DbTxOutput = {
       txId: result.tx_id as string,
       index: result.index as number,
-      tokenId: result.token_id as string,
+      tokenId: result.token_id as string | null,
       address: result.address as string,
-      value: BigInt(result.value),
+      value: result.value === null || result.value === undefined ? null : BigInt(result.value),
```
tuliomir: suggestion(non-blocking): Centralize this logic into a `parseNullableBigInt(result.value)` function, so that you don't need to repeat it multiple times in this file.

packages/daemon/src/db/index.ts
```diff
@@ -740,28 +1085,24 @@ export const updateAddressTablesWithTx = async (
...
-   * Later, when the corresponding wallet is started, index and walletId will be updated.
-   *
-   * If address is already present, just increment the transactions counter.
-   */
-  const addressEntries = Object.keys(addressBalanceMap).map((address) => [address, 1]);
-  if (addressEntries.length > 0) {
-    await mysql.query(
-      `INSERT INTO \`address\`(\`address\`, \`transactions\`)
-            VALUES ?
-                ON DUPLICATE KEY UPDATE transactions = transactions + 1`,
```
tuliomir, line 756: chore: This previously did `INSERT INTO address ... ON DUPLICATE KEY UPDATE`, which created the address row if absent. That insert is now gone. The function now assumes the row already exists (created upstream by `bumpAddressInvolvement` or `upsertShieldedAddressObservation`).

I confirmed `address_balance` has no FK to address in the migrations — which means there's no hard error to catch a violation. If PR5 ever calls these in the wrong order, you get orphaned `address_balance` / `address_tx_history` rows with no parent address row, silently.

The test had to pre-seed `address2` specifically to paper over this. Worth an explicit integration test asserting the call order, or a comment/assert documenting the precondition.

packages/daemon/src/db/index.ts
```diff
@@ -235,13 +235,352 @@ export const addUtxos = async (
...
+ *
+ * Idempotent against re-ingest of the same vertex: `ON DUPLICATE KEY UPDATE address = VALUES(address)`
+ * re-writes a no-op same value when the (tx_id, index) primary key already exists.
  *
  * @param mysql - Database connection
- * @param inputs - The transaction inputs
+ * @param args - The row to insert
+ */
+export const insertTxOutput = async (
+  mysql: any,
```
tuliomir, line 271: question(non-blocking): Can we type this in a better way?

There are other instances of `any` for `mysql` and in other places of the code too, flagging only this one for reference.

packages/daemon/src/db/index.ts
```diff
@@ -1505,16 +1941,24 @@ export const updateWalletTablesWithTx = async (
...
   if (entries.length > 0) {
     await mysql.query(
       `INSERT INTO \`wallet_tx_history\` (\`wallet_id\`, \`token_id\`,
                                           \`tx_id\`, \`balance\`,
+                                          \`shielded_balance_delta\`,
                                           \`timestamp\`)
-            VALUES ?`,
+            VALUES ?
+       ON DUPLICATE KEY UPDATE \`balance\` = VALUES(\`balance\`),
+                               \`shielded_balance_delta\` = VALUES(\`shielded_balance_delta\`)`,
```
tuliomir: question(non-blocking):  `address_tx_history` updates timestamp in its `ON DUPLICATE KEY UPDATE`.

Should we do it here too?

packages/daemon/src/db/index.ts
```diff
@@ -821,18 +1178,30 @@ export const updateAddressTablesWithTx = async (
...
-    [entries],
-  );
+  if (entries.length > 0) {
+    await mysql.query(
+      `INSERT INTO \`address_tx_history\`(\`address\`, \`tx_id\`,
+                                          \`token_id\`, \`balance\`,
+                                          \`shielded_balance_delta\`,
+                                          \`timestamp\`)
+       VALUES ?
+       ON DUPLICATE KEY UPDATE \`balance\` = VALUES(\`balance\`),
```
tuliomir: chore: `address_tx_history` and `wallet_tx_history` went from `INSERT` to `INSERT ... ON DUPLICATE KEY UPDATE`, but there is no test validating this new behavior.

We should have tests ensuring the idempotent properties of those methods.

packages/daemon/src/db/index.ts
```diff
@@ -235,13 +235,352 @@ export const addUtxos = async (
...
+ *
+ * Invariant: this helper is UPDATE-only. It does NOT insert a `token`
+ * row when one doesn't exist — `applyTokenSupplyUpdates` only ever
+ * mutates existing token rows, and `handleTokenCreated` is the single
+ * inserter (it calls `storeTokenInformation` followed by
+ * `setTokenTotalSupply`). A delta against a not-yet-created token row
+ * naturally no-ops here. The two writers never compete on the same
+ * column.
+ */
+export async function incrementTokenTotalSupply(
```
tuliomir, line 564: chore(test-coverage):

**1) `incrementTokenTotalSupply` underflow.** The doc comment claims "any negative result is a bug worth surfacing (the UPDATE would error)." That's an assertion about strict-mode behavior with no test. If the deploy DB ever runs without `STRICT_TRANS_TABLES` (replicas, a future RDS parameter group), the result silently clamps to `0` instead of erroring — a silent supply corruption.

At minimum there should be a test asserting the throw under the configured mode.

**2) `incrementTokenTotalSupply` against a non-existent token row.** The comment says it "naturally no-ops." Untested. A no-op on a missing row is a silent data loss if the ordering assumption (token created first) is ever violated.

**3) Negative delta that stays `≥ 0` (e.g. melt)**. Untested — this is the normal decrement path.

packages/daemon/src/db/index.ts
```diff
@@ -235,13 +235,352 @@ export const addUtxos = async (
...
+): Promise<void> {
+  const entries: [string, number][] = [];
+  for (const addr of addresses) {
+    entries.push([addr, 1]);
+  }
+  if (entries.length === 0) return;
+  await mysql.query(
+    `INSERT INTO \`address\`(\`address\`, \`transactions\`)
+          VALUES ?
+              ON DUPLICATE KEY UPDATE transactions = transactions + 1`,
```
tuliomir, line 601: chore: I feel we should test this insert behavior. We've changed the function that updates the counter, but added no direct tests to it.

packages/daemon/src/db/index.ts

packages/daemon/src/types/transaction.ts
r4mmer: Well, we can allow undefined in this field but all (current) queries populate this field so it would not mean much

packages/daemon/src/db/index.ts
r4mmer: `TxInput` has `value`, `token_data`, `script`, `token` and `decoded` fields which are not used (and need to be populated from the spent output)
A shielded utxo can be spent by an input, but we cannot populate the fields since the utxo will not have token/value, etc., so we would need to change the `TxInput` to fit the new shielded type, which would change a lot of places since these fields would become optional.
Since this method only uses tx_id and index I changed the argument type so all kinds of inputs can be sent without changing the type across the codebase

packages/daemon/src/db/index.ts
r4mmer: tx timestamp does not change so updating it on all utxos does nothing.

packages/daemon/src/db/index.ts
r4mmer: Maybe we should make this change on its own PR, to make the typing consistent.

packages/daemon/src/db/index.ts
r4mmer: Removed the duplicate statement.
It is safe because the only place that crates this entry is here and it is removed on voided transactions so the only way to have a duplicate entry is to ingest the same tx twice without a void in-between, which should never happen.

packages/daemon/src/db/index.ts
r4mmer: Added tests

packages/daemon/src/db/index.ts
r4mmer: Done

packages/daemon/src/db/index.ts
r4mmer: This is done in a following PR

packages/daemon/src/db/index.ts

pedroferreira1 (approved):

tuliomir (approved):

PR #446, opened 2026-05-29
feat(daemon): handle confidential transactions balance ingestion (5/6)

r4mmer:
## Motivation

With the DB helpers in place (PR4), this PR activates shielded output handling end-to-end. It rewires the ingestion pipeline so that every `VERTEX_ACCEPTED` event processes both transparent and shielded outputs, runs the CT rewind loop for owned shielded addresses, and writes unified per-token balances across both output kinds.

## Summary

**`packages/daemon/src/services/index.ts`** — unified dispatch:
- Recovery loop: for each shielded output with a known owner, calls `rewindAmount`/`rewindFully` and writes `markTxOutputRecovered`; any throw → `markTxOutputRecoveryFailed` + alert
- `bumpAddressInvolvement` now uses the full involved-address set (transparent in/out + shielded out decoded addresses + nano-header)
- `applyTokenSupplyUpdates` gated on transparent outputs only
- `handleTokenCreated` wired via `initial_amount` from the wire event

**`packages/daemon/src/utils/wallet.ts`** — rewritten helpers:
- `getInvolvedAddresses` covers transparent + shielded outputs
- `getUnifiedBalanceMap` builds per-token balance deltas across both kinds
- `prepareInputs` filters to `mode === 0` (transparent-only)
- `unlockUtxos` nullable-safe (removes the `@ts-ignore`s from PR4)

**Mock adaptation** (`__tests__/mocks/ct-crypto-node.ts`):
- Intercepts `../../src/crypto/ctRewind` (wrapper) instead of the former native `@hathor/ct-crypto-node` binding
- Exposes `rewindAmount(args)` / `rewindFully(args)` matching the wrapper API; returns `bigint` values directly (no string-to-bigint conversion)
- `services_with_db.test.ts` updated to `jest.mock('../../src/crypto/ctRewind', …)`

In production, `rewindAmount`/`rewindFully` still throw `RewindError` (the stub from PR1); shielded outputs land in `recovery_failed` until the native binding is wired in a later integration PR. Tests pass because the wrapper is mocked.

## Acceptance Criteria

- [ ] `services/index.ts` imports `rewindAmount`/`rewindFully` from `../crypto/ctRewind`
- [ ] No `@hathor/ct-crypto-node` import anywhere in the daemon source tree
- [ ] `unlockUtxos` has no `@ts-ignore`s (nullable handling is explicit)
- [ ] `prepareInputs` filters out non-zero `mode` outputs
- [ ] `services_with_db.test.ts` mocks `../../src/crypto/ctRewind`, not `@hathor/ct-crypto-node`
- [ ] `ct-crypto-node.ts` mock exports `rewindAmount`/`rewindFully` (wrapper-shaped)
- [ ] Full daemon suite: 21/21 suites, 280/280 tests

## Stack

1. #439 — foundations
2. #442 — migrations
3. #443 — event schema
4. #444 — DB helpers
5. **#446 — ingestion**

### DIFF
--- a/packages/daemon/__tests__/services/services_with_db.test.ts
+++ b/packages/daemon/__tests__/services/services_with_db.test.ts
@@ -5,9 +5,32 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+// Owned-wallet shielded tests trigger wallet-lib's gap-fill / address-derivation
+// path inside handleVertexAccepted (the wallet seed uses a real xpubkey, so each
+// ownership check derives a window of addresses, ~7-8s per test). Jest's default
+// 5000ms timeout fails these in CI; bump to 30s to give headroom. Follow-up: skip
+// gap-fill in these tests by pre-seeding addresses or mocking the derivation.
+jest.setTimeout(30000);
+
+// Redirect the ct-crypto NAPI binding to the deterministic mock for ALL tests
+// in this file. The shielded recovery tests prime the mock; everything else
+// never touches it, so existing tests are unaffected.
+jest.mock('../../src/crypto/ctRewind', () => require('../mocks/ct-crypto-node').mockCtCrypto);
+
+// Stub addAlert so the recovery-failed path doesn't reach SQS; keep the rest
+// of @wallet-service/common live by spreading the real module.
+const mockAddAlert = jest.fn();
+jest.mock('@wallet-service/common', () => {
+  const actual = jest.requireActual('@wallet-service/common');
+  return {
+    ...actual,
+    addAlert: mockAddAlert,
+  };
+});
+
 import { TokenVersion } from '@hathor/wallet-lib';
 import * as db from '../../src/db';
-import { handleVoidedTx, voidTx, handleTokenCreated } from '../../src/services';
+import { handleVoidedTx, voidTx, handleTokenCreated, handleVertexAccepted } from '../../src/services';
 import { LRU } from '../../src/utils';
 import {
   addOrUpdateTx,
@@ -26,9 +49,13 @@ import {
   getWalletBalance,
   insertWalletBalance,
   getWalletTxHistoryCount,
+  XPUBKEY,
 } from '../utils';
 import { DbTxOutput, EventTxInput } from '../../src/types';
 import { Connection } from 'mysql2/promise';
+import eventsFixture from '../__fixtures__/events';
+import { primeAmountRewind, resetCtCryptoMock } from '../mocks/ct-crypto-node';
+import { Severity } from '@wallet-service/common';
 
 /**
  * @jest-environment node
@@ -269,7 +296,7 @@ describe('voidTransaction with input unspending', () => {
     await updateTxOutputSpentBy(mysql, [inputB], txIdB);
     const outputB = createOutput(0, 100n, address2, tokenId);
     await addUtxos(mysql, txIdB, [outputB], null);
-    
+
     // Only update address transaction counts (not balances) to prevent negative decrements
     await mysql.query('INSERT INTO address (address, transactions) VALUES (?, 1) ON DUPLICATE KEY UPDATE transactions = transactions + 1', [address1]);
     await mysql.query('INSERT INTO address (address, transactions) VALUES (?, 1) ON DUPLICATE KEY UPDATE transactions = transactions + 1', [address2]);
@@ -280,8 +307,8 @@ describe('voidTransaction with input unspending', () => {
     await updateTxOutputSpentBy(mysql, [inputC], txIdC);
     const outputC = createOutput(0, 100n, address3, tokenId);
     await addUtxos(mysql, txIdC, [outputC], null);
-    
-    // Only update address transaction counts (not balances) to prevent negative decrements  
+
+    // Only update address transaction counts (not balances) to prevent negative decrements
     await mysql.query('INSERT INTO address (address, transactions) VALUES (?, 1) ON DUPLICATE KEY UPDATE transactions = transactions + 1', [address2]);
     await mysql.query('INSERT INTO address (address, transactions) VALUES (?, 1) ON DUPLICATE KEY UPDATE transactions = transactions + 1', [address3]);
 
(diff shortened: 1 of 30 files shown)


packages/daemon/src/services/index.ts
```diff
@@ -432,18 +530,183 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
...
 
-        // Mark tx utxos as spent
-        await withSpan('updateTxOutputSpentBy', () => updateTxOutputSpentBy(mysql, txInputs, hash));
+        // Resolve a shielded output's token_data to the matching token UID.
+        // Mirrors the transparent-path logic in `prepareOutputs`: a token_data
+        // of 0 selects the native token; otherwise it indexes into `vertex.tokens[]`.
+        const shieldedOutputs = fullNodeEvent.event.data.shielded_outputs ?? [];
+        const transparentCount = txOutputs.length;
+        const resolveShieldedTokenId = (tokenData: number): string | null => {
+          const idx = (tokenData & hathorLib.constants.TOKEN_INDEX_MASK) - 1;
```
pedroferreira1: Don't we have a lib method with this operation?

packages/daemon/__tests__/services/services_with_db.test.ts
```diff
@@ -2107,3 +2134,1004 @@ describe('handleTokenCreated with TokenVersion.FEE (token_version: 2)', () => {
     expect(rows[0].first_block).toBe(newBlock);
   });
 });
+
+describe('handleVertexAccepted with shielded outputs', () => {
```
tuliomir, line 2140: suggestion(non-blocking): Move this `describe` block to a new file. This one is large enough already and should even be split from its current size.

packages/daemon/src/utils/wallet.ts
```diff
@@ -181,59 +184,259 @@ export const getAddressBalanceMap = (
...
+ *    their address as involved so an observer can see something happened.
+ *  - Shielded inputs: `spent_output.decoded.address` — present on all
+ *    shielded spent_output variants.
+ *  - Nano-contract headers: `nc_address`.
+ *
+ * Pure function over wire data; no DB lookups. The caller is responsible
+ * for using the returned set to drive `address.transactions` separately
+ * from the per-(address, token) balance map.
+ */
+export const getInvolvedAddresses = (
```
tuliomir, line 207: suggestion (non-blocking): This is a pure function but has no direct unit test. Add a table-driven one, with no DB interaction. Cases to pin:

- input with no `spent_output` → skipped
- input/output whose decode failed (no address) → skipped
- shielded output with no `decoded.address` → skipped
- nano-header `nc_address` → added
- same address from multiple sources → deduped to one entry

packages/daemon/src/utils/wallet.ts
```diff
@@ -290,22 +493,35 @@ export const unlockTimelockedUtxos = async (mysql: MysqlConnection, now: number)
...
  *                            in the transaction.
- * @returns - An array of prepared inputs, each enriched with additional data.
+ * @returns - An array of prepared transparent inputs, each enriched with additional data.
  */
 export const prepareInputs = (inputs: EventTxInput[], tokens: string[]): TxInput[] => {
   const preparedInputs: TxInput[] = inputs.reduce((newInputs: TxInput[], _input: EventTxInput): TxInput[] => {
     const output = _input.spent_output;
-    // @ts-ignore — SpentOutput union introduced in PR3; prepareInputs filters to transparent in follow-up
+
+    if (output.mode !== 0) {
```
tuliomir, line 516: suggestion (non-blocking): This is a load-bearing filter, but is only tested indirectly — add a direct unit test:

- mixed input array (transparent + AmountShielded + FullyShielded) → only the transparent `TxInput` survives
- transparent custom-token input → token index resolved
- transparent input with invalid decode → `decoded: null`

packages/daemon/src/utils/wallet.ts
```diff
@@ -181,59 +184,259 @@ export const getAddressBalanceMap = (
...
+    map[address] = TokenBalanceMap.merge(map[address], tokenMap);
+  }
+
+  for (const output of txOutputs) {
+    if (!isDecodedValid(output.decoded)) {
+      throw new Error('Output has no decoded script');
+    }
+    if (!output.decoded.address) {
+      throw new Error('Decoded output data has no address');
+    }
```
tuliomir, line 319: suggestion (non-blocking): Unit-test the pure in-memory portion. The DB reversal branches can stay on the indirect/integration tests:

- output with no decoded → throws `Output has no decoded script`
- output decoded but no address → throws `Decoded output data has no address`
- transparent input with invalid decode → skipped
- nano header → empty-HTR seed added

packages/daemon/src/services/index.ts
```diff
@@ -247,6 +261,92 @@ export const isBlock = (version: number): boolean => version === hathorLib.const
...
+ *    and negative for any burn-address loss (since the burned value
+ *    contributes to the inputs side but not the outputs side, naturally
+ *    appearing as a supply decrease).
+ *
+ * Reads from the prepared (transparent) `TxInput[]` / `TxOutputWithIndex[]`,
+ * which already carry resolved `token` UIDs.
+ *
+ * Void/unvoid reversal is wired in a follow-up phase.
+ */
+async function applyTokenSupplyUpdates(
```
tuliomir: suggestion (non-blocking): Pure arithmetic is cheaper to test with unit tests, having `incrementTokenTotalSupply` mocked:

- multi-token tx (two tokens in one delta map)
- `delta === 0` → no write
- block whose summed reward is `0n` → no write

packages/daemon/src/services/index.ts
```diff
@@ -432,18 +530,183 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
...
+                  value: r.value,
+                  token_id: tokenIdHex,
+                });
+                shieldedRecoveryResults.push({
+                  address: so.decoded.address,
+                  tokenId: tokenIdHex,
+                  value: r.value,
+                  locked: shieldedLocked,
+                });
+              } else {
```
tuliomir, line 649: issue: This `else` branch is a real correctness fork from the `AmountShielded` path, and nothing tests it.

The two branches get the token UID from fundamentally different places — `AmountShielded` resolves it from the wire before the rewind (`resolveShieldedTokenId(so.token_data)`), while `FullyShielded` gets it back from the crypto itself (`const tokenIdHexFull = r.tokenUid`).

Because every fixture is `mode === 1`, this branch never runs, so the contract "for FullyShielded, `token_id` comes from the rewind result" is unverified — a refactor that reused `resolveShieldedTokenId` here, or grabbed the wrong field off the result, would pass CI silently.

I'd make this new test branch now rather than defer it with the other shielded gaps: it's a distinct code path, not more-of-the-same, and it's testable today since `primeFullyRewind` already returns a `tokenUid` (the existing AmountShielded recovery test is mock-driven too, so this is no more premature than what's already shipping).

The test would feed a `mode === 2` owned output through `handleVertexAccepted` and assert the recovered row's `token_id` equals the primed `tokenUid`.

packages/daemon/src/services/index.ts
r4mmer: Good call, switched to `hathorLib.tokensUtils.getTokenIndexFromData(tokenData) - 1`, which is the same masking the lib already exposes. (`Output.getTokenIndex()` exists too but needs an `Output` instance; this path only has the raw token_data.)

packages/daemon/__tests__/services/services_with_db.test.ts
r4mmer: Agreed the file is too big. Deferring the split to the last cleanup PR of the stack: the remaining shielded-outputs PRs (void/unvoid/remove parity, realtime) all add tests to this same file, so splitting now would force rebase conflicts across the open stack.

packages/daemon/src/utils/wallet.ts
r4mmer: Added a table-driven unit test in __tests__/utils/wallet.test.ts (no DB) covering:
 - no spent_output → skipped;
 - failed decode (no address) on input/output → skipped;
 - shielded output with no decoded.address → skipped;
 - nano-header nc_address → added;
 - same address from multiple sources → deduped.

packages/daemon/src/utils/wallet.ts
r4mmer: Added a direct unit test in __tests__/utils/wallet.test.ts:
- mixed transparent + AmountShielded + FullyShielded → only the transparent input survives;
- custom-token input → token UID resolved from tokens[];
- transparent input with invalid decode → decoded: null.

packages/daemon/src/utils/wallet.ts
r4mmer: Added unit tests for the in-memory branches (DB reversal stays on integration):
- output with no decoded → throws `Output has no decoded script`;
- decoded but no address → throws `Decoded output data has no address`;
- transparent input with invalid decode → skipped;
- nano header → empty-HTR seed added.

(Calls pass no shielded eventInputs, so the DB handle is never touched.)

packages/daemon/src/services/index.ts
r4mmer: Exported the function and added __tests__/services/applyTokenSupplyUpdates.test.ts with incrementTokenTotalSupply mocked:
- multi-token delta map (two writes)
- delta === 0 → no write
- block with summed reward 0n → no write.

packages/daemon/src/services/index.ts
r4mmer: Done

tuliomir (approved):

pedroferreira1 (approved):

PR #449, opened 2026-06-03
feat(daemon): void parity (7)

r4mmer:
### Motivation

The previous PRs landed the unified-address ingest path for shielded outputs: `handleVertexAccepted` writes both transparent and shielded column families on the same `(address, token_id)` row from a unified balance map.

This PR makes every reversal path symmetric with ingestion. The invariant it enforces: **a voided/removed tx leaves the DB as if it was never ingested**, each effect is reversed using the same source set the ingest path used, with the sign flipped.

Stacked on `feat/shielded-outputs-6-wallet-service`. No schema migrations (all landed in PR 1).

### Summary

**Balance reversal (daemon)**
- `voidTx` builds the unified balance map including shielded contributions, sourced from the local `tx_output` rows (recovered + owned only) about to be voided — the same `getUnifiedBalanceMap` the ingest path uses.
- `voidAddressTransaction` / `voidWalletTransaction` negate both column families per row and extend the zeroed-row cleanup `DELETE` guard with the shielded columns, so a row whose transparent columns reach zero but still holds shielded balance is not wrongly deleted.
- `handleVertexRemoved` reverses balances via the unified `voidTx` before deleting `tx_output` rows; the satellite (`shielded_tx_output_data`) is cleaned by FK `ON DELETE CASCADE`.
- `handleUnvoidedTx` round-trip locked in (cleanup-delete + state-machine re-ingest = credit fresh).

**Involvement counter (daemon)**
- New `decrementAddressInvolvement` reverses `address.transactions` on void/remove over the same wire involvement set (`getInvolvedAddresses`) that `bumpAddressInvolvement` bumps on ingest — including unowned shielded outputs and nano addresses. Previously the void path decremented only balance-map keys, drifting the counter upward.

**Token `total_supply` (daemon)**
- `applyTokenSupplyUpdates` gains a `sign` parameter; void/remove replay block-reward / mint-melt / burn deltas with the sign flipped, unvoid restores via re-ingest. The reversal runs after `deleteTokens` so a token created by the voided vertex is reversed by row deletion (not a delta), avoiding an UNSIGNED underflow.
- `CREATE_TOKEN_TX` vertices are skipped entirely: the created token's supply is owned by `handleTokenCreated` (wire `initial_amount`), and the HTR deposit a creation tx consumes is not a supply change.

**Reorg recompute (wallet-service)**
- `rebuildAddressBalancesFromUtxos` and `getAffectedAddressTotalReceivedFromTxList` partition `tx_output` rows by `mode` in a single aggregation, producing both transparent (`mode = 0`) and shielded (`mode IN (1,2) AND recovery_state = 'recovered'`) balance/total columns; authorities stay transparent-only.

**Out of scope (later phases):** realtime WS + push payload changes (Phase 8); end-to-end integration tests (Phase 9); `@hathor/wallet-lib` bump + `@ts-ignore` removal on `processNftEvent` (Phase 10).

### Acceptance Criteria

- [x] Voiding a recovered shielded receive reverses `unlocked_shielded_balance` / `locked_shielded_balance` and decrements `total_shielded_received`; an unowned/unrecovered receive is a no-op.
- [x] Voiding a vertex that spent a recovered shielded UTXO restores the spent shielded balance and clears `spent_by`.
- [x] The zeroed-row cleanup `DELETE` in both `voidAddressTransaction` and `voidWalletTransaction` requires all shielded columns to be zero (no silent shielded-balance loss).
- [x] `handleVertexRemoved` reverses balances and deletes `tx_output` + cascades the satellite; `handleUnvoidedTx` round-trips back to the post-ingest state.
- [x] `address.transactions` is decremented symmetrically on void/remove for the full involvement set (incl. unowned shielded and nano addresses).
- [x] `token.total_supply` reverses block-reward / mint / melt / burn on void and restores on unvoid; voiding a `CREATE_TOKEN_TX` neither double-counts the created token nor moves HTR, and never underflows.
- [x] `rebuildAddressBalancesFromUtxos` recomputes both transparent and shielded balance/total columns, partitioned by `mode`.
- [x] Full daemon unit suite green (296/296) and wallet-service `db.test.ts` green (92/92); pre-existing transparent tests unchanged.

### Checklist
- [ ] If you are requesting a merge into `master`, confirm this code is production-ready and can be included in future releases as soon as it gets merged `<!-- stacked PR: base is feat/shielded-outputs-6-wallet-service, not master -->`

## Release Notes

* **New Features**
  * Enhanced support for shielded asset management during transaction state transitions, including proper voiding and unvoiding of shielded outputs.

* **Bug Fixes**
  * Improved address balance recalculation for both transparent and shielded assets after transaction reversals.
  * Fixed address involvement tracking during transaction voiding operations.

* **Tests**
  * Expanded test coverage for shielded asset behavior and transaction reversal scenarios.

### DIFF
--- a/packages/daemon/__tests__/services/totalSupply.test.ts
+++ b/packages/daemon/__tests__/services/totalSupply.test.ts
@@ -469,6 +469,41 @@ describe('token.total_supply tracking', () => {
     );
     expect(rows).toHaveLength(0);
   });
+
+  it('CREATE_TOKEN_TX (version 2) leaves HTR and the created token total_supply untouched', async () => {
+    expect.hasAssertions();
+
+    // CREATE_TOKEN_TX: token_id === tx_id. The created token's supply is owned
+    // by handleTokenCreated (from the wire initial_amount); seed it here as if
+    // that event already landed. The vertex also consumes an HTR deposit
+    // (100 in, 99 change). applyTokenSupplyUpdates must skip version-2 entirely
+    // so neither the created token (would otherwise gain the +500 mint delta)
+    // nor HTR (would otherwise lose the 1 HTR deposit) is altered on ingest.
+    const txId = 'a1'.repeat(32);
+    const tokenId = txId;
+
+    await seedToken(HTR_TOKEN_ID, 1_000_000n);
+    await mysql.query(
+      `INSERT INTO token (id, name, symbol, total_supply) VALUES (?, 'Created', 'CRT', 500)`,
+      [tokenId],
+    );
+
+    const data = buildVertexData({
+      hash: txId,
+      version: hathorLib.constants.CREATE_TOKEN_TX_VERSION,
+      tokens: [tokenId],
+      inputs: [transparentInput('parent-tx', 0, 100n, 0, 'addr-depositor')],
+      outputs: [
+        transparentOutput(99, 0, 'addr-change'),
+        transparentOutput(500, 1, 'addr-mint-recipient'),
+      ],
+    });
+
+    await handleVertexAccepted(buildVertexContext(data) as any, undefined as any);
+
+    expect(await selectTotalSupply(HTR_TOKEN_ID)).toStrictEqual(1_000_000n);
+    expect(await selectTotalSupply(tokenId)).toStrictEqual(500n);
+  });
 });
 
 describe('token.total_supply reversal on void / unvoid / remove', () => {
@@ -664,14 +699,14 @@ describe('token.total_supply reversal on void / unvoid / remove', () => {
   it('void of a token-creation vertex deletes the token without underflowing total_supply', async () => {
     expect.hasAssertions();
 
-    // CREATE_TOKEN_TX: token_uid === tx_id by convention. The supply reversal
-    // for a token CREATED by the voided vertex (design path 1) must be its row
-    // deletion, NOT a delta subtraction — mirroring ingest, where the +1 delta
-    // no-ops because the token row doesn't exist yet. We seed initial_amount
-    // (300) deliberately LOWER than the token output sum (1000): if the
-    // signed reversal ran against the still-present token row it would compute
-    // 300 - 1000 and underflow the UNSIGNED total_supply column, erroring out
-    // the whole void. Running it after deleteTokens makes it a clean no-op.
+    // CREATE_TOKEN_TX: token_uid === tx_id by convention. A token CREATED by
+    // the voided vertex is reversed by its row deletion (deleteTokens), never by
+    // a supply delta — applyTokenSupplyUpdates skips version-2 entirely, so the
+    // void reversal does not run for the created token. We seed initial_amount
+    // (300) deliberately LOWER than the token output sum (1000): a signed
+    // reversal against the still-present row would compute 300 - 1000 and
+    // underflow the UNSIGNED total_supply column, erroring out the whole void.
+    // The version-2 skip prevents that.
     const tokenId = 'a'.repeat(64);
     const txId = tokenId;
 
--- a/packages/daemon/__tests__/services/totalSupply.test.ts
+++ b/packages/daemon/__tests__/services/totalSupply.test.ts
@@ -660,4 +660,52 @@ describe('token.total_supply reversal on void / unvoid / remove', () => {
     // delta is reversed.
     expect(await selectTotalSupply(tokenId)).toStrictEqual(1000n);
   });
+
+  it('void of a token-creation vertex deletes the token without underflowing total_supply', async () => {
+    expect.hasAssertions();
+
+    // CREATE_TOKEN_TX: token_uid === tx_id by convention. The supply reversal
+    // for a token CREATED by the voided vertex (design path 1) must be its row
+    // deletion, NOT a delta subtraction — mirroring ingest, where the +1 delta
+    // no-ops because the token row doesn't exist yet. We seed initial_amount
+    // (300) deliberately LOWER than the token output sum (1000): if the
+    // signed reversal ran against the still-present token row it would compute
+    // 300 - 1000 and underflow the UNSIGNED total_supply column, erroring out
+    // the whole void. Running it after deleteTokens makes it a clean no-op.
+    const tokenId = 'a'.repeat(64);
+    const txId = tokenId;
+
+    const data = buildVertexData({
+      hash: txId,
+      version: hathorLib.constants.CREATE_TOKEN_TX_VERSION,
+      tokens: [tokenId],
+      inputs: [],
+      outputs: [transparentOutput(1000, 1, 'addr-mint-recipient')],
+    });
+
+    // Ingest the creation vertex. The token row does not exist yet, so the
+    // forward supply delta no-ops (handleTokenCreated is what inserts it).
+    await handleVertexAccepted(buildVertexContext(data) as any, undefined as any);
+
+    // Simulate the TOKEN_CREATED event: insert the token row + token_creation
+    // record and set total_supply from the (lower) wire initial_amount.
+    // (Insert directly rather than via seedToken — a 64-char token_id would
+    // overflow the token.name column there.)
+    await mysql.query(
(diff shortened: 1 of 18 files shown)


packages/daemon/__tests__/services/totalSupply.test.ts
```diff
@@ -463,4 +469,278 @@ describe('token.total_supply tracking', () => {
...
+    const tokenId = 'token-mint-roundtrip';
+    await seedToken(tokenId, 1000n);
+
+    const data = buildVertexData({
+      hash: '9'.repeat(64),
+      tokens: [tokenId],
+      inputs: [transparentInput('parent-tx', 0, 1n, AUTHORITY_BIT | 1, 'addr-prev-auth')],
+      outputs: [
+        transparentOutput(500, 1, 'addr-recipient'),
+        transparentOutput(1, AUTHORITY_BIT | 1, 'addr-new-auth'),
```
tuliomir, line 623: suggestion(non-blocking): Use the mint token mask directly instead of building the number manually.

```suggestion
        transparentOutput(1, MINT_TOKEN_MASK, 'addr-new-auth'),
```

packages/daemon/src/db/index.ts
```diff
@@ -566,6 +603,36 @@ export async function bumpAddressInvolvement(
...
+): Promise<void> {
+  const entries: [string, number][] = [];
+  for (const addr of addresses) {
+    entries.push([addr, 0]);
+  }
+  if (entries.length === 0) return;
+  await mysql.query(
+    `INSERT INTO \`address\`(\`address\`, \`transactions\`)
+          VALUES ?
+              ON DUPLICATE KEY UPDATE transactions = transactions - 1`,
```
tuliomir, line 631: question: Do we really add a `0` value to a row that doesn't yet exist, or would this try to add a `-1` that would break it?

packages/daemon/src/services/index.ts
```diff
@@ -1024,15 +1040,51 @@ export const voidTx = async (
...
 
-  const addressBalanceMap: StringMap<TokenBalanceMap> = getAddressBalanceMap(txInputs, txOutputsWithLocked, headers);
+  // Build shielded output contributions from local tx_output rows.
+  // A shielded receive is only credited when it was recovered against an
+  // owned address; those same rows must be reversed when the vertex is voided.
+  // We read them from the DB now — before markUtxosAsVoided — so the rows
+  // are still present and still carry value / token_id.
+  //
+  // Ownership is resolved in one batch query to avoid N round-trips.
+  const candidateRows = dbTxOutputs.filter(
```
tuliomir, line 1064: suggestion(test): `findShieldedAddressOwnershipBatch` has three documented behaviors that no test pins directly — only the owned-and-present path is hit (indirectly, via the spend-restore void test).

Worth a small DB-unit test in `db.test.ts` covering: 
  (a) empty-array → empty map (early return), 
  (b) the same address passed twice → one entry (dedup), and 
  (c) **an existing-but-unowned row is absent from the map** (wrong `bip32_account` / NULL `scan_privkey`).

That third case is the discriminator that decides whether a same-address row gets its balance reversed — it should be locked down independently of the void orchestration.

packages/daemon/src/db/index.ts
```diff
@@ -915,6 +976,9 @@ export const voidAddressTransaction = async (
         AND \`locked_balance\` = 0
         AND \`unlocked_authorities\` = 0
         AND \`locked_authorities\` = 0
+        AND \`unlocked_shielded_balance\` = 0
```
tuliomir, line 979: suggestion(test): These shielded guard arms (and their twin in `voidWalletTransaction`, ~L1124) are only ever exercised with `unlocked_shielded_balance` as the survivor (incidentally, via the shielded-spend-restore test).

The AC "a row whose transparent columns reach zero but still holds shielded balance is not wrongly deleted" isn't tested in its stated shape: void a *transparent* tx that shares an `(address, token)` row with an independent recovered shielded balance, and assert the row survives the cleanup DELETE.

A mutation that drops these three `= 0` arms currently keeps the suite green.

packages/daemon/src/services/index.ts
tuliomir, line 1064: suggestion(test), low priority: Every fixture carries a single shielded output, so the `ownershipMap` dedup and the one-recovery-result-per-row mapping here are never exercised against **multiple shielded outputs landing on the same owned address** in one voided tx.

Worth one fixture with 2+ shielded outputs to the same address to confirm the reversal counts each row once.

packages/daemon/src/db/index.ts
```diff
@@ -859,6 +914,9 @@ export const voidAddressTransaction = async (
       { column: '`unlocked_authorities`', op: 'bitor', getValue: (p) => p.balance.unlockedAuthorities.toUnsignedInteger() },
       // locked_authorities: OR only, no recalculation needed — locked authorities can't be spent before unlocking
       { column: '`locked_authorities`', op: 'bitor', getValue: (p) => p.balance.lockedAuthorities.toUnsignedInteger() },
+      { column: '`unlocked_shielded_balance`', op: 'subtract', getValue: (p) => p.balance.unlockedShieldedAmount },
+      { column: '`locked_shielded_balance`', op: 'subtract', getValue: (p) => p.balance.lockedShieldedAmount },
```
tuliomir, line 918: suggestion(test): No reversal test voids a *locked* shielded receive, so the `locked_shielded_balance` subtract (here and at ~L1083) and its `= 0` cleanup-guard arm (~L980 / L1125) never fire with a non-zero value.

The wallet-service rebuild test covers locked shielded, but the daemon void path's locked-shielded handling is untested — dropping these lines keeps the suite green.

packages/daemon/src/db/index.ts
r4mmer: This query is the same as the old one that was part of the void process, I just did the same query here but I agree that setting the value 0 should never happen (since the address row should always exist here)

pedroferreira1 (approved):

packages/daemon/__tests__/services/totalSupply.test.ts
r4mmer: The second argument is token_data, so `AUTHORITY_BIT | 1` means this is an authority for the token 1.
`TOKEN_MINT_MASK` would be the value mask.

We have a convenience constant for `AUTHORITY_BIT | 1` but I think this shows more information to the reader (i.e. authority for token 1) instead of `AUTHORITY_TOKEN_DATA` which only by reading the docstring of the constant you would know its for the first token.

packages/daemon/src/services/index.ts
r4mmer: Done

packages/daemon/src/db/index.ts
r4mmer: Done

packages/daemon/src/services/index.ts
r4mmer: Done

packages/daemon/src/db/index.ts
r4mmer: Done

packages/daemon/src/services/index.ts
```diff
@@ -286,14 +288,17 @@ const BURN_ADDRESS = 'HDeadDeadDeadDeadDeadDeadDeagTPgmn';
...
+ * `-1` when reversing a voided / removed vertex. The shielded gate and the
+ * block short-circuit apply symmetrically to both directions.
  */
 export async function applyTokenSupplyUpdates(
   mysql: any,
   version: number,
   txOutputs: TxOutputWithIndex[],
   txInputs: TxInput[],
   shieldedOutputCount: number,
+  sign: 1 | -1 = 1,
```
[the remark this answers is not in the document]
r4mmer: token supply updates with shielded outputs will be implemented later when we add support for the mint/melt shielded headers.

tuliomir (approved):

pedroferreira1 (approved):

PR #92, opened 2023-10-03
chore: wallet-service monorepo [part 1] - Unify projects

andreabadesso:
### Acceptance Criteria
- The CI should not pass, as it will be implemented in part 2
- We should unify the wallet-service and the daemon projects
- We should have a flake.nix that setups a devshell with yarn and nodejs
- We should use [yarn workspaces](https://classic.yarnpkg.com/lang/en/docs/workspaces/) to build both projects 

### Security Checklist
- [X] Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. More dependencies increase the possibility of one of them being hijacked and affecting us.

PR #450, opened 2026-06-03
feat(push-notification): push notification events (8)

r4mmer:
### Motivation

The realtime `new-tx` message described only transparent outputs, and the push-notification balance summary was computed from transparent inputs/outputs alone, so a recovered shielded receive reached the client through neither channel.

This PR makes the event-delivery paths shielded-aware: the realtime `new-tx` message gains the data a client needs to detect shielded involvement and refetch, and push notifications surface the shielded amount received.

Stacked on `feat/shielded-outputs-daemon-7-void-parity`. No schema migrations.

### Summary

**Realtime `new-tx` (WebSocket)** — additive, backwards-compatible fields inside `data`:
- `shielded_outputs`: a lightweight projection of each shielded output (`mode`, `token_data` for AmountShielded, `decoded.address`) — no on-chain crypto blobs, since the client can't derive encrypted values and only needs to know which addresses were involved.
- `addresses`: the full involved-address set the daemon already computes (the same set used for the involvement counter). The client intersects it with its own addresses and refetches affected balances/history.
- The wallet-service `onNewTx` needs no change — `newTxbodySchema.tx` is a keyless `Joi.object()` that passes the fields through; the contract is documented so it stays that way.

**Push notifications** — surface the shielded amount:
- `getWalletBalancesForTx` (daemon) now takes the unified balance map (transparent + shielded) computed for the vertex instead of recomputing transparent-only, and `TokenBalanceValue` gains a `shieldedAmount` field, sourced from the gross-received accumulator (`totalShieldedReceived`) so it is never negative on spends.
- The wallet-service push handler accepts `shieldedAmount` (Joi, defaulted to 0 for older daemons), notifies when the transparent **or** shielded amount is positive, and combines the two for the displayed amount.
- The per-token sort feeding the notification gate/display is now shielded-aware (ranks by combined transparent + shielded magnitude), so a shielded receive isn't pushed behind transparent tokens and dropped.

### Acceptance Criteria

- [x] The realtime `new-tx` payload carries `shielded_outputs` (projection, no crypto blobs) and `addresses` (involved set incl. shielded output addresses) inside `data`; old clients ignoring them are unaffected.
- [x] A FullyShielded (mode 2) output's address is included in `addresses`; `token_data` is omitted from its projection.
- [x] A recovered shielded receive produces a push entry; `getWalletBalancesForTx` reflects shielded amounts without changing transparent `total`.
- [x] `shieldedAmount` is never negative (0 on a shielded spend); the push notifies and displays the combined transparent + shielded amount.
- [x] A shielded receive accompanied by transparent activity is not mis-gated or hidden — the balance sort ranks by combined magnitude.
- [x] Older daemon payloads without `shieldedAmount` still validate (Joi default 0) and behave as before.
- [x] Full daemon unit suite green (300/300); wallet-service push-handler tests green (13/13); pre-existing transparent tests unchanged.

### Checklist
- [ ] If you are requesting a merge into `master`, confirm this code is production-ready and can be included in future releases as soon as it gets merged `<!-- stacked PR: base is feat/shielded-outputs-daemon-7-void-parity, not master -->`

### DIFF
--- a/packages/daemon/src/services/index.ts
+++ b/packages/daemon/src/services/index.ts
@@ -834,7 +834,7 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
 
           try {
             if (PUSH_NOTIFICATION_ENABLED) {
-              const walletBalanceMap = await getWalletBalancesForTx(mysql, txData);
+              const walletBalanceMap = await getWalletBalancesForTx(mysql, txData, addressBalanceMap);
               const { length: hasAffectWallets } = Object.keys(walletBalanceMap);
               if (hasAffectWallets) {
                 invokeOnTxPushNotificationRequestedLambda(walletBalanceMap)
--- a/packages/daemon/src/services/index.ts
+++ b/packages/daemon/src/services/index.ts
@@ -811,7 +811,9 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
             // same set bumpAddressInvolvement consumed). Clients intersect
             // `addresses` with their own and refetch.
             shielded_outputs: shieldedOutputs.map((so) => {
-              const decoded = so.decoded ? { address: so.decoded.address } : null;
+              // `decoded` is schema-guaranteed present (the socket safeParse rejects
+              // any shielded output without it), matching the unguarded ingest path.
+              const decoded = { address: so.decoded.address };
               // token_data only exists on AmountShielded; FullyShielded hides it.
               return so.mode === ShieldedOutputMode.AmountShielded
                 ? { mode: so.mode, token_data: so.token_data, decoded }
--- a/packages/daemon/src/services/index.ts
+++ b/packages/daemon/src/services/index.ts
@@ -806,6 +806,18 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
             token_name,
             token_symbol,
             signal_bits: 0, // TODO: we should actually receive this and store in the database
+            // Additive realtime fields: a lightweight shielded-output projection
+            // (no crypto blobs) and the full involved-address set (reusing the
+            // same set bumpAddressInvolvement consumed). Clients intersect
+            // `addresses` with their own and refetch.
+            shielded_outputs: shieldedOutputs.map((so) => {
+              const decoded = so.decoded ? { address: so.decoded.address } : null;
+              // token_data only exists on AmountShielded; FullyShielded hides it.
+              return so.mode === ShieldedOutputMode.AmountShielded
+                ? { mode: so.mode, token_data: so.token_data, decoded }
+                : { mode: so.mode, decoded };
+            }),
+            addresses: Array.from(involvedAddresses),
           };
 
           try {
--- a/packages/daemon/src/utils/wallet.ts
+++ b/packages/daemon/src/utils/wallet.ts
@@ -715,17 +715,24 @@ export class FromTokenBalanceMapToBalanceValueList {
       unlockedAuthorities: balance.unlockedAuthorities.toJSON(),
       totalAmountSent: balance.totalAmountSent,
       total: balance.total(),
-      shieldedAmount: balance.unlockedShieldedAmount + balance.lockedShieldedAmount,
+      // Gross shielded received this tx (the lifetime accumulator, suppressed
+      // to 0 on spends), so the receive-oriented push amount is never negative.
+      shieldedAmount: balance.totalShieldedReceived,
     } as TokenBalanceValue));
     return balances;
   }
 }
 
 export const sortBalanceValueByAbsTotal = (balanceA: TokenBalanceValue, balanceB: TokenBalanceValue): number => {
-  function abs(num: bigint) {
-    return num >= 0n ? num : -num;
-  }
-  if (abs(balanceA.total) - abs(balanceB.total) >= 0n) return -1;
+  const abs = (num: bigint): bigint => (num >= 0n ? num : -num);
+  // Rank by the combined magnitude of transparent and shielded movement so a
+  // shielded receive (transparent total 0) isn't pushed behind transparent
+  // tokens — which would otherwise mis-gate it out of the [0]-based push check
+  // or hide it past the top-N display.
+  const magnitudeA = abs(balanceA.total) + abs(balanceA.shieldedAmount);
+  const magnitudeB = abs(balanceB.total) + abs(balanceB.shieldedAmount);
+  if (magnitudeA > magnitudeB) return -1;
+  if (magnitudeA < magnitudeB) return 1;
   return 0;
 };
 
--- a/packages/daemon/src/utils/wallet.ts
+++ b/packages/daemon/src/utils/wallet.ts
@@ -646,8 +646,14 @@ export const validateAddressBalances = async (mysql: MysqlConnection, addresses:
  * @param tx - The transaction to get related wallets and their token balances
  * @returns
  */
-export const getWalletBalancesForTx = async (mysql: MysqlConnection, tx: Transaction): Promise<StringMap<WalletBalanceValue>> => {
-  const addressBalanceMap: StringMap<TokenBalanceMap> = getAddressBalanceMap(tx.inputs, tx.outputs, tx.headers ?? []);
+export const getWalletBalancesForTx = async (
+  mysql: MysqlConnection,
+  tx: Transaction,
+  addressBalanceMap: StringMap<TokenBalanceMap>,
+): Promise<StringMap<WalletBalanceValue>> => {
+  // The caller passes the unified balance map (transparent + shielded) computed
+  // for the vertex, so the push summary reflects recovered shielded receives —
+  // not just the transparent inputs/outputs on `tx`.
   // return only wallets that were started
   const addressWalletMap: StringMap<Wallet> = await getAddressWalletInfo(mysql, Object.keys(addressBalanceMap));
 
@@ -709,6 +715,7 @@ export class FromTokenBalanceMapToBalanceValueList {
       unlockedAuthorities: balance.unlockedAuthorities.toJSON(),
       totalAmountSent: balance.totalAmountSent,
       total: balance.total(),
+      shieldedAmount: balance.unlockedShieldedAmount + balance.lockedShieldedAmount,
     } as TokenBalanceValue));
     return balances;
   }
--- a/packages/wallet-service/src/api/txPushNotificationRequested.ts
+++ b/packages/wallet-service/src/api/txPushNotificationRequested.ts
@@ -56,6 +56,8 @@ class TxPushNotificationRequestValidator {
       unlockedAuthorities: TxPushNotificationRequestValidator.authoritiesSchema,
       lockExpires: Joi.number().integer().min(0).allow(null),
       total: Joi.number().required(),
+      // Optional + defaulted so payloads from daemons predating the field still validate.
+      shieldedAmount: Joi.number().default(0),
     }),
   ).required();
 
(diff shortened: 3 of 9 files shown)


packages/wallet-service/src/types.ts
```diff
@@ -425,6 +425,9 @@ export type TokenBalanceValue = {
   unlockedAuthorities: Record<string, unknown>;
   lockExpires: number | null;
   total: bigint;
+  // Net shielded balance change for this tx (transparent `total` excludes it).
+  // Optional for backwards-compatibility with daemons that predate the field.
+  shieldedAmount?: bigint;
```
pedroferreira1, line 430: Is this type used only by the push notifications? Just thinking if this change could affect other places of the code

pedroferreira1 (approved):

packages/wallet-service/src/types.ts
r4mmer: This type is used in another place, but this place also uses the `FromTokenBalanceMapToBalanceValueList.convert` method that was adapted in this PR, so it is working fine.

There is also another type with the same name on wallet-service, but I dont think it is being used.

packages/wallet-service/src/api/txPushNotificationRequested.ts
```diff
@@ -122,8 +124,12 @@ export const handleRequest: Handler<StringMap<WalletBalanceValue>, { success: bo
     // filter wallets in which the token balance > 0
     .filter((eachSettings) => {
       const wallet = body[eachSettings.walletId];
-      // verify by the first token balance
-      return wallet.walletBalanceForTx[0].total > 0;
+      // verify by the first token balance — count the shielded amount too so a
+      // pure shielded receive (transparent total 0) still notifies.
+      const firstBalance = wallet.walletBalanceForTx[0];
+      // Joi validates these as numbers, so compare as numbers (the declared
+      // bigint types don't reflect the runtime values arriving over JSON).
```
tuliomir, line 131: question(non-blocking): Should we take the opportunity and upgrade these `number`s to `bigint`s?

packages/daemon/src/services/index.ts
```diff
@@ -806,6 +806,18 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
...
             token_symbol,
             signal_bits: 0, // TODO: we should actually receive this and store in the database
+            // Additive realtime fields: a lightweight shielded-output projection
+            // (no crypto blobs) and the full involved-address set (reusing the
+            // same set bumpAddressInvolvement consumed). Clients intersect
+            // `addresses` with their own and refetch.
+            shielded_outputs: shieldedOutputs.map((so) => {
+              const decoded = so.decoded ? { address: so.decoded.address } : null;
+              // token_data only exists on AmountShielded; FullyShielded hides it.
+              return so.mode === ShieldedOutputMode.AmountShielded
```
tuliomir, line 818: suggestion(test): The FullyShielded (mode 2) branch here omits `token_data`, and acceptance criterion number 2 explicitly claims this behavior ("`token_data` is omitted from its projection").

But the only realtime fixture, `VERTEX_WITH_SHIELDED`, carries `mode: 1`, so `newTxRealtime.test.ts` never executes this `: { mode, decoded }` branch.

Mode 2 is well covered for ingest/db/types, but not for this projection — the criterion is checked against a path no test runs.

Can we add a mode-2 case (or a mixed mode-1 + mode-2 fixture) asserting the projected entry has `decoded.address`, the address rides in `addresses`, and `token_data` is absent?

packages/daemon/src/services/index.ts
```diff
@@ -806,6 +806,18 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
             token_name,
             token_symbol,
             signal_bits: 0, // TODO: we should actually receive this and store in the database
+            // Additive realtime fields: a lightweight shielded-output projection
+            // (no crypto blobs) and the full involved-address set (reusing the
+            // same set bumpAddressInvolvement consumed). Clients intersect
+            // `addresses` with their own and refetch.
+            shielded_outputs: shieldedOutputs.map((so) => {
+              const decoded = so.decoded ? { address: so.decoded.address } : null;
```
tuliomir: suggestion(test): `so.decoded ? … : null` defends against a null `decoded`, but the ingest path on the same vertex dereferences `so.decoded.address` / `so.decoded.timelock` unguarded (`services/index.ts:580`-ish), so a null `decoded` would already have thrown before reaching this projection.

Is the `: null` branch reachable in practice? If it is, ingest has a latent crash; if it isn't, this is dead defensive code. Either way no test covers the null path — worth resolving the inconsistency one way or the other.

packages/wallet-service/src/api/txPushNotificationRequested.ts
```diff
@@ -174,7 +180,9 @@ const _assembleSpecificMessage = (deviceId: string, txId: string, tokenBalanceLi
 
   const tokens = [];
   for (const eachBalance of tokenBalanceList.slice(0, upperLimit)) {
-    const amount = eachBalance.total;
+    // Combine the transparent and shielded amounts for the displayed value.
+    // Joi validates these as numbers; add as numbers to avoid a bigint/number mix.
+    const amount = Number(eachBalance.total) + Number(eachBalance.shieldedAmount ?? 0);
```
tuliomir, line 185: suggestion(test): This `Number(total) + Number(shieldedAmount)` combine is the core new behavior, but every test uses a pure case — transparent-only, or shielded-only (`total:0, shielded:150` → `"150 T2"`).

Nothing exercises a single token with both amounts positive (e.g. `total:100, shielded:50` → expected `"150"`). As written the suite would stay green if this were a `-`, a `Math.max`, or dropped a term. Same gap on the gate at line 132 (`|| shieldedAmount > 0`).

Could we add a both-positive case for one token?

packages/wallet-service/src/api/txPushNotificationRequested.ts
```diff
@@ -122,8 +124,12 @@ export const handleRequest: Handler<StringMap<WalletBalanceValue>, { success: bo
     // filter wallets in which the token balance > 0
     .filter((eachSettings) => {
       const wallet = body[eachSettings.walletId];
-      // verify by the first token balance
-      return wallet.walletBalanceForTx[0].total > 0;
+      // verify by the first token balance — count the shielded amount too so a
+      // pure shielded receive (transparent total 0) still notifies.
+      const firstBalance = wallet.walletBalanceForTx[0];
```
tuliomir, line 129: suggestion(test): This gate trusts `walletBalanceForTx[0]`, relying on the daemon having sorted shielded-first.

Criterion number 5 ("shielded receive accompanied by transparent activity is not mis-gated/hidden") is only tested in halves: the comparator in isolation (daemon `sortBalanceValueByAbsTotal` unit test) and this gate in isolation (shielded-only, so `[0]` is trivially the shielded token).

No test feeds a wallet holding `[small transparent token, large shielded receive]` through the handler to prove the shielded entry actually lands at `[0]` and survives the gate.

A single end-to-end case across that seam would lock the criterion down.

packages/daemon/src/utils/wallet.ts
```diff
@@ -709,16 +715,24 @@ export class FromTokenBalanceMapToBalanceValueList {
...
-  if (abs(balanceA.total) - abs(balanceB.total) >= 0n) return -1;
+  const abs = (num: bigint): bigint => (num >= 0n ? num : -num);
+  // Rank by the combined magnitude of transparent and shielded movement so a
+  // shielded receive (transparent total 0) isn't pushed behind transparent
+  // tokens — which would otherwise mis-gate it out of the [0]-based push check
+  // or hide it past the top-N display.
+  const magnitudeA = abs(balanceA.total) + abs(balanceA.shieldedAmount);
+  const magnitudeB = abs(balanceB.total) + abs(balanceB.shieldedAmount);
+  if (magnitudeA > magnitudeB) return -1;
+  if (magnitudeA < magnitudeB) return 1;
```
tuliomir, line 735: suggestion(test): This is now a proper three-way comparator, whereas the old one returned `-1` on ties (`>= 0n`), violating the sort contract — a good fix.

But the `return 0` equal-magnitude branch is untested; the unit test only compares 10 vs 500.

A tie case (e.g. two tokens with equal combined magnitude) would lock in the corrected behavior and document that it's intentional.

packages/daemon/src/utils/wallet.ts
```diff
@@ -709,16 +715,24 @@ export class FromTokenBalanceMapToBalanceValueList {
       unlockedAuthorities: balance.unlockedAuthorities.toJSON(),
       totalAmountSent: balance.totalAmountSent,
       total: balance.total(),
+      // Gross shielded received this tx (the lifetime accumulator, suppressed
+      // to 0 on spends), so the receive-oriented push amount is never negative.
+      shieldedAmount: balance.totalShieldedReceived,
```
tuliomir, line 720: suggestion(test): `shieldedAmount: balance.totalShieldedReceived` is documented as gross received, not net.

Tests cover pure receive (gross == net == 150) and pure spend (gross == 0), but neither distinguishes gross from net — that only shows up in a tx that both receives and spends shielded for the same token (e.g. receive 100, spend 300 → net −200, gross 100).

A mixed case asserting `shieldedAmount === 100` would actually pin the "gross" semantic the comment claims.

packages/daemon/src/utils/wallet.ts
```diff
@@ -709,16 +715,24 @@ export class FromTokenBalanceMapToBalanceValueList {
...
     } as TokenBalanceValue));
     return balances;
   }
 }
 
 export const sortBalanceValueByAbsTotal = (balanceA: TokenBalanceValue, balanceB: TokenBalanceValue): number => {
```
tuliomir, line 726: suggestion(test): There's a second copy of `sortBalanceValueByAbsTotal` on `packages/wallet-service/src/db/utils.ts:155-160`, untouched by this PR.

I's still transparent-only and still the asymmetric `>= 0n return -1; return 0` form (never returns `1`) and consumed by `WalletBalanceMapConverter.toValue`.

The active push path uses the daemon's sorted output, so this isn't hit today, but it's a latent divergence: the moment any wallet-service-driven path carries `shieldedAmount`, this mis-sorts and the `[0]` gate drops shielded receives — exactly the bug the daemon-side change fixes.

Should this copy be updated in lockstep (or the two de-duplicated into the `common` package), and is it worth a guard test so they can't drift?

tuliomir (approved):

packages/daemon/src/utils/wallet.ts
r4mmer: Confirmed — this copy is still the asymmetric >= 0n → -1; return 0 form and its convert never emits shieldedAmount, so it would mis-sort once shielded amounts flow through it.

That path is dormant today, though: this comparator is only reached via WalletBalanceMapConverter.toValue → getWalletBalancesForTx (commons.ts:607), and that function has no production  caller, it's referenced only by its own unit test (legacy from the monorepo unification in #92).

Rather than patch the dormant copy in this PR I would prefer to remove the dead code in a future PR.

packages/daemon/src/services/index.ts
r4mmer: Done

packages/daemon/src/services/index.ts
r4mmer: Done

packages/wallet-service/src/api/txPushNotificationRequested.ts
r4mmer: Done

packages/wallet-service/src/api/txPushNotificationRequested.ts
r4mmer: Done

packages/daemon/src/utils/wallet.ts
r4mmer: Done

packages/daemon/src/utils/wallet.ts
r4mmer: Done

packages/wallet-service/src/api/txPushNotificationRequested.ts
r4mmer: I rather not increase the scope of this PR.
Also, since the number values here come from the request body we need to check how we would validate the bigint values.

pedroferreira1 (approved):

tuliomir (approved):

PR #459, opened 2026-06-19
feat(wallet-service): forward token deposit percentage fraction from /version

raul-oliveira:
### Motivation

[hathor-core PR #1711](https://github.com/HathorNetwork/hathor-core/pull/1711) exposes the token deposit percentage on `/version` as an integer fraction in **parts per billion** — `token_deposit_percentage_numerator` (`10^7`) / `token_deposit_percentage_denominator` (`10^9` = 1%) — alongside the now-deprecated float `token_deposit_percentage`. The fraction lets wallet clients compute token deposit/withdraw amounts with exact integer math instead of float multiplication, which loses precision for large amounts.

This PR makes the wallet-service support and forward those new fields end to end, so wallet clients receive them from the wallet-service `/version` endpoint, while staying backward compatible with older fullnodes that only return the float.

### Acceptance Criteria

- The `/version` endpoint forwards `tokenDepositPercentageNumerator` / `tokenDepositPercentageDenominator` (camelCase) when the connected fullnode provides them.
- Older fullnodes that return only the float `token_deposit_percentage` keep working: the new fields are simply absent from the response (no breaking change, no migration needed — the value flows through the existing JSON `version_data` cache).
- The schema validates the two new fields as optional positive integers and rejects non-integer values.
- The deprecated float `token_deposit_percentage` is accepted as nullable (`number | null`) and marked `@deprecated` in the types, anticipating fullnodes dropping it in the future.
- Unit tests cover: schema acceptance of a valid numerator/denominator pair, schema rejection of a non-integer numerator, `convertApiVersionData` forwarding the fraction, and `convertApiVersionData` omitting it for fullnodes that don't provide it.

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for exact token deposit percentages via `numerator`/`denominator` integer fields.
  * Kept the legacy floating-point `tokenDepositPercentage` as deprecated/optional for compatibility.

* **Bug Fixes**
  * Improved `/version` input validation: at least one deposit-percentage representation is required, and numerator/denominator must be provided together.

* **Tests**
  * Expanded tests covering schema validation, API `/version` responses, and conversion behavior for both legacy and fraction-based inputs.

### DIFF
--- a/packages/wallet-service/src/nodeConfig.ts
+++ b/packages/wallet-service/src/nodeConfig.ts
@@ -51,6 +51,9 @@ export function convertApiVersionData(data: FullNodeApiVersionResponse): FullNod
     minTxWeightCoefficient: data.min_tx_weight_coefficient,
     minTxWeightK: data.min_tx_weight_k,
     tokenDepositPercentage: data.token_deposit_percentage,
+    // Integer fraction that lets clients do exact integer math; both fields are absent on older fullnodes.
+    tokenDepositPercentageNumerator: data.token_deposit_percentage_numerator,
+    tokenDepositPercentageDenominator: data.token_deposit_percentage_denominator,
     rewardSpendMinBlocks: data.reward_spend_min_blocks,
     maxNumberInputs: data.max_number_inputs,
     maxNumberOutputs: data.max_number_outputs,
--- a/packages/wallet-service/src/schemas.ts
+++ b/packages/wallet-service/src/schemas.ts
@@ -24,7 +24,10 @@ export const FullnodeVersionSchema = Joi.object<FullNodeApiVersionResponse>({
   min_tx_weight: Joi.number().integer().positive().allow(0).required(),
   min_tx_weight_coefficient: Joi.number().positive().allow(0).required(),
   min_tx_weight_k: Joi.number().integer().positive().allow(0).required(),
+  // Deprecated float, always present; superseded by the integer numerator/denominator fraction.
   token_deposit_percentage: Joi.number().positive().required(),
+  token_deposit_percentage_numerator: Joi.number().integer().positive(),
+  token_deposit_percentage_denominator: Joi.number().integer().positive(),
   reward_spend_min_blocks: Joi.number().integer().positive().required(),
   max_number_inputs: Joi.number().integer().positive().required(),
   max_number_outputs: Joi.number().integer().positive().required(),
--- a/packages/wallet-service/src/schemas.ts
+++ b/packages/wallet-service/src/schemas.ts
@@ -24,8 +24,8 @@ export const FullnodeVersionSchema = Joi.object<FullNodeApiVersionResponse>({
   min_tx_weight: Joi.number().integer().positive().allow(0).required(),
   min_tx_weight_coefficient: Joi.number().positive().allow(0).required(),
   min_tx_weight_k: Joi.number().integer().positive().allow(0).required(),
-  // Deprecated float, always present; superseded by the integer numerator/denominator fraction.
-  token_deposit_percentage: Joi.number().positive().required(),
+  // Deprecated float, now optional: fullnodes may drop it in favor of the integer fraction below.
+  token_deposit_percentage: Joi.number().positive(),
   token_deposit_percentage_numerator: Joi.number().integer().positive(),
   token_deposit_percentage_denominator: Joi.number().integer().positive(),
   reward_spend_min_blocks: Joi.number().integer().positive().required(),
@@ -40,7 +40,20 @@ export const FullnodeVersionSchema = Joi.object<FullNodeApiVersionResponse>({
     symbol: Joi.string().min(1).max(5).required(),
     version: Joi.number().integer(),
   }),
-}).unknown(true);
+})
+  .unknown(true)
+  // Reject a response with no deposit-percentage source at all: without the deprecated
+  // float nor the integer fraction, clients would silently fall back to a compiled-in default.
+  .or(
+    'token_deposit_percentage',
+    'token_deposit_percentage_numerator',
+    'token_deposit_percentage_denominator',
+  )
+  // The integer fraction is only usable as a complete pair, so require both halves together.
+  .and(
+    'token_deposit_percentage_numerator',
+    'token_deposit_percentage_denominator',
+  );
 
 export const EnvironmentConfigSchema = Joi.object<EnvironmentConfig>({
   defaultServer: Joi.string().required(),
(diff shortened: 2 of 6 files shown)


packages/wallet-service/src/nodeConfig.ts
```diff
@@ -51,6 +51,13 @@ export function convertApiVersionData(data: FullNodeApiVersionResponse): FullNod
     minTxWeightCoefficient: data.min_tx_weight_coefficient,
     minTxWeightK: data.min_tx_weight_k,
     tokenDepositPercentage: data.token_deposit_percentage,
+    // Forward the #1711 integer fraction so clients can do exact integer math (absent on older fullnodes).
```
raul-oliveira: what about returning undefined instead of using the spread operator?

raul-oliveira: remove the reference to the pr in the comments

packages/wallet-service/src/schemas.ts
```diff
@@ -25,6 +25,8 @@ export const FullnodeVersionSchema = Joi.object<FullNodeApiVersionResponse>({
   min_tx_weight_coefficient: Joi.number().positive().allow(0).required(),
   min_tx_weight_k: Joi.number().integer().positive().allow(0).required(),
   token_deposit_percentage: Joi.number().positive().required(),
```
raul-oliveira: check if this schema can be nullable. Since it will be deprecated

packages/wallet-service/src/types.ts
```diff
@@ -95,6 +95,8 @@ export interface FullNodeVersionData {
   minTxWeightCoefficient: number;
   minTxWeightK: number;
   tokenDepositPercentage: number;
```
raul-oliveira: this number will be deprecated in the future

packages/wallet-service/src/types.ts
```diff
@@ -118,7 +120,10 @@ export interface FullNodeApiVersionResponse {
   min_tx_weight: number;
   min_tx_weight_coefficient: number; // float
   min_tx_weight_k: number;
-  token_deposit_percentage: number; // float
+  token_deposit_percentage: number; // float; deprecated in favor of the integer fraction below
+  // Deposit percentage as an integer fraction in parts per billion (#1711); absent on older fullnodes.
```
raul-oliveira: remove the reference to the pr

raul-oliveira (commented):
claude, address the comments

packages/wallet-service/src/nodeConfig.ts
raul-oliveira: Claudiao: Done. Replaced the conditional spread with direct assignment, so each field is forwarded as `undefined` when the fullnode doesn't provide it. Both type fields are optional, and the served `/version` JSON is unchanged since `JSON.stringify` drops `undefined`-valued keys.

raul-oliveira: Claudiao: Removed the `#1711` reference from the comment.

packages/wallet-service/src/schemas.ts
raul-oliveira: Claudiao: Made it nullable — `Joi.number().positive().allow(null).required()`. Kept `.required()` so presence is still guaranteed for current and older fullnodes, but `null` is now accepted for when the float gets deprecated. Updated both `FullNodeApiVersionResponse.token_deposit_percentage` and the converted `FullNodeVersionData.tokenDepositPercentage` to `number | null` so the types match the schema. Verified no consumer reads the field (`txProposalCreate` only uses `maxNumberOutputs`).

packages/wallet-service/src/types.ts
raul-oliveira: Claudiao: Marked it `@deprecated` pointing to `tokenDepositPercentageNumerator`/`tokenDepositPercentageDenominator`, and changed it to `number | null` to match the now-nullable schema.

packages/wallet-service/src/types.ts
raul-oliveira: Claudiao: Removed the `#1711` reference from the comment.

packages/wallet-service/src/schemas.ts
tuliomir: This field is still not nullable nor optional. I think it should be one of both at least.

packages/wallet-service/tests/schemas.test.ts
```diff
@@ -41,3 +41,23 @@ test('FullnodeVersionSchema rejects a non-integer native_token.version', () => {
...
+  const payload = {
+    ...defaultTestVersionData(),
+    token_deposit_percentage_numerator: 10000000,
+    token_deposit_percentage_denominator: 1000000000,
+  };
+  const { error } = FullnodeVersionSchema.validate(payload);
+  expect(error).toBeUndefined();
+});
+
+test('FullnodeVersionSchema rejects a non-integer token_deposit_percentage_numerator', () => {
```
tuliomir, line 77: question: Should we have a business rule where `token_deposit_percentage_numerator / denominator` cannot be defined at the same time with `token_deposit_percentage`?

packages/wallet-service/src/schemas.ts
raul-oliveira: Done — `token_deposit_percentage` is now optional (dropped `.required()`), and both type interfaces (`FullNodeApiVersionResponse` and the converted `FullNodeVersionData`) were updated to match.

While addressing this I also added group-level guards on the schema so optionality doesn't open a silent-fallback hole:

- `.or('token_deposit_percentage', 'token_deposit_percentage_numerator', 'token_deposit_percentage_denominator')` — at least one source must be present, so a response with none of them is rejected instead of silently falling back to the compiled-in default downstream.
- `.and('token_deposit_percentage_numerator', 'token_deposit_percentage_denominator')` — the integer fraction is only usable as a complete pair, so a half-populated fraction (one half without the other) is rejected.

Covered by new tests in `schemas.test.ts` (float omitted but fraction present → accepted; all three absent → rejected; numerator-without-denominator and denominator-without-numerator → rejected).

packages/wallet-service/tests/schemas.test.ts
raul-oliveira: We intentionally did not add a mutual-exclusivity rule. The fullnode (core #1711) returns both the deprecated float and the new integer fraction during the transition period, so rejecting "both defined at the same time" would break every current `/version` response.

Instead the schema now enforces two complementary rules:

- `.or('token_deposit_percentage', 'token_deposit_percentage_numerator', 'token_deposit_percentage_denominator')` — at least one of the deposit-percentage fields must be present.
- `.and('token_deposit_percentage_numerator', 'token_deposit_percentage_denominator')` — the fraction must come as a complete pair.

Prioritization of the new fraction over the deprecated float is the consumer's responsibility, not the wallet-service (which is a pass-through). wallet-lib `Storage.getTokenDepositPercentageFraction()` uses the numerator/denominator when both are present and only falls back to the float otherwise, so when both arrive together the fraction wins. That is covered by `storage.test.ts` ("should prefer the numerator/denominator fields from the version", which sets the float and the fraction simultaneously and asserts the fraction is used).

tuliomir (approved):

pedroferreira1 (approved):

PR #473, opened 2026-07-07
feat(shielded-outputs): wallet registration core

r4mmer:
## Motivation

Shielded outputs hide an amount/token on-chain; a wallet only sees a shielded
receive by rewinding the commitment with its scan key. The foundations PR gave
the daemon **live** recovery, but a wallet that registers its keys **after** its
outputs were observed still needs to recover its **historical** shielded receives,
and that recovery runs in the wallet-service. This PR is the shared base for it:
the CT crypto + derivation move to `common`, and the wallet-service gains a
recovery/reconstruction engine. Registration, the async worker, daemon
gap-extension, and the retry helper build on this.

## Acceptance criteria

- [ ] CT rewind wrapper + shielded address derivation live in `common` and are
      importable by the wallet-service; daemon is repointed and its suites stay green.
- [ ] Derived shielded addresses match `@hathor/wallet-lib` exactly (parity test).
- [ ] `reconstructWallet` recovers a wallet's historical shielded receives and folds
      them into `wallet_balance` / `wallet_tx_history` alongside transparent balances,
      **idempotently** (safe to re-run / error-restart).
- [ ] No CT addresses (old client) → a clean transparent-only reconstruction.
- [ ] All tests + `check-types` green (recovery exercised against a mock provider).

## Summary of changes

**`common`**
- Move the CT rewind wrapper `daemon → common` with `@hathor/ct-crypto-provider`
  types + a registered-provider seam (unset → throws until the real provider ships).
- Upgrade `@hathor/wallet-lib` 2.12.0 → 4.0.0.
- Add shielded address derivation (`deriveCtAddress` et al.) — fast `bitcoinjs`,
  parity-tested against wallet-lib.

**`wallet-service`**
- `db/shielded.ts` — ownership resolution; recovery-state marks (guard
  `<> 'recovered'`); non-recovered getter with keyset-cursor pagination; four
  rebuild steps (shielded `address_*` from `tx_output`; unified `wallet_*`
  aggregation subsuming `initWalletBalance`/`initWalletTxHistory`).
- `shieldedRecovery.ts` — `rewind → mark → alert`, the `findAndRewindShielded`
  driver, and `reconstructWallet(walletId, transparentAddresses, ctSpendAddresses)`.
- One-time idempotent DB-reconstruction (not delta-apply); `recovery_failed`
  durable + re-driven, no reset; receive-side history only (spend-side follow-up).

## Testing
`common` rewind-seam + derivation parity; 25 wallet-service tests (recovery, rebuilds,
end-to-end reconstruction) against a migrated DB; `check-types` clean.

## Summary by CodeRabbit

* **New Features**
  * Added shielded address derivation/encoding and shielded output rewind support to shared wallet services.
  * Implemented shielded-output recovery and end-to-end wallet reconstruction, including database rebuilds for shielded balances and history.
* **Bug Fixes**
  * Improved recovery robustness: failures are safely recorded, alerts are reported without blocking recovery, and retried outputs can recover later.
  * Updated wallet-lib compatibility across affected packages to align with version 4.0.0.
* **Tests**
  * Added/expanded Jest coverage for rewind/provider behavior, shielded derivation, recovery flows, reconstruction idempotency, and shielded DB rebuild routines.


packages/wallet-service/src/db/shielded.ts
```diff
@@ -0,0 +1,340 @@
...
+
+/**
+ * Rebuild a wallet's `wallet_balance` rows by aggregating its addresses'
+ * `address_balance` (both transparent and shielded column families), with the
+ * per-token `transactions` count taken from `address_tx_history` (distinct txs,
+ * so shared txs aren't double-counted). Upsert → idempotent. This subsumes the
+ * transparent-only `initWalletBalance`, so it credits caught-up shielded receives
+ * in the same pass old clients get their transparent balance.
+ */
+export const rebuildWalletBalance = async (
```
tuliomir, line 264: suggestion(test): Add more fixtures to `rebuildWalletBalance`

### "subsumes `initWalletBalance`" is under-verified (authorities + timelock)

`rebuildWalletBalance` aggregates `BIT_OR(unlocked_authorities)`, `BIT_OR(locked_authorities)`, and `MIN(timelock_expires)` — this is the part that's supposed to replace the transparent-only `initWalletBalance`. But the `insertAddressBalance` test helper hardcodes `unlocked_authorities=0, locked_authorities=0, timelock_expires=NULL` in **every** test. So those three aggregations are executed but never asserted against non-trivial values.

**Regression that stays green:** drop the `BIT_OR` on authorities (or the `MIN` on timelock), and a wallet holding a mint/melt authority — or timelocked funds — reconstructs with its authority/timelock silently lost. Nothing in the suite catches it. Given the PR's own acceptance criterion is that reconstruction folds shielded receives in "alongside transparent balances," the transparent-equivalence half deserves at least one row with real authority bits and a timelock.

packages/wallet-service/src/db/shielded.ts
```diff
@@ -0,0 +1,340 @@
...
+}
+
+/**
+ * Read a page of the wallet's not-yet-recovered shielded outputs (`unowned` or
+ * `recovery_failed` alike), joined to the per-index scan key (`address`) and the
+ * on-chain crypto bytes (`shielded_tx_output_data`) needed to rewind them. So a
+ * catch-up re-drives previously-failed outputs too, and failed ones are never
+ * downgraded to `unowned`. Ordered by (tx_id, index) so pagination is stable.
+ */
+export const getShieldedOutputsToRecover = async (
```
tuliomir, line 140: suggestion(test): Voided exclusion tests

### voided-exclusion is asserted in test *names* but never exercised

Every rebuild query and `getShieldedOutputsToRecover` filter on `voided = FALSE`, but no test ever inserts a voided row:

- `db.shielded.test.ts` — the test literally named `'excludes recovered, voided, and other-wallet outputs'` (line ~1661) never inserts a voided output. The `insertShieldedOutput` helper hardcodes `voided=FALSE`, so it *can't*. Delete `AND t.voided = FALSE` from `getShieldedOutputsToRecover` and this test still passes.
- `rebuildShieldedAddressBalances`, `rebuildShieldedAddressTxHistory`, `rebuildWalletBalance`, `rebuildWalletTxHistory` all filter `voided = FALSE` and none is tested with a voided row (even though `db.shieldedReconstruct.test.ts`'s `insertOutput` helper *does* support `voided`).

**Regression that stays green:** a voided (reversed) shielded receive gets counted into the wallet balance and history. This is a real correctness rule — voided txs must not contribute — and it's completely unverified.

packages/wallet-service/src/db/shielded.ts
```diff
@@ -0,0 +1,340 @@
...
+                BIT_OR(\`locked_authorities\`) AS \`locked_authorities\`,
+                SUM(\`unlocked_shielded_balance\`) AS \`unlocked_shielded_balance\`,
+                SUM(\`locked_shielded_balance\`) AS \`locked_shielded_balance\`,
+                SUM(\`total_shielded_received\`) AS \`total_shielded_received\`
+           FROM \`address_balance\`
+          WHERE \`address\` IN (?)
+          GROUP BY \`token_id\`
+       ) b
+       LEFT JOIN (
+         SELECT \`token_id\`, COUNT(DISTINCT \`tx_id\`) AS \`transactions\`
```
tuliomir, line 295: suggestion(test): Less priority than the ones above

### `COUNT(DISTINCT tx_id)` distinctness not verified in the balance path

`rebuildWalletBalance`'s per-token `transactions` count uses `COUNT(DISTINCT tx_id)`. The balance test uses two distinct txs on two addresses (count=2). The *shared-tx-across-two-addresses* case is only tested in `rebuildWalletTxHistory` (the `'shared'` tx), which doesn't assert the count column.

**Regression that stays green:** change it to `COUNT(tx_id)` (non-distinct) and a wallet whose single tx touches two of its own addresses double-counts its transaction total. No test catches it.

packages/wallet-service/src/shieldedRecovery.ts
```diff
@@ -0,0 +1,167 @@
...
+
+/**
+ * Find and rewind all of a wallet's not-yet-recovered shielded outputs — a
+ * registration catch-up that also re-drives any `recovery_failed` rows, so an
+ * error-restart needs no separate reset. Pages forward with a (tx_id, index)
+ * keyset cursor: a re-driven output that fails again stays in the set, so the
+ * cursor (rather than set membership) is what guarantees the loop advances and
+ * terminates. Never throws — a failed output is marked + alerted and counted.
+ */
+export const findAndRewindShielded = async (
```
tuliomir, line 128: suggestion(test): Also medium priority

### keyset cursor's index tie-breaker is never exercised

The cursor is `(tx_id > ? OR (tx_id = ? AND index > ?))`, but every pagination test (`'honours the limit and cursors forward'`, and the pageSize-2 test in `findAndRewindShielded`) uses distinct `tx_id`s all at `index 0`. The `index > ?` tie-breaker branch never runs against a real multi-output tx.

**Regression that stays green:** break the index tie-breaker and a shielded tx with multiple outputs to the same wallet that straddles a page boundary could skip or repeat an output. This is plausible — the history test itself uses `rtx2` at index 0 and 1 — so it's a realistic shape that pagination never sees.

packages/wallet-service/src/shieldedRecovery.ts
```diff
@@ -0,0 +1,167 @@
...
+ * One-time seed of a wallet's balances + history from current DB state, unified
+ * across the two derivation paths. Transparent addresses are already daemon-
+ * maintained, so they only feed the wallet-level aggregation; CT-spend addresses
+ * are first found + rewound and their `address_*` rebuilt from `tx_output`.
+ * Everything is recompute-from-source, so a repeat (or an error-restart) is safe.
+ *
+ * Passing an empty `ctSpendAddresses` (an old client with no CT keys) yields a
+ * clean transparent-only reconstruction.
+ */
+export const reconstructWallet = async (
```
tuliomir, line 161: suggestion(test): Low priority


### `reconstructWallet` composite idempotency isn't tested

Idempotency ("safe to re-run / error-restart") is a headline acceptance criterion. Each rebuild function is individually re-run, but `reconstructWallet` as a whole is never invoked twice. The second-pass interaction (already-`recovered` outputs → `getShieldedOutputsToRecover` returns empty → rebuilds re-snapshot) is only reasoned about in comments, not asserted.

### FullyShielded (mode 2) never flows end-to-end

`findAndRewindShielded` and `reconstructWallet` fixtures hardcode `mode 1` (`insertUnownedOutput`). Mode 2 is only covered in the `recoverShieldedOutput` unit test. So the FullyShielded path where **token_id comes from the rewind** (a non-`'00'` token) folding into `rebuildShieldedAddressBalances`'s `GROUP BY token_id` is never exercised at the aggregation level. This overlaps with the fact that *every* rebuild fixture uses token `'00'` — the multi-token grouping is untested, which matters most precisely for shielded outputs that can recover arbitrary tokens.

packages/wallet-service/src/db/shielded.ts
```diff
@@ -0,0 +1,340 @@
...
+  shieldedIndex: number;
+  scanPrivkey: Buffer;
+}
+
+/**
+ * Resolve the CTSpend ownership (wallet + per-index scan key) for a shielded
+ * spend address. Returns null unless a wallet has claimed the address and its
+ * scan key is recorded.
+ */
+export const findShieldedAddressOwnership = async (
```
tuliomir, line 26: suggestion(test): Low priority

### `findShieldedAddressOwnership` null-guards not independently tested

The `returns null for transparent/unclaimed/unknown` test uses a `Legacy` address. The `scan_privkey IS NOT NULL` and `wallet_id IS NOT NULL` guards on a *CTSpend* row are never hit independently — a CTSpend address with a null scan key would be a regression source (an "owned" address you can't actually rewind) that no test pins. (Also worth noting: this function doesn't appear to be called anywhere in *this* PR's reconstruct path — likely staged for the registration follow-up.)

pedroferreira1 (dismissed):

packages/wallet-service/src/shieldedRecovery.ts
```diff
@@ -0,0 +1,167 @@
...
+        tx_id: output.txId,
+        index: output.index,
+        wallet_id: walletId,
+        error: String(e),
+        source: 'wallet-service',
+      },
+      logger,
+    );
+    return { ...base, recovered: false };
+  }
```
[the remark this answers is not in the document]
r4mmer: Fixed in `c7a447c`. The catch path now wraps `markShieldedTxOutputRecoveryFailed` + `addAlert` in an inner try/catch that logs and swallows, so a transient failure there can't escape `recoverShieldedOutput` and abort the batch. A swallowed mark leaves the output non-`recovered`, so the next catch-up re-drives it — and a full DB outage still aborts on the next `getShieldedOutputsToRecover` (outside the guard), so we don't mask a real outage. Covered by `never rejects even if failure-reporting (addAlert) throws` in `shieldedRecovery.test.ts`.

packages/wallet-service/src/db/shielded.ts
r4mmer: Added in `1cd9fac`. `insertAddressBalance`/`readWalletBalance` now carry authority bits + timelock, and `folds authority bits (BIT_OR) and the earliest timelock (MIN) across addresses` (`db.shieldedReconstruct.test.ts`) asserts `BIT_OR(01,10)=11`, `BIT_OR(00,01)=01`, `MIN(900,500)=500`. Dropping either aggregation now fails.

packages/wallet-service/src/db/shielded.ts
r4mmer: Added in `1cd9fac`, covering both voided filters:

- `tx_output.voided` — `excludes recovered, voided, and other-wallet outputs` now inserts an own-wallet unowned **voided** row (the helper gained a `voided` param), and the `rebuildShieldedAddressBalances` recompute test adds a voided output that must be excluded from `total_shielded_received`.
- `address_tx_history.voided` — new `excludes voided address-history rows from the wallet aggregation` (`rebuildWalletTxHistory`).

Dropping any `voided = FALSE` now fails a test.

packages/wallet-service/src/db/shielded.ts
r4mmer: Added in `1cd9fac` — `counts a tx touching two wallet addresses once (COUNT DISTINCT)` seeds one tx on two of the wallet's addresses and asserts `transactions = 1`; switching to non-distinct `COUNT(tx_id)` yields 2 and fails.

packages/wallet-service/src/shieldedRecovery.ts
r4mmer: Added in `1cd9fac` — `cursors within a single tx_id using the index tie-breaker` pages three outputs at indexes 0/1/2 under one `tx_id` with `pageSize 2`, so the cursor must use the `index > ?` branch. Dropping it returns an empty second page and fails.

packages/wallet-service/src/shieldedRecovery.ts
r4mmer: Added in `1cd9fac`, all three sub-gaps:

- composite idempotency — `is idempotent end-to-end (safe to re-run)` runs `reconstructWallet` twice and asserts the balance/history don't double (second pass returns `{recovered:0}`).
- mode-2 end-to-end + multi-token — `recovers a fully-shielded (mode 2) output and folds a second token via GROUP BY token_id` drives a FullyShielded output (token from the rewind) alongside a mode-1 `'00'` output through the rebuilds and asserts one `wallet_balance` row per token.

packages/wallet-service/src/db/shielded.ts
r4mmer: Added in `1cd9fac` — `returns null for a CTSpend address missing its wallet or scan key` seeds one CTSpend row with null `scan_privkey` and one with null `wallet_id`, each isolating one guard; dropping either makes a row resolve and fails. Kept in this PR even though the only caller lands with the registration wiring.

r4mmer:
Follow-ups on the remaining nitpicks:

**`VALUES(col)` in the shielded upserts** — correct that it's soft-deprecated on our MySQL 8.0 target, but it's the repo-wide convention: `src/db/index.ts` uses `VALUES(...)` throughout and the row-alias form (`AS new … = new.col`) appears nowhere. Converting only `shielded.ts` would make it the lone inconsistent file, so leaving it here; better handled as a `src/db`-wide migration if/when we do one.

**`rewindFully` error-wrapping test** — valid parity gap. It's a `packages/common` test, outside the wallet-service surface this PR hardens, so tracking it for a common-package follow-up rather than adding it here.

**Optional mode-2 test in `reconstructWallet.test.ts`** — done in `1cd9fac` (`recovers a fully-shielded (mode 2) output and folds a second token via GROUP BY token_id`).

pedroferreira1 (approved):

tuliomir (approved):

PR #474, opened 2026-07-10
feat(shielded-outputs): wallet registration request + reconciliation

r4mmer:
## Summary

First PR of the **shielded wallet-registration** stack (design `0001`), on top of the shielded engine from #473. Lets a client register shielded scan/spend keys via the existing `POST /wallet/init` — at creation or as an upgrade of an existing transparent wallet — with cryptographic proof, and surfaces a unified wallet status. Scope is the **request → validate → persist → status** layer; the combined async load body and the daemon catch-up scan are follow-ups.

## Acceptance criteria

- [ ] The five shielded fields (`scanXpriv`, `spendXpub`, `firstCtAddress`, `spendXpubSignature`, `ctAddressSignature`) are accepted **all-or-none** (`400` if partial).
- [ ] Registration is rejected unless: the keys re-derive to the submitted `firstCtAddress`, `spendXpub` is self-signed, and the first `ct_address` is signed by the auth key (`403`/`400` otherwise).
- [ ] A **transparent-only** request runs the existing (now `deprecated`) transparent load, unchanged.
- [ ] A request **carrying shielded keys** runs the single canonical combined load (`invokeLoadWalletAsync`) instead — never the transparent load and a separate catch-up (avoids double work + a `wallet_balance` write race). An already-loaded transparent wallet can **upgrade** through the same combined load (idempotent).
- [ ] Re-submitting identical keys is a no-op (no load re-triggered); a different key set returns `409`.
- [ ] The response `status` is the unified transparent+shielded value, plus `shieldedMaxGap` / `lastUsedShieldedIndex`; `scan_xpriv` never appears in any response.

## Proof scheme

1. **Matched pair** — server re-derives the first `ct_address` from `scanXpriv`+`spendXpub` and requires it to equal `firstCtAddress` (forces same-account keys).
2. **Spend-key control** — `spendXpubSignature` is `spendXpub` signed by the spend key.
3. **Auth consent + anti-replay** — `ctAddressSignature` is the first `ct_address` signed by the auth key. Both messages use `timestamp || walletId || payload`, so the existing timestamp check blocks replay.

## Async load routing

The combined transparent+shielded load is the canonical path; the transparent-only load is deprecated:

| serverless function | handler | TS invoke |
|---|---|---|
| `loadWalletAsync` | `wallet.loadWalletCombined` (**stub** — body next PR) | `invokeLoadWalletAsync` (shielded requests) |
| `deprecatedLoadWalletAsync` | `wallet.loadWallet` (existing, unchanged) | `invokeDeprecatedLoadWalletAsync` (transparent-only) |

## Notable notes

- `verifySignature` refactored onto reusable `buildAuthMessage` + `verifyMessageSignature`.
- `WALLET_ALREADY_LOADED` is bypassed when shielded fields are present (enables upgrade).
- `scan_xpriv` stored as UTF-8 bytes of the base58 xpriv (matches the shipped `BLOB` column, shaped for future encryption-at-rest); `spend_xpub` stays a string.

## Testing

TDD — new tests cover: signature helpers, all-or-none schema, the five proof cases, unified-status truth table, the DB helper, and the full reconciliation flow through the handler — asserting the **combined** load fires (and the deprecated one never does) for new/upgrade, and **neither** fires for no-op / `409` / `403`, plus a no-`scan_xpriv`-leak assertion. Existing transparent load/wallet suite green (switched to the deprecated-invoke spy).

> Aggregate coverage % is validated by CI — the local arm64 host OOMs `tiny-secp256k1` WASM partway through the full suite. The combined worker is a stub, so a shielded registration leaves the wallet in `creating` until the follow-up PR.

## Out of scope (follow-ups)

- Implement `loadWalletCombined`: derive transparent + shielded ownership rows and `reconstructWallet(both)`, then flip statuses to ready.
- Daemon catch-up scan (`address.catchup_state`).
- Rate-limit; encryption-at-rest for `scan_xpriv`.

## Summary by CodeRabbit

* **New Features**
  * Added shielded wallet registration with proof validation (first shielded address and required signatures) for POST `/wallet`.
  * Wallet responses now return a unified transparent+shielded lifecycle, including shielded metadata (`max gap`, `last used index`) with defaults.
  * Persist and upgrade shielded keys; introduced `SHIELDED_MAX_ADDRESS_GAP` configuration.
* **Bug Fixes**
  * Shielded registration is rejected before any wallet state is modified.
  * Idempotent re-submission of identical keys; mismatched keys return **409 Conflict**.
  * Improved authenticated message/signature verification.
* **Tests**
  * Expanded test coverage for shielded registration, conflict cases, unified status, and key persistence.


packages/wallet-service/src/db/index.ts
```diff
@@ -355,6 +371,35 @@ export const updateWalletAuthXpub = async (
...
+  shieldedMaxGap: number,
+): Promise<void> => {
+  await mysql.query(
+    `UPDATE \`wallet\`
+        SET \`scan_xpriv\` = ?,
+            \`spend_xpub\` = ?,
+            \`shielded_max_gap\` = ?,
+            \`ct_status\` = 'creating'
+      WHERE \`id\` = ?`,
+    [Buffer.from(scanXpriv, 'utf8'), spendXpub, shieldedMaxGap, walletId],
```
pedroferreira1, line 399: Why do you save `scanXpriv` as buffer?

packages/wallet-service/src/api/wallet.ts
```diff
@@ -53,6 +59,34 @@ const mysql = getDbConnection();
 
 const MAX_LOAD_WALLET_RETRIES: number = config.maxLoadWalletRetries;
 
+// Network instance used to encode shielded (ct) addresses — carries the
+// shielded/p2pkh version bytes that address derivation reads.
+const shieldedNetwork = new Network(config.network);
+
+// Default shielded gap limit (matches the transparent default and the
+// wallet.shielded_max_gap column default).
+const DEFAULT_SHIELDED_MAX_GAP = 20;
```
pedroferreira1: I think we have an env var with the MAX_GAP for the legacy addresses, don't we?

packages/wallet-service/serverless.yml
```diff
@@ -241,6 +241,24 @@ functions:
...
+    warmup:
+      walletWarmer:
+        enabled: false
+    iamRoleStatements:
+      - Effect: Allow
+        Action:
+          - SNS:Publish
+        Resource:
+          arn:aws:sns:${self:provider.region}:${self:provider.environment.ACCOUNT_ID}:WalletServiceLoadAsyncFailed_${self:custom.stage}
+  deprecatedLoadWalletAsync:
```
pedroferreira1: This name will probably be over the lambda limit, we should decrease it somehow `deprecatedLoad` maybe

packages/wallet-service/src/db/index.ts
r4mmer: The column type of `scanXpriv` is blob instead of varchar because we may want to encrypt this field later.
The encryption process would be able to encrypt in place without a column type change.

packages/wallet-service/src/api/wallet.ts
r4mmer: The legacy uses the `config.maxAddressGap`.
Want me to use the same gap value?

tuliomir: I think it's important to have this well defined, since the Wallet Lib Integration Tests use a custom gap limit to differentiate tests between its two facades.

Adding a new configurable gap limit should also make sure it ends up in the `/status` response and may end up demanding new test scenarios from its clients too.

packages/wallet-service/tests/api.test.ts
```diff
@@ -2373,3 +2384,323 @@ test('GET /address/info', async () => {
...
+    const result = await walletLoad(makeGatewayEvent({}, JSON.stringify(body)), null, null) as APIGatewayProxyResult;
+    // A shielded upgrade must not be rejected as WALLET_ALREADY_LOADED.
+    expect(result.statusCode).toBe(200);
+
+    const wallet = await Db.getWallet(mysql, getWalletId(XPUBKEY));
+    expect(wallet.ctStatus).toBe('creating');
+    expect(wallet.scanXpriv).toBe(body.scanXpriv);
+    // The upgrade runs the combined load (which idempotently re-reconstructs
+    // transparent too), not the deprecated transparent-only load.
+    expect(combined).toHaveBeenCalledTimes(1);
```
tuliomir, line 2661: suggestion(test):

The PR's headline output is the **unified transparent+shielded status** (`toWalletStatusResponse` → `computeUnifiedStatus(wallet.status, wallet.ctStatus)`). But no handler test ever asserts a response `status` for an input where the unified value *differs* from the raw transparent `status`:

- The new-wallet test (line 2581) asserts `status.status === CREATING`, but there transparent status is already `CREATING` — identity.
- This upgrade test seeds a `ready` transparent wallet, so after the upgrade transparent is `READY` while `ct_status` is `creating` → unified should be `CREATING`. That is exactly the divergent case — and the test only asserts `wallet.ctStatus` from a fresh DB read, never `returnBody.status.status`.
- The "re-runs after failed load" test (line 2677) has transparent `error` + ct `creating` → unified `ERROR`, also unasserted in the response.

Regression that passes the whole suite green: change `toWalletStatusResponse` to `status: wallet.status` (dropping `computeUnifiedStatus`). Every handler test still passes, and `computeUnifiedStatus`'s own unit tests in `db.test.ts` pass because they call the function directly, not through the response mapping. A client polling an upgrading wallet would then see `ready` while the shielded load is still running.

Add a response-status assertion to this upgrade test:
```ts
expect(returnBody.status.status).toBe(WalletStatus.CREATING);
```
(and ideally one on the failed-load test at line 2677 asserting `ERROR`).

packages/wallet-service/tests/api.test.ts
```diff
@@ -2373,3 +2384,323 @@ test('GET /address/info', async () => {
...
+    const now = Math.floor(Date.now() / 1000);
+    const body = { ...buildShieldedLoadBody(now), ctAddressSignature: Buffer.from('invalid').toString('base64') };
+
+    const result = await walletLoad(makeGatewayEvent({}, JSON.stringify(body)), null, null) as APIGatewayProxyResult;
+    expect(result.statusCode).toBe(403);
+    // No wallet should have been created — shielded validation runs before any load.
+    expect(await Db.getWallet(mysql, getWalletId(XPUBKEY))).toBeNull();
+    expect(combined).not.toHaveBeenCalled();
+    expect(deprecated).not.toHaveBeenCalled();
+  }, SHIELDED_TEST_TIMEOUT_MS);
```
tuliomir, line 2712: suggestion(test):

There is no handler test for the shielded **combined-load invocation failure**. Every test in this describe block installs `invokeLoadWalletAsync` via `spyInvokes()` with `mockResolvedValue(undefined)` (line 2564), so the shielded `try/catch` → `onAsyncInvokeError` branch (mark wallet `error`, bump `retryCount`, re-read status) never executes.

The existing failure tests (POST /wallet line 929; "loadWallet should update wallet status to ERROR") exercise `invokeDeprecatedLoadWalletAsync` — a different call site.

Regression that passes green: break the shielded error path only — e.g. drop its `try/catch`, or fail to increment `retryCount` for the combined load. A shielded registration whose async invoke throws would surface an unhandled rejection / stay stuck, and no test notices.

Add a shielded test mirroring the transparent failure path:
```ts
    const { combined } = spyInvokes();
    combined.mockRejectedValueOnce(new Error('invoke failed'));
    // ... POST a valid shielded body on a fresh wallet ...
    expect(result.statusCode).toBe(200);
    expect(returnBody.status.status).toBe(WalletStatus.ERROR);
    expect(returnBody.status.retryCount).toBe(1);
```

packages/wallet-service/tests/api.test.ts
```diff
@@ -2373,3 +2384,323 @@ test('GET /address/info', async () => {
...
+    await mysql.query('UPDATE `wallet` SET `retry_count` = 5 WHERE `id` = ?', [walletId]);
+    mockedAddAlert.mockClear();
+
+    const result = await walletLoad(makeGatewayEvent({}, JSON.stringify(body)), null, null) as APIGatewayProxyResult;
+    expect(result.statusCode).toBe(400);
+    expect(JSON.parse(result.body as string).error).toBe(ApiError.WALLET_MAX_RETRIES);
+    // No retry is kicked off, and reaching the cap raises an ops alert.
+    expect(combined).not.toHaveBeenCalled();
+    expect(deprecated).not.toHaveBeenCalled();
+    expect(mockedAddAlert).toHaveBeenCalledTimes(1);
```
tuliomir, line 2754: suggestion(test): Low priority

`expect(mockedAddAlert).toHaveBeenCalledTimes(1)` verifies the ops alert *fired* but not *what* it carried. A regression that raises the alert at the wrong `Severity`, with an empty message, or with the wrong `wallet_id`/`retry_count` metadata passes green — yet that payload is the entire operational value of the alert.

Pin the payload:
```ts
    expect(mockedAddAlert).toHaveBeenCalledWith(
      'Wallet load exceeded max retries',
      expect.stringContaining(walletId),
      Severity.MINOR,
      expect.objectContaining({ wallet_id: walletId, retry_count: 5 }),
      expect.anything(),
    );
```

packages/wallet-service/tests/api.test.ts
```diff
@@ -2373,3 +2384,323 @@ test('GET /address/info', async () => {
...
+    // types here; the proof is validated in a later step).
+    const event = makeGatewayEvent({}, JSON.stringify({
+      xpubkey, xpubkeySignature, authXpubkey, authXpubkeySignature, firstAddress, timestamp,
+      scanXpriv: 'scan', spendXpub: 'spend', firstCtAddress: 'ct',
+      spendXpubSignature: 'sps', ctAddressSignature: 'cts',
+    }));
+
+    const result = await walletLoad(event, null, null) as APIGatewayProxyResult;
+    // Before the schema accepted these fields, the unknown keys yielded 400 INVALID_PAYLOAD.
+    expect(JSON.parse(result.body as string).error).not.toBe(ApiError.INVALID_PAYLOAD);
```
tuliomir: suggestion(test): Low priority

This test is named "accepts a complete shielded field set" and mocks `invokeLoadWalletAsync` at line 2532 as if the load will be reached — but the dummy values (`'scan'`, `'spend'`, …) make `validateShieldedRegistration` fail (`deriveCtAddress` throws → 400) *before* any load, so the mock is never consumed.

The lone assertion `.not.toBe(ApiError.INVALID_PAYLOAD)` only proves the schema didn't reject the five fields as unknown keys; it asserts neither the status code nor that a complete set is actually accepted end-to-end (the "all-present ⇒ accepted" half of all-or-none).

It works as a narrow schema regression-guard, but the name over-promises.

Tighten to what it actually checks, and drop the unreachable mock:
```ts
expect(result.statusCode).not.toBe(400); // fails only on a schema INVALID_PAYLOAD
// or, to truly exercise acceptance, use a valid buildShieldedLoadBody() and assert 200.
```

packages/wallet-service/serverless.yml
r4mmer: Done

packages/wallet-service/tests/api.test.ts
r4mmer: `computeUnifiedStatus` is well tested and it does include these cases.

packages/wallet-service/tests/api.test.ts
r4mmer: Done

packages/wallet-service/tests/api.test.ts
r4mmer: Done

packages/wallet-service/tests/api.test.ts
r4mmer: Done

tuliomir (dismissed):

packages/wallet-service/src/api/wallet.ts
pedroferreira1: Yes, let's keep the same gap limit for legacy and shielded addresses. I'm using this logic in the wallet lib as well

pedroferreira1 (dismissed):

pedroferreira1 (approved):

packages/wallet-service/tests/api.test.ts
tuliomir, line 1280: chore(non-blocking): [Out of Scope] unified-status wrapping of `changeAuthXpub` is unasserted

`changeAuthXpub` ( `api/wallet.ts:295` ) now returns `status: toWalletStatusResponse(updatedWallet)` instead of the raw wallet.

The change is benign for existing clients (for a no-shielded wallet `computeUnifiedStatus` collapses to the transparent status, and the two shielded fields are additive), but its only success-path test — `PUT /wallet/auth should change the auth_xpub …` — is an `expect(1).toStrictEqual(1)` placeholder, so nothing actually verifies the new response shape on this endpoint.

If the endpoint matters to a client, an assertion on the returned `status` object would close the gap.

tuliomir (approved):

packages/wallet-service/tests/api.test.ts
r4mmer: Yes, the change is intentional since the updatedWallet now contains the scan xpriv which should not be sent back in responses.

PR #475, opened 2026-07-14
feat(shielded-outputs): combined transparent+shielded wallet load

r4mmer:
## Summary

Implements `loadWalletCombined` — the canonical async wallet-load Lambda — replacing the stub PR #474 left behind. It derives both the transparent and shielded (CTSpend) address paths, claims ownership rows, runs `reconstructWallet` to rebuild both balances, and flips the wallet to ready.

The worker mirrors the transparent `loadWallet` template:

1. **Read-only derivation** — transparent `generateAddresses` plus a new shielded gap loop that infers usage from observed `tx_output` rows (no rewinding). Already-claimed rows are reused from storage, so a retry re-derives nothing.
2. **One short claim transaction** — upsert address ownership + advance the frontier indices.
3. **`reconstructWallet` outside any transaction** — it internally rewinds owned outputs and rebuilds the balance/history tables; idempotent by recompute, and must not hold locks while doing crypto + paged UPDATEs.
4. **A settle rewind pass** — closes the claim-visibility race against in-flight daemon ingests.
5. **Status flips.**

## Acceptance criteria

- **Fresh load** — a new shielded wallet claims rows, recovers, credits balances, and flips both sides ready.
- **Upgrade** — a shielded upgrade to an already-READY transparent wallet leaves the transparent side untouched.
- **Idempotent retry** — a second run converges without double-crediting.
- **Error handling** — a fresh-load failure marks both sides; an upgrade failure pins only the shielded side.
- **Transparent-only** — a wallet without shielded keys behaves exactly as before.
- **Gap invariant** — the derivation window extends past chained usage; a transparent (mode 0) output to a shielded spend address counts as usage; voided outputs do not.
- **Daemon partition** — transparent gap math ignores CTSpend rows. The CT index pair the unified query now returns is intentionally unconsumed (daemon-side shielded gap extension is a follow-up).

## Follow-ups (explicit non-goals)

- Daemon live shielded gap-extension — will consume the CT index pair (`maxCtWalletIndex`/`maxCtAmongAddresses`) this PR's unified query already returns, plus `last_used_shielded_index` advancement on recovery and a daemon-side claim upsert.
- The post-`ready` re-drive sweep — must key on `tx_output.recovery_state` joined to claimed CTSpend rows, **not** on `catchup_state`.
- `shielded-catchup-complete` realtime event.
- Real ct-crypto provider wiring — blocked on a `@hathor/ct-crypto-node` release that bundles a provider.
- Spend-side history for pre-registration spends — accepted-wrong; balances remain correct.

## Summary by CodeRabbit

* **New Features**
  * Wallet loading now follows a single unified async flow to rebuild both transparent (legacy) and shielded (CTSpend) balances, address windows, and transaction history.
  * Added CTSpend address discovery with per-index ownership tracking and catch-up completion.

* **Bug Fixes**
  * Transparent gap-extension now derives strictly from legacy index tracking (CTSpend progress no longer affects it).
  * Wallet “max index” calculations are partitioned by legacy vs CTSpend, preventing cross-impact.
  * Improved concurrency-safe load retries/state transitions to avoid demotion and double-crediting, including clearer legacy vs shielded error pinning.
  * Canonicalized shielded token IDs to prevent duplicate “all-zero” token rows.

### DIFF
--- a/packages/daemon/src/services/index.ts
+++ b/packages/daemon/src/services/index.ts
@@ -668,9 +668,13 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
               continue;
             }
 
-            const { maxAmongAddresses, maxWalletIndex } = indices;
+            // Transparent gap extension reads the transparent pair only — a
+            // claimed shielded (CTSpend) index must never drive or suppress
+            // transparent derivation. The CT pair feeds the shielded gap
+            // extension (follow-up work).
+            const { maxTransparentAmongAddresses, maxTransparentWalletIndex } = indices;
 
-            if (maxAmongAddresses == null || maxWalletIndex == null) {
+            if (maxTransparentAmongAddresses == null || maxTransparentWalletIndex == null) {
               // Do nothing, wallet is most likely not loaded yet.
               if (walletDetails.status === WalletStatus.READY) {
                 logger.error('[ERROR] A wallet marked as READY does not have a max wallet index or address index was not found in the database');
@@ -678,12 +682,12 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
               continue;
             }
 
-            const diff = maxWalletIndex - maxAmongAddresses;
+            const diff = maxTransparentWalletIndex - maxTransparentAmongAddresses;
 
             if (diff < walletDetails.maxGap) {
               // We need to generate addresses
-              const addresses = await generateAddresses(NETWORK as string, walletDetails.xpubkey, maxWalletIndex + 1, walletDetails.maxGap - diff);
-              await addNewAddresses(mysql, walletId, addresses, maxAmongAddresses);
+              const addresses = await generateAddresses(NETWORK as string, walletDetails.xpubkey, maxTransparentWalletIndex + 1, walletDetails.maxGap - diff);
+              await addNewAddresses(mysql, walletId, addresses, maxTransparentAmongAddresses);
             }
           }
 
--- a/packages/daemon/src/services/index.ts
+++ b/packages/daemon/src/services/index.ts
@@ -668,13 +668,13 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
               continue;
             }
 
-            // Transparent gap extension reads the transparent pair only — a
+            // Legacy gap extension reads the legacy pair only — a
             // claimed shielded (CTSpend) index must never drive or suppress
-            // transparent derivation. The CT pair feeds the shielded gap
+            // legacy derivation. The CT pair feeds the shielded gap
             // extension (follow-up work).
-            const { maxTransparentAmongAddresses, maxTransparentWalletIndex } = indices;
+            const { maxLegacyAmongAddresses, maxLegacyWalletIndex } = indices;
 
-            if (maxTransparentAmongAddresses == null || maxTransparentWalletIndex == null) {
+            if (maxLegacyAmongAddresses == null || maxLegacyWalletIndex == null) {
               // Do nothing, wallet is most likely not loaded yet.
               if (walletDetails.status === WalletStatus.READY) {
                 logger.error('[ERROR] A wallet marked as READY does not have a max wallet index or address index was not found in the database');
@@ -682,12 +682,12 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
               continue;
             }
 
-            const diff = maxTransparentWalletIndex - maxTransparentAmongAddresses;
+            const diff = maxLegacyWalletIndex - maxLegacyAmongAddresses;
 
             if (diff < walletDetails.maxGap) {
               // We need to generate addresses
-              const addresses = await generateAddresses(NETWORK as string, walletDetails.xpubkey, maxTransparentWalletIndex + 1, walletDetails.maxGap - diff);
-              await addNewAddresses(mysql, walletId, addresses, maxTransparentAmongAddresses);
+              const addresses = await generateAddresses(NETWORK as string, walletDetails.xpubkey, maxLegacyWalletIndex + 1, walletDetails.maxGap - diff);
+              await addNewAddresses(mysql, walletId, addresses, maxLegacyAmongAddresses);
             }
           }
 
--- a/packages/daemon/src/services/index.ts
+++ b/packages/daemon/src/services/index.ts
@@ -561,7 +561,7 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
                   rangeProof: range,
                   assetCommitment: assetCommit,
                 });
-                const tokenIdHexFull = r.tokenUid;
+                const tokenIdHexFull = r.tokenUid; // canonicalized by rewindFully
                 await markTxOutputRecovered(mysql, hash, idx, {
                   value: r.value,
                   token_id: tokenIdHexFull,
--- a/packages/daemon/src/types/db.ts
+++ b/packages/daemon/src/types/db.ts
@@ -149,8 +149,8 @@ export interface TokenSymbolsRow extends RowDataPacket {
 }
 
 export interface MaxAddressIndexRow extends RowDataPacket {
-  max_transparent_among_addresses: number | null,
-  max_transparent_wallet_index: number | null,
+  max_legacy_among_addresses: number | null,
+  max_legacy_wallet_index: number | null,
   max_ct_among_addresses: number | null,
   max_ct_wallet_index: number | null,
 }
--- a/packages/daemon/src/types/db.ts
+++ b/packages/daemon/src/types/db.ts
@@ -161,6 +161,8 @@ export interface AddressesWalletsRow extends RowDataPacket {
   auth_xpubkey: string,
   xpubkey: string,
   maxGap: number,
+  status: string,
+  ct_status: string,
 }
 
 export interface AddressRow extends RowDataPacket {
--- a/packages/daemon/src/types/db.ts
+++ b/packages/daemon/src/types/db.ts
@@ -149,8 +149,10 @@ export interface TokenSymbolsRow extends RowDataPacket {
 }
 
 export interface MaxAddressIndexRow extends RowDataPacket {
-  max_among_addresses: number,
-  max_wallet_index: number
+  max_transparent_among_addresses: number | null,
+  max_transparent_wallet_index: number | null,
+  max_ct_among_addresses: number | null,
+  max_ct_wallet_index: number | null,
 }
 
 export interface AddressesWalletsRow extends RowDataPacket {
--- a/packages/wallet-service/src/api/wallet.ts
+++ b/packages/wallet-service/src/api/wallet.ts
@@ -26,6 +26,7 @@ import {
   advanceLastUsedShieldedIndex,
   markWalletLoadReady,
   markWalletLoadError,
+  markLegacyLoadError,
 } from '@src/db';
 import {
   GenerateShieldedAddresses,
@@ -74,6 +75,14 @@ import errorHandler from '@src/api/middlewares/errorHandler';
 const mysql = getDbConnection();
 
 const MAX_LOAD_WALLET_RETRIES: number = config.maxLoadWalletRetries;
+
+/**
+ * Alert-safe error detail. `String(e)` drops the stack of a real Error, leaving
+ * the alert with only a message and no way to join it to the logged throw.
+ */
+const errorDetail = (e: unknown): string => (
+  e instanceof Error ? (e.stack ?? `${e.name}: ${e.message}`) : String(e)
+);
 
 // Network instance used to encode shielded (ct) addresses — carries the
 // shielded/p2pkh version bytes that address derivation reads.
@@ -528,12 +537,17 @@ export const load: APIGatewayProxyHandler = middy(async (event) => {
   }
 
   /* The async loads below are invoked as "Event", so we don't care about the
-   * response here, only whether the invocation itself failed. On failure the
-   * wallet is marked 'error' and the latest status is re-read for the response. */
+   * response here, only whether the invocation itself failed. On failure only the
+   * side this request was loading is marked 'error' — an upgrade whose invoke
+   * fails must leave the already-ready legacy wallet working — and the latest
+   * status is re-read for the response. */
   const onAsyncInvokeError = async (e: unknown): Promise<void> => {
     logger.error(e);
-    const newRetryCount = wallet.retryCount ? wallet.retryCount + 1 : 1;
-    await updateWalletStatus(mysql, walletId, WalletStatus.ERROR, newRetryCount);
+    if (hasShielded) {
+      await markWalletLoadError(mysql, walletId, wallet.status !== WalletStatus.READY);
+    } else {
+      await markLegacyLoadError(mysql, walletId);
+    }
     wallet = await getWallet(mysql, walletId);
   };
 
@@ -797,11 +811,19 @@ export const loadWallet: Handler<LoadEvent, LoadResult> = async (event) => {
     //    from tx_output), so everything revealed lands in a single rebuild.
     const ctSpendAddresses = hasShieldedKeys ? await getWalletCtSpendAddresses(mysql, walletId) : [];
     if (hasShieldedKeys) {
-      await findAndRewindShielded(mysql, walletId, logger);
+      const firstSweep = await findAndRewindShielded(mysql, walletId, logger);
       // Settle drain: a daemon ingest whose ownership check snapshotted the
       // world before our claim committed lands its output unowned moments
       // later — one more (cheap when empty) sweep closes that window.
-      await findAndRewindShielded(mysql, walletId, logger);
+      const settleSweep = await findAndRewindShielded(mysql, walletId, logger);
+      // A rewind that fails is recorded as `recovery_failed` and re-driven by a
+      // later catch-up, so the load still completes — but the balance is
+      // incomplete until then, so make the count greppable next to the
+      // per-output alerts rather than letting it pass silently.
+      const failedRewinds = firstSweep.failed + settleSweep.failed;
+      if (failedRewinds > 0) {
+        logger.error('Shielded outputs failed to recover during load', { walletId, failed: failedRewinds });
+      }
       await rebuildShieldedAddressBalances(mysql, ctSpendAddresses);
       await rebuildShieldedAddressTxHistory(mysql, ctSpendAddresses);
     }
@@ -817,23 +839,37 @@ export const loadWallet: Handler<LoadEvent, LoadResult> = async (event) => {
       await updateWalletStatus(mysql, walletId, WalletStatus.READY);
     }
 
-    await closeDbConnection(mysql);
     return { success: true, walletId, xpubkey };
   } catch (e) {
-    logger.error('Erroed on loadWallet: ', e);
-    await addAlert(
-      'Wallet load failed',
-      `The legacy+shielded load for wallet ${walletId} failed and was marked for retry.`,
-      Severity.MINOR,
-      { wallet_id: walletId, error: String(e), source: 'wallet-service' },
-      logger,
-    );
-    if (hasShieldedKeys) {
-      await markWalletLoadError(mysql, walletId, !legacyWasReady);
-    } else {
-      await updateWalletStatus(mysql, walletId, WalletStatus.ERROR, (wallet.retryCount ?? 0) + 1);
+    logger.error('Errored on loadWallet: ', e);
+    // The dominant failure here is a dead connection/pool, which would make the
+    // recovery writes below throw too. Contain that: a secondary failure must not
+    // escape and flip this Lambda from "don't retry" to the crash/DLQ channel,
+    // where the same degraded pool is waiting.
+    try {
+      await addAlert(
+        'Wallet load failed',
+        `The legacy+shielded load for wallet ${walletId} failed and was marked for retry.`,
+        Severity.MINOR,
+        { wallet_id: walletId, error: errorDetail(e), source: 'wallet-service' },
+        logger,
+      );
+      if (hasShieldedKeys) {
+        await markWalletLoadError(mysql, walletId, !legacyWasReady);
+      } else {
+        await markLegacyLoadError(mysql, walletId);
+      }
+    } catch (recoveryError) {
+      logger.error('Failed to record the load failure', { walletId, originalError: e, recoveryError });
     }
-    await closeDbConnection(mysql);
     return { success: false, walletId, xpubkey };
+  } finally {
+    // Covers both paths — a throw in the finalize writes must not leak the
+    // pooled connection in a long-lived Lambda.
+    try {
+      await closeDbConnection(mysql);
+    } catch (closeError) {
+      logger.error('Failed to close the db connection', { walletId, closeError });
+    }
   }
 };
--- a/packages/wallet-service/src/api/wallet.ts
+++ b/packages/wallet-service/src/api/wallet.ts
@@ -23,6 +23,8 @@ import {
   updateWalletAuthXpub,
   registerWalletShieldedKeys,
   casWalletErrorToCreating,
+  pinTransparentLoadFailed,
+  pinShieldedLoadFailed,
   upsertNewAddresses,
   advanceLastUsedShieldedIndex,
   markWalletCombinedReady,
@@ -550,18 +552,40 @@ export const load: APIGatewayProxyHandler = middy(async (event) => {
     wallet = await getWallet(mysql, walletId);
   };
 
+  /** 409 for a submitted key set that disagrees with the one already stored. */
+  const rejectKeysConflict = async () => {
+    await closeDbConnection(mysql);
+    return {
+      statusCode: 409,
+      body: JSON.stringify({ success: false, error: ApiError.SHIELDED_KEYS_CONFLICT }),
+    };
+  };
+
   if (hasShielded) {
+    // Whether this request is the one attaching keys to a transparent wallet, and
+    // whether it won that race — only the winner may drive the load.
+    let upgrading = false;
+    let attachedKeys = false;
+
     // Reconcile keys for an existing wallet: reject a conflicting set; attach keys
     // to a transparent wallet being upgraded. A fresh wallet already has its keys.
     if (walletExisted) {
       if (wallet.scanXpriv == null) {
-        await registerWalletShieldedKeys(mysql, walletId, value.scanXpriv, value.spendXpub, config.shieldedMaxAddressGap);
+        upgrading = true;
+        attachedKeys = await registerWalletShieldedKeys(
+          mysql, walletId, value.scanXpriv, value.spendXpub, config.shieldedMaxAddressGap,
+        );
+        if (!attachedKeys) {
+          // A concurrent request attached keys between our read and this write.
+          // Re-read and hold a differing set to the same conflict rule a
+          // sequential submit would have hit.
+          wallet = await getWallet(mysql, walletId);
+          if (wallet.scanXpriv !== value.scanXpriv || wallet.spendXpub !== value.spendXpub) {
+            return rejectKeysConflict();
+          }
+        }
       } else if (wallet.scanXpriv !== value.scanXpriv || wallet.spendXpub !== value.spendXpub) {
-        await closeDbConnection(mysql);
-        return {
-          statusCode: 409,
-          body: JSON.stringify({ success: false, error: ApiError.SHIELDED_KEYS_CONFLICT }),
-        };
+        return rejectKeysConflict();
       }
     }
 
@@ -582,19 +606,27 @@ export const load: APIGatewayProxyHandler = middy(async (event) => {
       && wallet.spendXpub === value.spendXpub
       && computeUnifiedStatus(wallet.status, wallet.ctStatus ?? 'none') !== WalletStatus.ERROR;
 
-    if (!settledWithSameKeys) {
+    let shouldInvoke: boolean;
+    if (upgrading) {
+      // Only the request that actually attached the keys drives the upgrade load;
+      // the loser of that race would otherwise spawn a duplicate.
+      shouldInvoke = attachedKeys;
+    } else if (settledWithSameKeys) {
+      shouldInvoke = false;
+    } else {
       // A retry of an errored load must atomically transition back to creating
       // first — a lost race means another request already spawned the load, so
       // this one must not double-invoke.
       const entryUnified = computeUnifiedStatus(wallet.status, wallet.ctStatus ?? 'none');
-      const shouldInvoke = entryUnified !== WalletStatus.ERROR
+      shouldInvoke = entryUnified !== WalletStatus.ERROR
         || await casWalletErrorToCreating(mysql, walletId);
-      if (shouldInvoke) {
-        try {
-          await invokeLoadWalletAsync(xpubkeyStr, maxGap);
-        } catch (e) {
-          await onAsyncInvokeError(e);
-        }
+    }
+
+    if (shouldInvoke) {
+      try {
+        await invokeLoadWalletAsync(xpubkeyStr, maxGap);
+      } catch (e) {
+        await onAsyncInvokeError(e);
       }
     }
     wallet = await getWallet(mysql, walletId);
@@ -673,17 +705,12 @@ export const loadWalletFailed: Handler<SNSEvent> = async (event) => {
       // working transparent state untouched.
       const failedWallet = await getWallet(mysql, walletId);
       if (failedWallet && failedWallet.scanXpriv != null) {
-        if (failedWallet.status === WalletStatus.READY) {
-          await mysql.query(
-            'UPDATE `wallet` SET `ct_status` = ?, `retry_count` = ? WHERE `id` = ?',
-            [WalletStatus.ERROR, MAX_LOAD_WALLET_RETRIES, walletId],
-          );
-        } else {
-          await updateWalletStatus(mysql, walletId, WalletStatus.ERROR, MAX_LOAD_WALLET_RETRIES);
-          await mysql.query('UPDATE `wallet` SET `ct_status` = ? WHERE `id` = ?', [WalletStatus.ERROR, walletId]);
+        if (failedWallet.status !== WalletStatus.READY) {
+          await pinTransparentLoadFailed(mysql, walletId, MAX_LOAD_WALLET_RETRIES);
         }
+        await pinShieldedLoadFailed(mysql, walletId, MAX_LOAD_WALLET_RETRIES);
       } else {
-        await updateWalletStatus(mysql, walletId, WalletStatus.ERROR, MAX_LOAD_WALLET_RETRIES);
+        await pinTransparentLoadFailed(mysql, walletId, MAX_LOAD_WALLET_RETRIES);
       }
 
       logger.error(`${walletId} failed to load.`);
--- a/packages/wallet-service/src/api/wallet.ts
+++ b/packages/wallet-service/src/api/wallet.ts
@@ -642,9 +642,24 @@ export const loadWalletFailed: Handler<SNSEvent> = async (event) => {
 
       const walletId = getWalletId(loadEvent.xpubkey);
 
-      // update wallet status to 'error' and set the number of retries to MAX so
-      // it doesn't get retried
-      await updateWalletStatus(mysql, walletId, WalletStatus.ERROR, MAX_LOAD_WALLET_RETRIES);
+      // Pin the wallet at the retry cap so it is not retried. A wallet with
+      // shielded keys pins its shielded lifecycle too, and an upgrade — whose
+      // transparent side was already ready before the crashed load — keeps its
+      // working transparent state untouched.
+      const failedWallet = await getWallet(mysql, walletId);
+      if (failedWallet && failedWallet.scanXpriv != null) {
+        if (failedWallet.status === WalletStatus.READY) {
+          await mysql.query(
+            'UPDATE `wallet` SET `ct_status` = ?, `retry_count` = ? WHERE `id` = ?',
+            [WalletStatus.ERROR, MAX_LOAD_WALLET_RETRIES, walletId],
+          );
+        } else {
+          await updateWalletStatus(mysql, walletId, WalletStatus.ERROR, MAX_LOAD_WALLET_RETRIES);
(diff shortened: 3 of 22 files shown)


packages/daemon/src/types/db.ts
```diff
@@ -149,8 +149,10 @@ export interface TokenSymbolsRow extends RowDataPacket {
 }
 
 export interface MaxAddressIndexRow extends RowDataPacket {
-  max_among_addresses: number,
-  max_wallet_index: number
+  max_transparent_among_addresses: number | null,
+  max_transparent_wallet_index: number | null,
```
pedroferreira1: I'm not sure I like the name "transparent" here. The new addresses will be used for transparent transactions as well, so I believe the best fit is legacy. What do you think?

packages/wallet-service/src/db/index.ts
```diff
@@ -382,22 +382,176 @@ export const updateWalletAuthXpub = async (
...
+  await mysql.query(
+    `UPDATE \`wallet\`
+        SET \`ct_status\` = ?,
+            \`retry_count\` = ?
+      WHERE \`id\` = ? AND \`ct_status\` IN (?, ?)`,
+    [WalletStatus.ERROR, retryCount, walletId, WalletStatus.CREATING, WalletStatus.ERROR],
+  );
+};
+
+export const casWalletErrorToCreating = async (
```
tuliomir, line 598: chore(non-blocking): Add jsdocs explaining what this `cas` prefix mean

packages/daemon/src/services/index.ts
```diff
@@ -668,22 +668,26 @@ export const handleVertexAccepted = async (context: Context, _event: Event) => {
...
 
-            const { maxAmongAddresses, maxWalletIndex } = indices;
+            // Transparent gap extension reads the transparent pair only — a
+            // claimed shielded (CTSpend) index must never drive or suppress
+            // transparent derivation. The CT pair feeds the shielded gap
+            // extension (follow-up work).
+            const { maxTransparentAmongAddresses, maxTransparentWalletIndex } = indices;
 
-            if (maxAmongAddresses == null || maxWalletIndex == null) {
+            if (maxTransparentAmongAddresses == null || maxTransparentWalletIndex == null) {
```
tuliomir: issue: daemon logs a false `[ERROR]` on every shielded-only tx to a READY wallet

The gap-extension guard was field-renamed but its logic was not adapted to the partition it now reads:

```ts
const { maxTransparentAmongAddresses, maxTransparentWalletIndex } = indices;
if (maxTransparentAmongAddresses == null || maxTransparentWalletIndex == null) {
  if (walletDetails.status === WalletStatus.READY) {
    logger.error('[ERROR] A wallet marked as READY does not have a max wallet index ...');
  }
  continue;
}
```

`handleVertexAccepted` funnels shielded-only vertices through this loop. For a READY combined wallet touched by a tx via **only** its CTSpend addresses (the normal shielded receive/spend case), `maxTransparentAmongAddresses` is legitimately `null` while `maxTransparentWalletIndex` is non-null (a READY wallet always has transparent rows). The `||` fires the READY branch and logs a corruption-implying ERROR on the daemon hot path. Pre-split, a CTSpend index satisfied the merged `maxAmongAddresses`, so this did not fire.

The two null cases are not equivalent: only `maxTransparentWalletIndex == null` is the genuine "READY wallet has no transparent index" invariant violation. A null `maxTransparentAmongAddresses` with a non-null wallet index just means "this tx had no transparent involvement" and should `continue` silently.

```ts
if (maxTransparentWalletIndex == null) {
  if (walletDetails.status === WalletStatus.READY) {
    logger.error('[ERROR] A wallet marked as READY has no transparent max index');
  }
  continue;
}
if (maxTransparentAmongAddresses == null) {
  // Shielded-only tx — no transparent gap work to do. Not anomalous.
  continue;
}
```

packages/wallet-service/src/api/wallet.ts
```diff
@@ -750,21 +817,108 @@ export const loadWallet: Handler<LoadEvent, LoadResult> = async (event) => {
...
+    //    from the database so previously-claimed rows stay covered on re-loads.
+    //    The rewind drains run BEFORE the rebuilds (which recompute absolutes
+    //    from tx_output), so everything revealed lands in a single rebuild.
+    const ctSpendAddresses = hasShieldedKeys ? await getWalletCtSpendAddresses(mysql, walletId) : [];
+    if (hasShieldedKeys) {
+      await findAndRewindShielded(mysql, walletId, logger);
+      // Settle drain: a daemon ingest whose ownership check snapshotted the
+      // world before our claim committed lands its output unowned moments
+      // later — one more (cheap when empty) sweep closes that window.
+      await findAndRewindShielded(mysql, walletId, logger);
```
tuliomir: issue: a wallet is flipped `ready` even when shielded outputs failed to recover

`findAndRewindShielded` returns `{ recovered, failed }` and, by design, never throws — a failed rewind is marked `recovery_failed`, an alert is emitted, and the batch continues. Both calls discard the return value, so when `failed > 0` control falls straight through to `markShieldedCatchupDone` + `markWalletCombinedReady`. The user is shown a fully `ready` wallet whose balance silently omits every `recovery_failed` output; the only trace is the on-call alert.

Capture and sum `failed` from both sweeps. If `failed > 0`, do not flip plainly `ready` — route to the error/degraded lifecycle so the retry path re-drives those rows, and `logger.error` the failed count with the walletId so it is greppable next to the per-output alerts.

packages/wallet-service/src/api/wallet.ts
```diff
@@ -750,21 +817,108 @@ export const loadWallet: Handler<LoadEvent, LoadResult> = async (event) => {
...
+      await markShieldedCatchupDone(mysql, walletId, highestDerivedIndex);
+      await markWalletCombinedReady(mysql, walletId, !transparentWasReady);
+    } else {
+      // Defensive transparent-only path: no shielded lifecycle is fabricated.
+      await updateWalletStatus(mysql, walletId, WalletStatus.READY);
+    }
+
+    await closeDbConnection(mysql);
+    return { success: true, walletId, xpubkey };
+  } catch (e) {
```
tuliomir, line 871: issue: a catch-path DB error masks the root cause, leaks the connection, and reroutes to the crash channel

The dominant failure mode for this DB-heavy worker is a connection/pool error. When `e` is one, the recovery writes in the catch (`markWalletCombinedError` at :917 / `updateWalletStatus` at :919) run on the same broken connection and throw again.

That second throw escapes the catch: `closeDbConnection` (:921) never runs (pooled-connection leak in a long-lived Lambda) and the handler *rejects* instead of returning `{ success: false }` — flipping AWS from "don't retry" to the crash/DLQ channel, where `loadWalletFailed` immediately hits the same degraded pool.

The wallet's error state may never be recorded and the true root cause is masked. The success path has the same gap (:899/:902/:905 can throw with no cleanup).

Wrap the recovery writes in their own `try/catch` that logs the secondary failure with the original `e`, and move `closeDbConnection` into a `finally` covering both paths.

packages/wallet-service/src/api/wallet.ts
```diff
@@ -528,18 +547,40 @@ export const load: APIGatewayProxyHandler = middy(async (event) => {
...
+
     // Reconcile keys for an existing wallet: reject a conflicting set; attach keys
     // to a transparent wallet being upgraded. A fresh wallet already has its keys.
     if (walletExisted) {
       if (wallet.scanXpriv == null) {
-        await registerWalletShieldedKeys(mysql, walletId, value.scanXpriv, value.spendXpub, config.shieldedMaxAddressGap);
+        upgrading = true;
+        attachedKeys = await registerWalletShieldedKeys(
+          mysql, walletId, value.scanXpriv, value.spendXpub, config.shieldedMaxAddressGap,
+        );
```
tuliomir, line 577: issue(non-blocking): upgrading a bricked (max-retries) transparent wallet strands the attached keys

`registerWalletShieldedKeys` (which sets `ct_status='creating'`) runs *before* the max-retries rejection.

If the existing transparent wallet is `status='error'` with `retryCount >= MAX`, the keys are attached and `ct_status='creating'`, then the request returns `rejectMaxRetries()` without ever invoking the load. Every resubmit re-hits the reject (unified status is ERROR because transparent is ERROR), leaving the wallet stranded with keys attached, `ct_status='creating'`, and no load running.

Narrow, but the write-before-check ordering is the smell — run the retry-cap check before attaching keys.

packages/wallet-service/src/api/wallet.ts
```diff
@@ -750,21 +817,108 @@ export const loadWallet: Handler<LoadEvent, LoadResult> = async (event) => {
...
+      'Combined wallet load failed',
+      `The combined transparent+shielded load for wallet ${walletId} failed and was marked for retry.`,
+      Severity.MINOR,
+      { wallet_id: walletId, error: String(e), source: 'wallet-service' },
+      logger,
+    );
+    if (hasShieldedKeys) {
+      await markWalletCombinedError(mysql, walletId, !transparentWasReady);
+    } else {
+      await updateWalletStatus(mysql, walletId, WalletStatus.ERROR, (wallet.retryCount ?? 0) + 1);
```
tuliomir: suggestion: the transparent-only failure branch does a non-atomic, stale retry bump

This line computes the new count from the `wallet` snapshot read at handler entry (:839), before the whole load ran. Its shielded sibling `markWalletCombinedError` already does the atomic `retry_count = retry_count + 1` in SQL. Under concurrent execution this read-modify-write can lose an increment and defeat the retry cap. Even as a defensive branch, make it use the same atomic form.

packages/wallet-service/src/api/wallet.ts
tuliomir: suggestion(test): the settle-drain second `findAndRewindShielded` is hollow coverage

Both sweeps run unconditionally after commit, and every combined test primes recoverable outputs *before* the run, so the first sweep drains everything.

**Deleting the entire second `findAndRewindShielded(...)` call leaves the whole suite green** — the exact race it exists to close (a daemon ingest that snapshotted ownership before the claim committed and writes its `unowned` output moments later) is never simulated.

Add a test that makes an output recoverable only *between* the two sweeps (or at minimum asserts the function is invoked twice), so the drain's purpose is actually pinned.

packages/wallet-service/src/api/wallet.ts
```diff
@@ -750,21 +817,108 @@ export const loadWallet: Handler<LoadEvent, LoadResult> = async (event) => {
...
+    await beginTransaction(mysql);
+    try {
+      await upsertNewAddresses(mysql, walletId, transparent.newAddresses, transparent.lastUsedAddressIndex);
+      await updateExistingAddresses(mysql, walletId, transparent.existingAddresses);
+      await upsertShieldedAddressOwnership(mysql, walletId, shielded.newRows);
+      if (shielded.lastUsedShieldedIndex != null) {
+        await advanceLastUsedShieldedIndex(mysql, walletId, shielded.lastUsedShieldedIndex);
+      }
+      await commitTransaction(mysql);
+    } catch (txError) {
```
tuliomir, line 817: suggestion(test): the claim-transaction rollback path is never executed

Both failure tests throw inside `generateShieldedAddresses`, which runs *before* `beginTransaction`, so the `catch (txError) { rollbackTransaction; throw }` branch around the upserts is never taken.

A rollback that leaves the connection dirty, or a partial claim (transparent committed, shielded not), would go unnoticed. Force a failure on `upsertShieldedAddressOwnership` (spy that throws after the transparent upsert) to cover it.

packages/wallet-service/src/db/index.ts
```diff
@@ -382,22 +382,176 @@ export const updateWalletAuthXpub = async (
...
+export const markWalletCombinedReady = async (
+  mysql: ServerlessMysql,
+  walletId: string,
+  alsoTransparent: boolean,
+): Promise<void> => {
+  if (alsoTransparent) {
+    await mysql.query(
+      'UPDATE `wallet` SET `status` = ?, `ready_at` = ?, `retry_count` = 0, `ct_status` = ? WHERE `id` = ?',
+      [WalletStatus.READY, getUnixTimestamp(), WalletStatus.READY, walletId],
+    );
```
tuliomir, line 471: suggestion(test): `ready_at` write is unasserted

Delete `ready_at = ?` from the fresh-load branch and every test still passes.

The e2e test asserts `status`/`ctStatus === READY` but not `ready_at`; the db-level test only checks `ready_at` on the *false* (upgrade) variant, where it must stay untouched. Nothing asserts the true variant actually *sets* `ready_at`.

Assert `ready_at` is populated on the fresh-load path — downstream status/response logic reads it.

packages/wallet-service/src/api/wallet.ts
```diff
@@ -637,9 +694,19 @@ export const loadWalletFailed: Handler<SNSEvent> = async (event) => {
...
       const walletId = getWalletId(loadEvent.xpubkey);
 
-      // update wallet status to 'error' and set the number of retries to MAX so
-      // it doesn't get retried
-      await updateWalletStatus(mysql, walletId, WalletStatus.ERROR, MAX_LOAD_WALLET_RETRIES);
+      // Pin the wallet at the retry cap so it is not retried. A wallet with
+      // shielded keys pins its shielded lifecycle too, and an upgrade — whose
+      // transparent side was already ready before the crashed load — keeps its
+      // working transparent state untouched.
+      const failedWallet = await getWallet(mysql, walletId);
```
tuliomir, line 719: thought: a stale DLQ event can momentarily pin a wallet that is actively re-loading

The pin guards protect an already-`ready` wallet (the tested case) but not one mid-retry in `creating`.

Sequence:
- attempt 1 crashes and queues a DLQ event; the user retries, `casWalletErrorToCreating` flips to `creating`
- attempt 2 starts; the attempt-1 DLQ then lands, matches `creating`, and pins `error` + `retry_count = MAX`.

It self-heals if attempt 2 succeeds, so the brick only sticks if attempt 2 hasn't finished or also fails.

Hard to fully close in the retry model — worth a comment acknowledging the `creating`-collision window rather than implying the guard covers every "already progressed" state.

packages/wallet-service/src/api/wallet.ts
```diff
@@ -750,21 +817,108 @@ export const loadWallet: Handler<LoadEvent, LoadResult> = async (event) => {
...
+      // world before our claim committed lands its output unowned moments
+      // later — one more (cheap when empty) sweep closes that window.
+      await findAndRewindShielded(mysql, walletId, logger);
+      await rebuildShieldedAddressBalances(mysql, ctSpendAddresses);
+      await rebuildShieldedAddressTxHistory(mysql, ctSpendAddresses);
+    }
+    await rebuildWalletBalance(mysql, walletId, [...transparent.addresses, ...ctSpendAddresses]);
+    await rebuildWalletTxHistory(mysql, walletId, [...transparent.addresses, ...ctSpendAddresses]);
+
+    if (hasShieldedKeys) {
```
tuliomir: thought: broad `catch (e)` conflates deterministic bugs with transient failures

Every error — crypto, DB, or plain bug — becomes one MINOR "marked for retry" alert, so a deterministic bug burns all `MAX_LOAD_WALLET_RETRIES` re-hitting itself while the user is told to retry.

Concretely, `shielded.rows[shielded.rows.length - 1].index` at :897 throws a generic `TypeError` if `shielded.rows` is empty, which only happens when `shieldedMaxGap <= 0` — practically unreachable with a valid config, so this is a defensive-guard suggestion, not a live bug.

Guard it with an explicit `throw new Error('no shielded rows derived …')` so an invariant violation reads as one, not as a retryable blip.

packages/wallet-service/src/db/index.ts
```diff
@@ -382,22 +382,176 @@ export const updateWalletAuthXpub = async (
...
+  }
+};
+
+/**
+ * Atomically transition an errored wallet back to creating before re-invoking
+ * the async load. Returns false when there was nothing to flip — the caller
+ * lost the race to another request and must not spawn a duplicate load.
+ */
+/**
```
tuliomir, line 529: nitpick: two stacked/stranded docstrings document the wrong symbol

`registerWalletShieldedKeys` has two `/** */` blocks stacked above it (378-384 and 385-392); the first is an orphaned param block.

Separately, the docstring at 501-505 describes `casWalletErrorToCreating` but sits directly above `pinTransparentLoadFailed`'s own docstring and function. The real `casWalletErrorToCreating` at :543 is left undocumented.

Collapse each to exactly one docstring directly above its function.

packages/wallet-service/src/api/wallet.ts
```diff
@@ -750,21 +817,108 @@ export const loadWallet: Handler<LoadEvent, LoadResult> = async (event) => {
...
+
+    await closeDbConnection(mysql);
+    return { success: true, walletId, xpubkey };
+  } catch (e) {
+    logger.error('Erroed on loadWalletCombined: ', e);
+    await addAlert(
+      'Combined wallet load failed',
+      `The combined transparent+shielded load for wallet ${walletId} failed and was marked for retry.`,
+      Severity.MINOR,
+      { wallet_id: walletId, error: String(e), source: 'wallet-service' },
```
tuliomir: nitpick: failure-path log quality

`packages/wallet-service/src/api/wallet.ts:908, 913`

`logger.error('Erroed on loadWalletCombined: ', e)` — "Erroed" typo, copied from the deprecated `loadWallet` at :803.

The alert metadata uses `error: String(e)` (:913), which drops the stack for a real `Error`; the stack survives only in the log line, with no shared identifier to join the two.

Use a stable identifier in both and log `e`, not `String(e)`. Consider a small `extractMessageFromError()` method for this purpose.

packages/wallet-service/tests/loadWalletCombined.test.ts
```diff
@@ -0,0 +1,184 @@
...
+  await cleanDatabase(mysql);
+  resetCtCryptoMock();
+  mockedAddAlert.mockClear();
+});
+
+afterAll(async () => {
+  await closeDbConnection(mysql);
+});
+
+describe('loadWalletCombined', () => {
```
tuliomir: suggestion(test): **Transparent idempotency not asserted**

Here we check only `unlocked_shielded_balance === 700` on the second run, so a transparent-side double-credit on re-load would pass.

packages/wallet-service/tests/db.test.ts
```diff
@@ -4125,3 +4146,70 @@ describe('createWallet with shielded keys', () => {
...
+    let w = await getWallet(mysql, wid);
+    expect(w.status).toBe(WalletStatus.ERROR);
+    expect(w.ctStatus).toBe(WalletStatus.ERROR);
+    expect(w.retryCount).toBe(1);
+    await markWalletCombinedError(mysql, wid, false); // upgrade variant: transparent untouched
+    w = await getWallet(mysql, wid);
+    expect(w.retryCount).toBe(2);
+  });
+
+  it('casWalletErrorToCreating wins only from an error state', async () => {
```
tuliomir, line 4126: suggestion(test): both-error and ct-only-error are covered, but not `status='error'`/`ct_status='ready'`; the two `IF()` legs are independent.

packages/wallet-service/src/db/index.ts
```diff
@@ -382,22 +382,176 @@ export const updateWalletAuthXpub = async (
...
+ * Pin a crashed wallet's transparent side at the retry cap so it is not retried.
+ *
+ * Guarded on the wallet still being mid-load: a crash DLQ event can be delivered
+ * after a later attempt already recovered the wallet, and demoting a recovered
+ * wallet would strand it behind the max-retries rejection with no client-side fix.
+ */
+export const pinTransparentLoadFailed = async (
```
tuliomir: suggestion(test): `ready_at` never asserted in the `loadWalletFailed` API tests.

The pin *guards* themselves are well covered — the two "does not demote … already recovered" tests fail if the guards are removed.

packages/wallet-service/src/api/wallet.ts
```diff
@@ -637,9 +694,19 @@ export const loadWalletFailed: Handler<SNSEvent> = async (event) => {
...
-      await updateWalletStatus(mysql, walletId, WalletStatus.ERROR, MAX_LOAD_WALLET_RETRIES);
+      // Pin the wallet at the retry cap so it is not retried. A wallet with
+      // shielded keys pins its shielded lifecycle too, and an upgrade — whose
+      // transparent side was already ready before the crashed load — keeps its
+      // working transparent state untouched.
+      const failedWallet = await getWallet(mysql, walletId);
+      if (failedWallet && failedWallet.scanXpriv != null) {
+        if (failedWallet.status !== WalletStatus.READY) {
+          await pinTransparentLoadFailed(mysql, walletId, MAX_LOAD_WALLET_RETRIES);
+        }
```
pedroferreira1, line 723: When can the wallet be ready in the loadFailed method? For wallets that were already loaded before the shielded?

packages/daemon/src/services/index.ts
r4mmer: Yes, the current implementation still needs the gap extension from the daemon.
This is part of a group of deffered work that will be implemented after this merges.

packages/daemon/src/types/db.ts
r4mmer: Agreed — renamed to `legacy` throughout (07c6c4f). The account-2 counterparts were already `ct` (`maxCt*` / `max_ct_*`), so the index math now reads legacy-vs-ct. Balance naming (transparent vs shielded) was left as-is since that distinction is correct.

packages/wallet-service/src/api/wallet.ts
r4mmer: It's the upgrade case: a wallet whose legacy side was already `READY` before a shielded-upgrade load that then crashed. `loadWalletFailed` sees `status=READY`, so it pins only `ct_status` and leaves the working legacy side untouched.

packages/wallet-service/src/db/index.ts
r4mmer: Done (63949a0) — added a docstring on `casWalletErrorToCreating` explaining the compare-and-set semantics.

packages/wallet-service/src/api/wallet.ts
r4mmer: Good catch. Both sweeps' `failed` counts are now summed and `logger.error`'d with the walletId when `> 0` (63949a0). We kept the ready flip deliberately — `recovery_failed` rows are durable and re-driven by the next catch-up — but if you'd prefer the wallet routed to a degraded/error lifecycle when `failed > 0`, that's an easy follow-up.

packages/wallet-service/src/api/wallet.ts
r4mmer: Fixed (63949a0): the recovery writes in the catch are wrapped in their own try/catch, and `closeDbConnection` moved into a `finally` covering both paths, so a dead-pool secondary throw can't leak the connection or flip the Lambda to the crash/DLQ channel.

packages/wallet-service/src/api/wallet.ts
r4mmer: Fixed (855cbf3): attaching keys on a first-time upgrade (guarded by `scan_xpriv IS NULL`) now resets `retry_count`, so an upgrade of a bricked wallet loads instead of being stranded behind the max-retries reject. A resubmit (keys already attached) is unaffected and still capped.

packages/wallet-service/src/api/wallet.ts
r4mmer: Fixed (63949a0): added `markLegacyLoadError` with atomic `retry_count = retry_count + 1` under the same mid-load guard; the legacy failure paths use it instead of the stale read-modify-write.

packages/wallet-service/src/api/wallet.ts
r4mmer: Fixed (63949a0): added a test where an output becomes recoverable only *between* the two sweeps (a spy seeds it after the first sweep runs), so only the second sweep can recover it — deleting the settle sweep now fails the test, which also asserts the function is called twice.

packages/wallet-service/src/api/wallet.ts
r4mmer: Fixed (63949a0): a test forces `upsertShieldedAddressOwnership` to throw inside the claim transaction, exercising the rollback branch and asserting the earlier legacy upsert left no claimed rows.

packages/wallet-service/src/db/index.ts
r4mmer: Fixed (63949a0): the fresh-load `markWalletLoadReady` test now clears `ready_at` first and asserts it's stamped.

packages/wallet-service/src/api/wallet.ts
r4mmer: Acknowledged. We considered making error-state wallets always retryable, but chose to keep the max-retries cap and only reset the counter on a genuine upgrade (855cbf3). So this `creating`-collision window can still pin a legacy-only wallet mid-retry; it self-heals if the active reload succeeds. We're accepting it as a known limitation of the retry model rather than adding more coordination for now.

packages/wallet-service/src/api/wallet.ts
r4mmer: Deferred: the empty-`shielded.rows` `TypeError` needs `shieldedMaxGap <= 0`, which a valid config never produces, so we're leaving the explicit guard out. The broader deterministic-vs-transient retry classification is a larger change we'd rather not scope into this PR.

packages/wallet-service/src/db/index.ts
r4mmer: Fixed (63949a0): collapsed the two stacked docstrings above `registerWalletShieldedKeys` into one, and moved the `casWalletErrorToCreating` docstring back onto its own function (it had drifted above `pinLegacyLoadFailed`).

packages/wallet-service/src/api/wallet.ts
r4mmer: Fixed (63949a0): "Erroed" → "Errored", and an `errorDetail()` helper keeps the `Error` stack in the alert metadata instead of `String(e)`.

packages/wallet-service/tests/loadWalletCombined.test.ts
r4mmer: Fixed (63949a0): the idempotency test now also seeds a transparent receive and asserts the transparent balance isn't double-credited on re-load. (The file is now `loadWallet.test.ts` after the load-handler rename.)

packages/wallet-service/tests/db.test.ts
r4mmer: Fixed (63949a0): added a test for the `status='error'` / `ct_status='ready'` leg, since the two legs move independently.

packages/wallet-service/src/db/index.ts
r4mmer: Fixed (63949a0): the `loadWalletFailed` fresh-crash test now asserts the legacy pin stamps `ready_at`.

packages/wallet-service/src/api/wallet.ts
```diff
@@ -574,10 +616,10 @@ export const load: APIGatewayProxyHandler = middy(async (event) => {
...
-    // Transparent-only load: the deprecated path, kept until the combined load
-    // subsumes it entirely.
+    // Transparent-only load: the load worker derives the transparent path and
+    // reconstructs its balance (shielded steps are no-ops without keys).
     try {
-      await invokeDeprecatedLoad(xpubkeyStr, maxGap);
+      await invokeLoadWalletAsync(xpubkeyStr, maxGap);
     } catch (e) {
       await onAsyncInvokeError(e);
     }
```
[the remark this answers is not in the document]
r4mmer: Addressed in 855cbf3 — the legacy-only path now CASes error→creating before invoking, mirroring the shielded path, so concurrent retries don't each spawn a duplicate load.

packages/wallet-service/src/db/index.ts
```diff
@@ -382,59 +382,177 @@ export const updateWalletAuthXpub = async (
...
+      'UPDATE `wallet` SET `status` = ?, `ct_status` = ?, `retry_count` = `retry_count` + 1 WHERE `id` = ?',
+      [WalletStatus.ERROR, WalletStatus.ERROR, walletId],
+    );
+  } else {
+    await mysql.query(
+      'UPDATE `wallet` SET `ct_status` = ?, `retry_count` = `retry_count` + 1 WHERE `id` = ?',
+      [WalletStatus.ERROR, walletId],
+    );
+  }
+};
```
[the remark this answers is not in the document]
r4mmer: Fixed (63949a0): `markWalletLoadError` now guards each leg on `status`/`ct_status IN ('creating','error')` and bumps `retry_count` only on a matched write; added a stale-failure-after-recovery regression test.

packages/wallet-service/src/api/wallet.ts
```diff
@@ -679,97 +731,109 @@ export const loadWalletFailed: Handler<SNSEvent> = async (event) => {
...
+    //    from the database so previously-claimed rows stay covered on re-loads.
+    //    The rewind drains run BEFORE the rebuilds (which recompute absolutes
+    //    from tx_output), so everything revealed lands in a single rebuild.
+    const ctSpendAddresses = hasShieldedKeys ? await getWalletCtSpendAddresses(mysql, walletId) : [];
+    if (hasShieldedKeys) {
+      await findAndRewindShielded(mysql, walletId, logger);
+      // Settle drain: a daemon ingest whose ownership check snapshotted the
+      // world before our claim committed lands its output unowned moments
+      // later — one more (cheap when empty) sweep closes that window.
+      await findAndRewindShielded(mysql, walletId, logger);
```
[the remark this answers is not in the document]
r4mmer: The two-sweep is the agreed mitigation for this race; a full ingestion checkpoint/handshake is a larger design change we're tracking as a follow-up rather than scoping here. `recovery_failed` outputs are durable and re-driven, so balance self-heals in the meantime.

packages/wallet-service/src/db/shielded.ts
```diff
@@ -338,3 +340,172 @@ export const rebuildWalletTxHistory = async (
...
+    `INSERT INTO \`address\`
+       (\`address\`, \`index\`, \`wallet_id\`, \`transactions\`, \`bip32_account\`, \`scan_privkey\`, \`catchup_state\`, \`ct_address\`)
+     VALUES ?
+     ON DUPLICATE KEY UPDATE
+       \`bip32_account\` = VALUES(\`bip32_account\`),
+       \`wallet_id\` = VALUES(\`wallet_id\`),
+       \`index\` = VALUES(\`index\`),
+       \`ct_address\` = VALUES(\`ct_address\`),
+       \`scan_privkey\` = VALUES(\`scan_privkey\`),
+       \`catchup_state\` = COALESCE(\`catchup_state\`, VALUES(\`catchup_state\`))`,
```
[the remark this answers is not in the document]
r4mmer: Noted; deferring. In practice this can't reassign across wallets — `walletId = hash(xpubkey)` and distinct wallets derive distinct addresses, so a cross-wallet collision would need a hash collision; the live case is claiming an unowned or same-wallet row. Happy to add the owner-guard if you'd prefer the belt-and-suspenders check.

packages/wallet-service/tests/loadWallet.test.ts
```diff
@@ -0,0 +1,307 @@
...
+    const spy = jest.spyOn(ShieldedDb, 'upsertShieldedAddressOwnership').mockRejectedValueOnce(new Error('claim boom'));
+
+    const result = await runWorker({ xpubkey: XPUBKEY, maxGap: MAX_GAP });
+    expect(result).toMatchObject({ success: false });
+
+    // Nothing from the claim survived: no CTSpend rows were committed.
+    const ctRows = await mysql.query(
+      'SELECT COUNT(*) AS c FROM `address` WHERE `wallet_id` = ? AND `bip32_account` = ?', [walletId, Bip32Account.CTSpend],
+    );
+    expect(Number(ctRows[0].c)).toBe(0);
```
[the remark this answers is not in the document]
r4mmer: Good point — fixed (dd67909): the test now also asserts the legacy upsert left no claimed rows, so it proves the rollback rather than relying on the mocked CTSpend count.

packages/wallet-service/tests/loadWallet.test.ts
```diff
@@ -0,0 +1,307 @@
...
+    await createWallet(mysql, walletId, XPUBKEY, AUTH_XPUBKEY, MAX_GAP, { scanXpriv, spendXpub, shieldedMaxGap: SHIELDED_GAP });
+    const realRewind = ShieldedRecovery.findAndRewindShielded;
+    let calls = 0;
+    const spy = jest.spyOn(ShieldedRecovery, 'findAndRewindShielded').mockImplementation(async (...args) => {
+      calls += 1;
+      if (calls === 1) {
+        await seedShieldedOutputAt(1, 'ctxLate', 0xd4, 900n); // lands after the claim, before sweep 2
+      }
+      return realRewind(...args);
+    });
```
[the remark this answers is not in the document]
r4mmer: Fixed (dd67909): the raced output is now seeded *after* the first sweep runs, so only the settle sweep can recover it.

packages/wallet-service/src/db/shielded.ts
```diff
@@ -338,3 +340,172 @@ export const rebuildWalletTxHistory = async (
...
+      [block.map((r) => r.spendAddress)],
+    );
+    const used = new Set(results.map((r) => r.address as string));
+    for (const row of block) {
+      if (used.has(row.spendAddress) && row.index > lastUsedIndex) {
+        lastUsedIndex = row.index;
+      }
+    }
+    highestCheckedIndex += maxGap;
+  } while (lastUsedIndex + maxGap > highestCheckedIndex);
```
pedroferreira1, line 501: It seems to be that `>=` here is the correct way but it doesn't really make a huge difference, we would just be generating one address more than the gap limit.

packages/wallet-service/src/db/index.ts
```diff
@@ -509,125 +658,6 @@ export const incrementAddressSeqnum = async (mysql: ServerlessMysql, walletId: s
...
- * Initialize a wallet's transaction history.
- *
- * @remarks
- * This function adds entries to wallet_tx_history table, using data from address_tx_history.
- *
- * @param mysql - Database connection
- * @param walletId - The wallet id
- * @param addresses - The addresses that belong to this wallet
- */
-export const initWalletTxHistory = async (mysql: ServerlessMysql, walletId: string, addresses: string[]): Promise<void> => {
```
pedroferreira1, line 522: Why you don't need this and the next methods anymore?

packages/wallet-service/src/api/wallet.ts
```diff
@@ -679,97 +758,131 @@ export const loadWalletFailed: Handler<SNSEvent> = async (event) => {
...
+    //    from the database so previously-claimed rows stay covered on re-loads.
+    //    The rewind drains run BEFORE the rebuilds (which recompute absolutes
+    //    from tx_output), so everything revealed lands in a single rebuild.
+    const ctSpendAddresses = hasShieldedKeys ? await getWalletCtSpendAddresses(mysql, walletId) : [];
+    if (hasShieldedKeys) {
+      const firstSweep = await findAndRewindShielded(mysql, walletId, logger);
+      // Settle drain: a daemon ingest whose ownership check snapshotted the
+      // world before our claim committed lands its output unowned moments
+      // later — one more (cheap when empty) sweep closes that window.
+      const settleSweep = await findAndRewindShielded(mysql, walletId, logger);
```
pedroferreira1, line 832: I didn't understand both calls. Why do you need them?

packages/wallet-service/src/db/index.ts
r4mmer: Because we removed the old loadWallet where this was used.
The rebuildWalletBalance and rebuildWalletTxHistory are the ones that will seed the wallet balance and tx history.

packages/wallet-service/src/api/wallet.ts
r4mmer: If a shielded output is processed by the daemon in the milissecond gap between phases 1 and 2 of the loadWallet it will be left unowned.
The second sweep does not close this gap, because of how the db transaction creates snapshots there is a possibility that the phase 2 and both `findAndRewindShielded` finish before the daemon commits the shielded output to the db, but it is extremely unlikely.

packages/wallet-service/src/db/shielded.ts
r4mmer: Good catch to question it, but I think `>` is actually the one we want here — the loop stops exactly on the boundary rather than one block late.

The key is what `highestCheckedIndex` means: it's the **highest index actually derived**. It starts at `-1`, each pass derives `blockStart … blockStart + maxGap - 1`, then advances by `maxGap`. So after the first pass with `maxGap = 5` we've derived `0..4` and `highestCheckedIndex = 4`.

The window we need to cover is `windowEnd = lastUsedIndex + maxGap`, so the loop should keep going only while we haven't derived that far — `windowEnd > highestCheckedIndex`, stopping once `highestCheckedIndex >= windowEnd`. That's what `>` does.

Walking the no-usage case (`maxGap = 5`): derive `0..4`, `highestCheckedIndex = 4`, `windowEnd = -1 + 5 = 4`. `4 4` is false → stop, and we keep indices `0..4` — exactly `maxGap` unused addresses. With usage at index 4: we derive `0..4`, set `lastUsedIndex = 4`, `windowEnd = 9`, `9 > 4` → derive `5..9`, then `9 > 9` is false → stop. Again exactly 5 unused addresses after the last used one.

Switching to `>=` changes two things, and it's a bit more than one address:

1. **It derives a whole extra block every run.** On the boundary (`windowEnd == highestCheckedIndex`) it loops once more and derives another `maxGap` addresses, which `rows.filter(r => r.index <= windowEnd)` then discards. At our default `shieldedMaxAddressGap = 20` that's 40 `deriveCtAddress` calls instead of 20 for the same result — and that's real crypto per address, on the load path.

2. **It can change the result, not just the cost.** If usage sits just past the gap — say used index 7 with `maxGap = 5` — `>` correctly stops after `0..4` (index 7 is beyond the gap limit, same as the transparent path would treat it), while `>=` derives the second block, discovers the usage at 7, and ends up keeping 13 addresses instead of 5. That effectively stretches the gap limit, which I don't think we want as a side effect of the loop condition.

So I'd keep `>` as-is. Let me know if you're seeing a case where it under-derives though — happy to add a test pinning the boundary if it's worth making explicit.

pedroferreira1 (approved):

tuliomir (approved):

PR #477, opened 2026-07-17
feat(shielded-outputs): shielded balances, history, utxos and addresses on wallet API

r4mmer:
## Summary

Implements [design 0002](https://github.com/HathorNetwork/rfcs/blob/master/projects/wallet-service/shielded-outputs/0002-shielded-outputs-wallet-api.md): surfaces shielded (confidential-transaction) data through the existing wallet-service REST API. Storage already carries the shielded columns (prior PRs); this PR extends the **read path**, back-compat by default, plus a small write-path invariant fix (see below). Third leg of the shielded-outputs stack, now rebased onto master after #475 merged.

> **Note:** this PR also touches `packages/daemon` (one write-path change — `addNewAddresses` tags `bip32_account`), so it is not wallet-service-only.

## What changed, by endpoint

| Endpoint | Change |
|---|---|
| `GET /balances` | `?split=true` → per-token `{ transparent, shielded, total }` for unlocked/locked; default merges into today's numbers. New top-level unified `status` (`computeUnifiedStatus(status, ct_status)`). |
| `GET /txhistory` | Each entry gains `tx_kind` (`transparent`\|`shielded`\|`mixed`) + `balanceBreakdown`; `balance` is now the merged transparent+shielded sum. |
| `GET /utxos` + `GET /tx_outputs` | `?kind=transparent\|shielded` filter (honored on the single-output `?txId&index` path too); every entry gets a `kind` discriminator; shielded entries hydrated with hex satellite bytes (`commitment`, `ephemeralPubkey`, `rangeProof`, `script`, plus mode-specific `tokenData` / `assetCommitment`+`surjectionProof`), `ctAddress`, `shieldedIndex`; account-aware `addressPath` (CTSpend → `m/44'/280'/2'/0/<i>`). |
| `GET /addresses` (+`?index=N`) | `?legacy=` account selector (default `true` = today's Legacy-account behavior, byte-identical); `legacy=false` returns CTSpend rows with `ct_address`. |
| `GET /addresses/new` | `?legacy=` (default `true`) = today's Legacy unused addresses. `?legacy=false` → `{ addresses, legacyAddresses }` where `addresses` are the CTSpend (account-2) unused addresses within the shielded gap (`last_used_shielded_index` / `shielded_max_gap`) on the account-2 path. |
| `POST /addresses/check_mine` | Resolves membership across **all** accounts — a boolean map has no account-specific fields, so a wallet's own CTSpend address correctly reports as mine (no `legacy` param). |

Supporting db-layer work: `filterTxOutputs` gains a `kind` filter with one mutually-exclusive mode/recovery clause per kind (default = transparent + recovered shielded); two **secret-free** hydration helpers in `db/shielded.ts` (`getShieldedTxOutputDataByIds`, `getShieldedAddressInfoByAddresses` — neither selects `scan_privkey`); `getNewAddresses` is parameterized per account with strict per-account filtering; and guards keeping CTSpend/shielded rows out of legacy/transparent-only reads (`getWalletUnlockedUtxos`, `getExpiredTimelocksUtxos`, `getUtxos`, `hasTransactionsOnNonFirstAddress`).

**Address-account invariant:** an owned address (`wallet_id` set) always has an explicit `bip32_account`; a NULL means an unowned observation row whose account is not yet known, and is never handed out as a new address. The legacy claim paths now enforce this on insert — wallet-service `upsertNewAddresses` (INSERT + ON DUPLICATE, so a claimed NULL row is corrected) and daemon `addNewAddresses`. No migration: existing owned-but-NULL rows are corrected when the wallet next derives them.

## Behavior notes for reviewers

- **Back-compat:** every default (no query param) response is a strict superset of today's. `?split`, `?kind`, `?legacy=false` are all opt-in.
- **`GET /balances` validation** flips `convert: false → true` so the `split=true` query string coerces to boolean (also makes `token_id` coercion-tolerant).
- **`mode` / `recoveryState`** are now part of the `/utxos` + `/tx_outputs` wire contract on **every** entry (including transparent ones, where `recoveryState` is `null`).
- **Data-integrity misses** in shielded hydration (missing satellite or ownership row) **degrade + alert** (`Severity.MAJOR`) rather than 500 the whole endpoint — one invisible UTXO instead of taking down the wallet's transparent funds too.
- **`hasTransactionsOnNonFirstAddress`** now counts CTSpend activity too, so a shielded-active wallet is not treated as a fresh single-address wallet.
- No secret key material (`scan_privkey` / `scan_xpriv`) reaches any response — verified by tracing every touched response shape.
- **Deferred:** the wallet-service does not build the tx proposal to *spend* shielded / account-2 UTXOs (clients select them via `/utxos` and build the spend themselves); shielded-spend construction is a follow-up.

## Acceptance criteria

Mapped to [design 0002](https://github.com/HathorNetwork/rfcs/blob/master/projects/wallet-service/shielded-outputs/0002-shielded-outputs-wallet-api.md):

- [x] `GET /balances` default response is unchanged in shape; `balance.unlocked`/`locked` equal `transparent + shielded` merged totals.
- [x] `GET /balances?split=true` returns `balance.unlocked`/`locked` as `{ transparent, shielded, total }` with `total` computed server-side.
- [x] `GET /balances` exposes a top-level unified `status` (`creating`\|`ready`\|`error`) derived from `legacy_status` + `ct_status`; a transparent-only wallet reads the same value as before.
- [x] `GET /txhistory` keeps its existing per-entry fields; `balance` = `balance_delta + shielded_balance_delta`; adds `tx_kind` and `balanceBreakdown`, from a single `wallet_tx_history` row per `(wallet_id, tx_id, token_id)` (no UNION).
- [x] `GET /utxos` and `GET /tx_outputs` default to returning **both** transparent and shielded UTXOs; each entry carries a `kind` discriminator. `?kind` filters both the list and single-output (`?txId&index`) paths.
- [x] Shielded entries carry `ct_address`, `shielded_index`, `mode`, the concatenated `index`, `value` (recovered amount), and the satellite bytes; `token_data` for mode 1, `asset_commitment`+`surjection_proof` for mode 2. Authorities are always `0` on shielded entries.
- [x] Unrecovered shielded rows (`recovery_state <> 'recovered'`) are excluded from UTXO results; `authority > 0` excludes shielded automatically.
- [x] `GET /addresses`, `GET /addresses?index=N` accept `legacy` (default `true`); `legacy=true` byte-identical to today; `legacy=false` returns CTSpend rows each with `ct_address`.
- [x] `GET /addresses/new` accepts `legacy` (default `true`); `legacy=false` returns CTSpend unused addresses within the shielded gap under `addresses`, and the Legacy ones under `legacyAddresses`.
- [x] `POST /addresses/check_mine` resolves membership across all accounts (a wallet owns its CTSpend addresses).
- [x] An owned address always carries an explicit `bip32_account`; the legacy claim paths set it on insert.
- [x] No `/v2` namespace and no `Accept-Version` header — same paths, same request shapes.
- [x] Old clients (no query params) see a strict superset of prior responses; the wallet-service never returns blinding factors or any secret key material.

## Testing

New tests are real DB round-trips (not mocks) asserting real serialized shapes. Covering db-layer suites (`db.test.ts`, `db.shielded.test.ts`, `txOutputs.test.ts`, `types.test.ts`) pass 169/169; the daemon `db/index.test.ts` passes 56/56 (covers the `addNewAddresses` invariant); the shielded API tests (balances, txhistory, addresses, addresses/new, check_mine, utxos/tx_outputs) all pass; `check-types` clean on both packages. Every commit was individually spec+quality reviewed and the branch passed a whole-branch review. (Note: `api.test.ts` has a pre-existing cluster of env-driven flaky failures around the combined-load / firebase lambda flow, unrelated to this PR — re-run on CI.)

## Follow-ups (not blocking)

- Align the SQL guards + address predicates to parameterized enum constants (some ship literal `mode = 0` / `<> 2`) for consistency with `filterTxOutputs`.
- `tx_kind` is derived from `wallet_tx_history` **deltas**, so a net-zero shielded tx reads `transparent` — confirm the field name/semantics (and the `tx_kind` snake vs `balanceBreakdown` camel mix) with the client team before the wire shape freezes.
- `getWalletSortedValueUtxos` is dead code (no callers) — delete in a cleanup PR.
- The wallet-service `getExpiredTimelocksUtxos` / `unlockTimelockedUtxos` are dead (only tests call them); the **live** timelock-unlock path is the daemon's `getExpiredTimelocksUtxos`, which still lacks the `AND mode = 0` guard. Add the guard there (and drop the dead wallet-service copy) in a follow-up.
- Shielded timelock-unlock accounting, and server-side shielded-spend proposal construction, are intentionally deferred.

## Summary by CodeRabbit

* **New Features**
  * Added `legacy` query support to address and new-address endpoints to select Legacy vs CTSpend rows (including CTSpend-only fields).
  * Added `split` to balances for transparent vs shielded breakdowns and unified wallet status.
  * Added `kind` filtering for wallet UTXOs/tx outputs and surfaced `tx_kind` plus per-kind balance breakdown in tx history.
* **Bug Fixes**
  * Excluded unrecovered shielded outputs from spendable results and queries.
  * Ensured claimed legacy addresses persist with the correct account classification.
  * Improved tx-output behavior for incomplete shielded metadata: returns empty results for affected items and raises major alerts.

### DIFF
--- a/packages/wallet-service/src/api/addresses.ts
+++ b/packages/wallet-service/src/api/addresses.ts
@@ -22,6 +22,7 @@ import { walletIdProxyHandler } from '@src/commons';
 import middy from '@middy/core';
 import cors from '@middy/http-cors';
 import errorHandler from '@src/api/middlewares/errorHandler';
+import { Bip32Account } from '@wallet-service/common';
 
 const mysql = getDbConnection();
 
@@ -32,11 +33,13 @@ const checkMineBodySchema = Joi.object({
     .min(1)
     .max(512) // max number of addresses in a tx (256 outputs and 256 inputs)
     .required(),
+  legacy: Joi.boolean().default(true),
 });
 
 class AddressAtIndexValidator {
   static readonly bodySchema = Joi.object({
     index: Joi.number().min(0).optional(),
+    legacy: Joi.boolean().default(true),
   });
 
   static validate(payload: unknown): { value: AddressAtIndexRequest, error: ValidationError } {
@@ -87,7 +90,8 @@ export const checkMine: APIGatewayProxyHandler = middy(walletIdProxyHandler(asyn
   }
 
   const sentAddresses = value.addresses;
-  const dbWalletAddresses: AddressInfo[] = await getWalletAddresses(mysql, walletId, sentAddresses);
+  const account = value.legacy ? Bip32Account.Legacy : Bip32Account.CTSpend;
+  const dbWalletAddresses: AddressInfo[] = await getWalletAddresses(mysql, walletId, sentAddresses, account);
   const walletAddresses: Set<string> = dbWalletAddresses.reduce((acc, { address }) => acc.add(address), new Set([]));
 
   await closeDbConnection(mysql);
@@ -107,6 +111,18 @@ export const checkMine: APIGatewayProxyHandler = middy(walletIdProxyHandler(asyn
   };
 })).use(cors())
   .use(errorHandler());
+
+// The default (legacy) response shape must stay byte-identical to what the
+// API returned before the shielded columns existed, so we build entries
+// explicitly instead of returning the internal AddressInfo shape as-is.
+// `ct_address` is only surfaced on CTSpend (legacy=false) entries.
+const toAddressResponse = (address: AddressInfo, legacy: boolean): Record<string, unknown> => ({
+  address: address.address,
+  index: address.index,
+  transactions: address.transactions,
+  seqnum: address.seqnum,
+  ...(legacy ? {} : { ct_address: address.ctAddress }),
+});
 
 /**
  * Get the addresses of a wallet, allowing an index filter
@@ -138,10 +154,12 @@ export const get: APIGatewayProxyHandler = middy(
       return closeDbAndGetError(mysql, ApiError.INVALID_PAYLOAD, { details });
     }
 
+    const account = body.legacy ? Bip32Account.Legacy : Bip32Account.CTSpend;
+
     let response = null;
 
     if ('index' in body) {
-      const address: AddressInfo | null = await dbGetAddressAtIndex(mysql, walletId, body.index);
+      const address: AddressInfo | null = await dbGetAddressAtIndex(mysql, walletId, body.index, account);
 
       if (!address) {
         return closeDbAndGetError(mysql, ApiError.ADDRESS_NOT_FOUND);
@@ -151,17 +169,17 @@ export const get: APIGatewayProxyHandler = middy(
         statusCode: 200,
         body: JSON.stringify({
           success: true,
-          addresses: [address],
+          addresses: [toAddressResponse(address, body.legacy)],
         }),
       };
     } else {
       // Searching for multiple addresses
-      const addresses = await getWalletAddresses(mysql, walletId);
+      const addresses = await getWalletAddresses(mysql, walletId, null, account);
       response = {
         statusCode: 200,
         body: JSON.stringify({
           success: true,
-          addresses,
+          addresses: addresses.map((address) => toAddressResponse(address, body.legacy)),
         }),
       };
     }
--- a/packages/wallet-service/src/api/addresses.ts
+++ b/packages/wallet-service/src/api/addresses.ts
@@ -117,13 +117,26 @@ export const checkMine: APIGatewayProxyHandler = middy(walletIdProxyHandler(asyn
 // before the shielded columns existed, so we build entries explicitly instead of
 // returning the internal AddressInfo shape as-is.
 // `ct_address` is only surfaced on CTSpend (legacy=false) entries.
-const toAddressResponse = (address: AddressInfo, legacy: boolean): Record<string, unknown> => ({
-  address: address.address,
-  index: address.index,
-  transactions: address.transactions,
-  seqnum: address.seqnum,
-  ...(legacy ? {} : { ct_address: address.ctAddress }),
-});
+const toAddressResponse = (address: AddressInfo, legacy: boolean): Record<string, unknown> => {
+  // Legacy: the on-chain address is the address. CTSpend (legacy=false): the
+  // user-facing CT address is the address, and the on-chain spend address is
+  // surfaced under `spendAddress`.
+  if (legacy) {
+    return {
+      address: address.address,
+      index: address.index,
+      transactions: address.transactions,
+      seqnum: address.seqnum,
+    };
+  }
+  return {
+    address: address.ctAddress,
+    spendAddress: address.address,
+    index: address.index,
+    transactions: address.transactions,
+    seqnum: address.seqnum,
+  };
+};
 
 /**
  * Get the addresses of a wallet, allowing an index filter
--- a/packages/wallet-service/src/api/addresses.ts
+++ b/packages/wallet-service/src/api/addresses.ts
@@ -33,7 +33,6 @@ const checkMineBodySchema = Joi.object({
     .min(1)
     .max(512) // max number of addresses in a tx (256 outputs and 256 inputs)
     .required(),
-  legacy: Joi.boolean().default(true),
 });
 
 class AddressAtIndexValidator {
@@ -90,8 +89,10 @@ export const checkMine: APIGatewayProxyHandler = middy(walletIdProxyHandler(asyn
   }
 
   const sentAddresses = value.addresses;
-  const account = value.legacy ? Bip32Account.Legacy : Bip32Account.CTSpend;
-  const dbWalletAddresses: AddressInfo[] = await getWalletAddresses(mysql, walletId, sentAddresses, account);
+  // Membership is account-agnostic: an address is "mine" if the wallet owns it,
+  // whether it is a Legacy or a CTSpend address. The response is a boolean map
+  // with no account-specific fields, so there is nothing to scope by account.
+  const dbWalletAddresses: AddressInfo[] = await getWalletAddresses(mysql, walletId, sentAddresses);
   const walletAddresses: Set<string> = dbWalletAddresses.reduce((acc, { address }) => acc.add(address), new Set([]));
 
   await closeDbConnection(mysql);
--- a/packages/wallet-service/src/api/addresses.ts
+++ b/packages/wallet-service/src/api/addresses.ts
@@ -113,9 +113,9 @@ export const checkMine: APIGatewayProxyHandler = middy(walletIdProxyHandler(asyn
 })).use(cors())
   .use(errorHandler());
 
-// The default (legacy) response shape must stay byte-identical to what the
-// API returned before the shielded columns existed, so we build entries
-// explicitly instead of returning the internal AddressInfo shape as-is.
+// The default (legacy) response must keep the same field set the API returned
+// before the shielded columns existed, so we build entries explicitly instead of
+// returning the internal AddressInfo shape as-is.
 // `ct_address` is only surfaced on CTSpend (legacy=false) entries.
 const toAddressResponse = (address: AddressInfo, legacy: boolean): Record<string, unknown> => ({
   address: address.address,
--- a/packages/wallet-service/src/api/balances.ts
+++ b/packages/wallet-service/src/api/balances.ts
@@ -15,6 +15,7 @@ import {
   getWallet,
   getTokenInformation,
 } from '@src/db';
+import { computeUnifiedStatus } from '@src/db/utils';
 import {
   closeDbConnection,
   getDbConnection,
@@ -33,6 +34,8 @@ const paramsSchema = Joi.object({
   token_id: Joi.string()
     .alphanum()
     .optional(),
+  split: Joi.boolean()
+    .default(false),
 });
 
 /*
@@ -49,7 +52,7 @@ export const get: APIGatewayProxyHandler = middy(walletIdProxyHandler(async (wal
 
   const { value, error } = paramsSchema.validate(params, {
     abortEarly: false,
-    convert: false,
+    convert: true, // query-string params always arrive as strings
   });
 
   if (error) {
@@ -90,7 +93,11 @@ export const get: APIGatewayProxyHandler = middy(walletIdProxyHandler(async (wal
 
   return {
     statusCode: 200,
-    body: bigIntUtils.JSONBigInt.stringify({ success: true, balances }),
+    body: bigIntUtils.JSONBigInt.stringify({
+      success: true,
+      balances: value.split ? balances.map((balance) => balance.toSplitJSON()) : balances,
+      status: computeUnifiedStatus(status.status, status.ctStatus ?? 'none'),
+    }),
   };
 })).use(cors())
   .use(warmupMiddleware())
--- a/packages/wallet-service/src/api/txOutputs.ts
+++ b/packages/wallet-service/src/api/txOutputs.ts
@@ -26,9 +26,12 @@ import {
   ShieldedTxOutputData,
   ShieldedAddressApiInfo,
 } from '@src/db/shielded';
-import { Bip32Account, RecoveryState, ShieldedOutputMode, isShieldedMode } from '@wallet-service/common';
+import { Bip32Account, RecoveryState, Severity, ShieldedOutputMode, isShieldedMode } from '@wallet-service/common';
+import { addAlert } from '@wallet-service/common/src/utils/alerting.utils';
+import createDefaultLogger from '@src/logger';
 
 const mysql = getDbConnection();
+const logger = createDefaultLogger();
 
 const positiveBigInt = Joi.custom(value => {
   const newVal = BigInt(value);
@@ -184,7 +187,7 @@ const hydrateAndFormat = async (
     walletId,
     [...new Set(shielded.map((output) => output.address))],
   );
-  return formatTxOutputEntries(withPath, satellite, shieldedAddresses);
+  return await formatTxOutputEntries(withPath, satellite, shieldedAddresses);
 };
 
 const _getFilteredTxOutputs = async (walletId: string, filters: IFilterTxOutput) => {
@@ -310,39 +313,67 @@ export const mapTxOutputsWithPath = (walletAddresses: AddressInfo[], txOutputs: 
 
 /**
  * Attach the `kind` discriminator and, for shielded entries, the satellite
- * crypto bytes (hex) and CTSpend ownership metadata. The recovered-only
- * filtering upstream guarantees every shielded row here has a satellite row
- * and a claimed CTSpend address row — a miss is a data-integrity bug.
+ * crypto bytes (hex) and CTSpend ownership metadata.
+ *
+ * The recovered-only filtering upstream means every shielded row here is
+ * expected to have a satellite row and a claimed CTSpend address row, but a
+ * miss can still happen (e.g. a partially-written recovery). When that
+ * happens the offending entry is skipped and a MAJOR alert is raised instead
+ * of failing the whole request — one invisible UTXO is a far smaller blast
+ * radius than a 500 on the entire /utxos or /tx_outputs response, which
+ * would otherwise also hide the wallet's unrelated transparent funds.
  */
-export const formatTxOutputEntries = (
+export const formatTxOutputEntries = async (
   txOutputs: DbTxOutputWithPath[],
   satellite: Map<string, ShieldedTxOutputData>,
   shieldedAddresses: Map<string, ShieldedAddressApiInfo>,
-): Record<string, unknown>[] => txOutputs.map((txOutput) => {
-  if (!isShieldedMode(txOutput.mode ?? 0)) {
-    return { ...txOutput, kind: 'transparent' };
+): Promise<Record<string, unknown>[]> => {
+  const entries: Record<string, unknown>[] = [];
+
+  for (const txOutput of txOutputs) {
+    if (!isShieldedMode(txOutput.mode ?? 0)) {
+      entries.push({ ...txOutput, kind: 'transparent' });
+      continue;
+    }
+
+    const data = satellite.get(`${txOutput.txId}:${txOutput.index}`);
+    const addressInfo = shieldedAddresses.get(txOutput.address);
+
+    if (!data || !addressInfo) {
+      const missing = [
+        ...(!data ? ['satellite'] : []),
+        ...(!addressInfo ? ['ownership'] : []),
+      ];
+
+      await addAlert(
+        'Shielded tx output missing satellite/ownership data',
+        `tx_output ${txOutput.txId}:${txOutput.index} is missing its ${missing.join(' and ')} row and was skipped.`,
+        Severity.MAJOR,
+        { txId: txOutput.txId, index: txOutput.index, missing },
+        logger,
+      );
+      continue;
+    }
+
+    entries.push({
+      ...txOutput,
+      kind: 'shielded',
+      ctAddress: addressInfo.ctAddress,
+      shieldedIndex: addressInfo.shieldedIndex,
+      commitment: data.commitment.toString('hex'),
+      ephemeralPubkey: data.ephemeralPubkey.toString('hex'),
+      rangeProof: data.rangeProof.toString('hex'),
+      script: data.script.toString('hex'),
+      ...(txOutput.mode === ShieldedOutputMode.AmountShielded ? { tokenData: data.tokenData } : {}),
+      ...(txOutput.mode === ShieldedOutputMode.FullyShielded ? {
+        assetCommitment: data.assetCommitment.toString('hex'),
+        surjectionProof: data.surjectionProof.toString('hex'),
+      } : {}),
+    });
   }
-  const data = satellite.get(`${txOutput.txId}:${txOutput.index}`);
-  const addressInfo = shieldedAddresses.get(txOutput.address);
-  if (!data || !addressInfo) {
-    throw new Error(`Missing shielded data for tx output ${txOutput.txId}:${txOutput.index}`);
-  }
-  return {
-    ...txOutput,
-    kind: 'shielded',
-    ctAddress: addressInfo.ctAddress,
-    shieldedIndex: addressInfo.shieldedIndex,
-    commitment: data.commitment.toString('hex'),
-    ephemeralPubkey: data.ephemeralPubkey.toString('hex'),
-    rangeProof: data.rangeProof.toString('hex'),
-    script: data.script.toString('hex'),
-    ...(txOutput.mode === ShieldedOutputMode.AmountShielded ? { tokenData: data.tokenData } : {}),
-    ...(txOutput.mode === ShieldedOutputMode.FullyShielded ? {
-      assetCommitment: data.assetCommitment.toString('hex'),
-      surjectionProof: data.surjectionProof.toString('hex'),
-    } : {}),
-  };
-});
+
+  return entries;
+};
 
 /**
  * Confirm that the requested addresses belongs to the user's wallet
--- a/packages/wallet-service/src/api/txOutputs.ts
+++ b/packages/wallet-service/src/api/txOutputs.ts
@@ -20,6 +20,13 @@ import { constants, bigIntUtils, transactionUtils } from '@hathor/wallet-lib';
 import middy from '@middy/core';
 import cors from '@middy/http-cors';
 import errorHandler from '@src/api/middlewares/errorHandler';
+import {
+  getShieldedTxOutputDataByIds,
+  getShieldedAddressInfoByAddresses,
+  ShieldedTxOutputData,
+  ShieldedAddressApiInfo,
+} from '@src/db/shielded';
+import { Bip32Account, RecoveryState, ShieldedOutputMode, isShieldedMode } from '@wallet-service/common';
 
 const mysql = getDbConnection();
 
@@ -50,6 +57,7 @@ const bodySchema = Joi.object({
   skipSpent: Joi.boolean().optional().default(true),
   txId: Joi.string().optional(),
   index: Joi.number().optional().min(0),
+  kind: Joi.string().valid('transparent', 'shielded').optional(),
 }).and('txId', 'index')
   .nand('totalAmount', 'maxAmount');
 
@@ -80,6 +88,7 @@ export const getFilteredUtxos = middy(walletIdProxyHandler(async (walletId, even
     totalAmount: queryString.totalAmount,
     maxAmount: queryString.maxAmount,
     maxOutputs: queryString.maxOutputs,
+    kind: queryString.kind,
   };
 
   const { value, error } = bodySchema.validate(eventBody, {
@@ -135,6 +144,7 @@ export const getFilteredTxOutputs = middy(walletIdProxyHandler(async (walletId, 
     totalAmount: queryString.totalAmount,
     maxAmount: queryString.maxAmount,
     maxOutputs: queryString.maxOutputs,
+    kind: queryString.kind,
   };
 
   const { value, error } = bodySchema.validate(eventBody, {
@@ -155,12 +165,34 @@ export const getFilteredTxOutputs = middy(walletIdProxyHandler(async (walletId, 
 })).use(cors())
   .use(errorHandler());
 
+const hydrateAndFormat = async (
+  walletId: string,
+  walletAddresses: AddressInfo[],
+  txOutputs: DbTxOutput[],
+): Promise<Record<string, unknown>[]> => {
+  const withPath = mapTxOutputsWithPath(walletAddresses, txOutputs);
+  const shielded = withPath.filter((output) => isShieldedMode(output.mode ?? 0));
+  if (shielded.length === 0) {
+    return formatTxOutputEntries(withPath, new Map(), new Map());
+  }
+  const satellite = await getShieldedTxOutputDataByIds(
+    mysql,
+    shielded.map((output) => ({ txId: output.txId, index: output.index })),
+  );
+  const shieldedAddresses = await getShieldedAddressInfoByAddresses(
+    mysql,
+    walletId,
+    [...new Set(shielded.map((output) => output.address))],
+  );
+  return formatTxOutputEntries(withPath, satellite, shieldedAddresses);
+};
+
 const _getFilteredTxOutputs = async (walletId: string, filters: IFilterTxOutput) => {
   const walletAddresses = await getWalletAddresses(mysql, walletId);
 
   // txId will only be on the body when the user is searching for specific tx outputs
   if (filters.txId !== undefined) {
-    let txOutputList: DbTxOutputWithPath[] = [];
+    let txOutputList: Record<string, unknown>[] = [];
     const txOutput: DbTxOutput = await getTxOutput(mysql, filters.txId, filters.index, filters.skipSpent);
 
     if (txOutput) {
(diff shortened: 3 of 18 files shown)


packages/wallet-service/src/api/addresses.ts
```diff
@@ -108,6 +112,18 @@ export const checkMine: APIGatewayProxyHandler = middy(walletIdProxyHandler(asyn
...
+// API returned before the shielded columns existed, so we build entries
+// explicitly instead of returning the internal AddressInfo shape as-is.
+// `ct_address` is only surfaced on CTSpend (legacy=false) entries.
+const toAddressResponse = (address: AddressInfo, legacy: boolean): Record<string, unknown> => ({
+  address: address.address,
+  index: address.index,
+  transactions: address.transactions,
+  seqnum: address.seqnum,
+  ...(legacy ? {} : { ct_address: address.ctAddress }),
+});
```
pedroferreira1: I don't understand this. If we use `legacy: false` it should return the ct address at the `address` key, or no?

packages/wallet-service/src/api/balances.ts
```diff
@@ -49,7 +52,7 @@ export const get: APIGatewayProxyHandler = middy(walletIdProxyHandler(async (wal
 
   const { value, error } = paramsSchema.validate(params, {
     abortEarly: false,
-    convert: false,
+    convert: true, // query-string params always arrive as strings
```
pedroferreira1, line 55: Why was it `false` before? Is it currently a bug?

packages/wallet-service/src/db/index.ts
pedroferreira1: Don't we need to also return shielded addresses in this method?

packages/wallet-service/src/db/index.ts
```diff
@@ -1483,6 +1517,9 @@ INNER JOIN token ON w.token_id = token.id
     const totalAmount = BigInt(result.total_received as string);
     const unlockedBalance = BigInt(result.unlocked_balance as string);
     const lockedBalance = BigInt(result.locked_balance as string);
```
pedroferreira1: Does it make sense to use `dbIntToBigInt` here as well?

packages/wallet-service/src/db/index.ts
```diff
@@ -1570,12 +1608,19 @@ LEFT OUTER JOIN transaction ON transaction.tx_id = wallet_tx_history.tx_id
...
+    const transparentDelta = BigInt(result.balance as string);
+    const shieldedDelta = BigInt((result.shielded_balance_delta ?? 0) as string | number);
     const tx: TxTokenBalance = {
       txId: <string>result.tx_id,
       timestamp: <number>result.timestamp,
       voided: <boolean>result.voided,
-      balance: BigInt(result.balance as string),
+      balance: transparentDelta + shieldedDelta,
       version: <number>result.version,
+      output_kind: deriveOutputKind(transparentDelta, shieldedDelta),
```
pedroferreira1: Why output kind? Isn't it better `tx_kind`?

packages/wallet-service/src/db/index.ts
```diff
@@ -1659,6 +1704,7 @@ export const getWalletUnlockedUtxos = async (
         AND (\`timelock\` <= ?
              OR \`timelock\` is NULL)
         AND \`locked\` = 1
+        AND \`mode\` = 0
```
pedroferreira1, line 1757: Does it make sense to have a parameter to filter by `mode` in this method?

packages/wallet-service/src/db/index.ts
```diff
@@ -998,8 +1005,9 @@ export const getUtxos = async (
       WHERE (\`tx_id\`, \`index\`)
          IN (?)
         AND \`spent_by\` IS NULL
-        AND \`voided\` = FALSE`,
-    [entries],
+        AND \`voided\` = FALSE
+        AND (\`mode\` = ? OR \`recovery_state\` = ?)`,
```
pedroferreira1, line 1037: Does it make sense to have a parameter to filter by `mode` in this method?

packages/wallet-service/src/db/index.ts
```diff
@@ -2586,6 +2640,9 @@ export const filterTxOutputs = async (
       WHERE \`address\`
          IN (?)
         AND \`token_id\` = ?
+        AND (\`mode\` = ? OR \`recovery_state\` = ?)
+        ${finalFilters.kind === 'transparent' ? 'AND `mode` = ?' : ''}
+        ${finalFilters.kind === 'shielded' ? 'AND `mode` IN (?)' : ''}
```
pedroferreira1: I think I understand this query but it seems weird to filter twice by mode. Can we have something like

```
${finalFilters.kind === 'transparent' ? 'AND `mode` = ShieldedOutputMode.Transparent' : ''}
${finalFilters.kind === 'shielded' ? 'AND `mode` IN (ShieldedOutputMode.AmountShielded, ShieldedOutputMode.FullyShielded) AND `recovery_state` = RecoveryState.Recovered' : ''}
``` 

What do you think?

packages/wallet-service/src/db/index.ts
```diff
@@ -2737,12 +2796,15 @@ export const getExpiredTimelocksUtxos = async (
...
 ): Promise<DbTxOutput[]> => {
+  // Shielded timelock-unlock accounting is daemon-side follow-up work; this
+  // feeds the transparent-only balance rebuild, so shielded rows are excluded.
   const results: DbSelectResult = await mysql.query(`
     SELECT *
       FROM tx_output
      WHERE locked = TRUE
        AND timelock IS NOT NULL
        AND timelock < ?
+       AND mode = 0
```
pedroferreira1, line 2872: Should we have a filter by `mode`?

packages/wallet-service/src/db/index.ts
```diff
@@ -3351,6 +3425,7 @@ export const hasTransactionsOnNonFirstAddress = async (
       WHERE \`wallet_id\` = ?
         AND \`index\` > 0
         AND \`transactions\` > 0
+        AND (\`bip32_account\` IS NULL OR \`bip32_account\` <> 2)
```
pedroferreira1: Why this filter? It should return true for shielded addresses as well

packages/wallet-service/src/api/balances.ts
r4mmer: No, the new split argument would be a bug since `?split=false` would give us `split = "false"` which evaluates to true if we make `if (split) {...`. This is why we need to convert it to boolean.

packages/wallet-service/src/db/utils.ts
```diff
@@ -108,6 +109,16 @@ export const computeUnifiedStatus = (status: WalletStatus, ctStatus: CtStatus):
   return WalletStatus.READY;
 };
 
+/**
+ * Classify a history row by which balance deltas are non-zero. A row with
+ * both deltas zero keeps the transparent reading it always had.
+ */
+export const deriveTxKind = (transparentDelta: bigint, shieldedDelta: bigint): TxKind => {
+  if (shieldedDelta === 0n) return 'transparent';
+  if (transparentDelta === 0n) return 'shielded';
```
tuliomir, line 118: question: Is there a possibility to mis-categorize an authority tx as `shielded` because it has no delta?

packages/wallet-service/src/api/txOutputs.ts
tuliomir, line 254: issue: default token-selection returns shielded UTXOs a transparent spend can't consume

`GET /wallet/tx_outputs?totalAmount=X` (and `/wallet/utxos`, and `maxAmount`) with no `kind` runs `filterTxOutputs` on the default clause `mode = 0 OR recovery_state = 'recovered'` — the *merged* transparent + recovered-shielded set — then hands it straight to `transactionUtils.selectUtxos`, which treats every row as fungible-by-value.

Unlike the plain listing (where merging kinds is the intended, back-compat superset), the selection path exists to be *spent*: a returned `kind: 'shielded'` output carries a shielded script, not the P2PKH the transparent input path assumes, so a client that funds a transparent tx from this selection builds inputs it cannot spend — and gets no signal, since the response is `success: true`.

Constrain the token-selection path to transparent by default: when `totalAmount`/`maxAmount` is set and `kind` is unset, force `kind: 'transparent'` into the filters (or filter `finalTxOutputs` to `mode === 0` before `selectUtxos`/the `maxAmount` loop), leaving explicit `kind=shielded` for callers who ask for it. No `totalAmount`/`maxAmount` test currently uses a mixed fixture, so this is untested today.

packages/wallet-service/src/api/txOutputs.ts
```diff
@@ -266,10 +309,75 @@ export const mapTxOutputsWithPath = (walletAddresses: AddressInfo[], txOutputs:
...
+      ...txOutput,
+      kind: 'shielded',
+      ctAddress: addressInfo.ctAddress,
+      shieldedIndex: addressInfo.shieldedIndex,
+      commitment: data.commitment.toString('hex'),
+      ephemeralPubkey: data.ephemeralPubkey.toString('hex'),
+      rangeProof: data.rangeProof.toString('hex'),
+      script: data.script.toString('hex'),
+      ...(txOutput.mode === ShieldedOutputMode.AmountShielded ? { tokenData: data.tokenData } : {}),
+      ...(txOutput.mode === ShieldedOutputMode.FullyShielded ? {
```
tuliomir, line 388: issue(non-blocking): a FullyShielded row with a NULL asset sub-field 500s the whole response, defeating the degrade design

The degrade guard is `if (!data || !addressInfo)` — it catches a *missing* satellite/ownership row and skips+alerts. But `getShieldedTxOutputDataByIds` maps `assetCommitment`/`surjectionProof` as `Buffer | null` (shielded.ts:189), so a *present-but-partial* mode-2 row passes the guard and then hits `data.assetCommitment.toString('hex')` on `null`.

That `TypeError` is uncaught inside the loop → the entire `/utxos` and `/tx_outputs` response 500s, taking the wallet's transparent funds down with it — the exact blast radius the doc-comment at `:317` says this path prevents, reached through a different door. Missing-row is handled gracefully; missing-sub-field crashes.

Extend the guard to treat a present row with any NULL required byte-field for its mode (assetCommitment/surjectionProof for FullyShielded) the same as a missing row — route it through the same skip+alert.

packages/wallet-service/src/api/txOutputs.ts
```diff
@@ -266,10 +309,75 @@ export const mapTxOutputsWithPath = (walletAddresses: AddressInfo[], txOutputs:
...
+    const data = satellite.get(`${txOutput.txId}:${txOutput.index}`);
+    const addressInfo = shieldedAddresses.get(txOutput.address);
+
+    if (!data || !addressInfo) {
+      const missing = [
+        ...(!data ? ['satellite'] : []),
+        ...(!addressInfo ? ['ownership'] : []),
+      ];
+
+      await addAlert(
```
tuliomir, line 368: suggestion(non-blocking): an `addAlert` failure escalates a single-UTXO skip into a full 500

The degrade path `await addAlert(...)` is the only statement between a bad shielded row and a returned response. `addAlert` hits an external backend (SNS/network); if it throws, the exception propagates out of `formatTxOutputEntries` and 500s the request.

So the "one invisible UTXO beats a 500" guarantee holds only while alerting is healthy — the moment it isn't, you get both a lost alert *and* the full-response 500. Wrap the `addAlert` in its own try/catch that falls back to `logger` on failure, so alert-delivery can never escalate a per-row skip into a total failure.

packages/wallet-service/src/api/txOutputs.ts
```diff
@@ -266,10 +309,75 @@ export const mapTxOutputsWithPath = (walletAddresses: AddressInfo[], txOutputs:
...
+ * would otherwise also hide the wallet's unrelated transparent funds.
+ */
+export const formatTxOutputEntries = async (
+  txOutputs: DbTxOutputWithPath[],
+  satellite: Map<string, ShieldedTxOutputData>,
+  shieldedAddresses: Map<string, ShieldedAddressApiInfo>,
+): Promise<Record<string, unknown>[]> => {
+  const entries: Record<string, unknown>[] = [];
+
+  for (const txOutput of txOutputs) {
```
tuliomir, line 346: suggestion(non-blocking): a dropped shielded UTXO is invisible to the caller — no partial-result signal

Whenever a recovered shielded UTXO is skipped, `/utxos` and `/tx_outputs` return `success: true` with the entry simply absent and no machine-readable "this list is incomplete" flag. A client reconciling against `/balances` (which counts that output in `unlocked_shielded_balance`) sees UTXOs that don't sum to the reported balance, with no explanation; the ops alert is the only trace.

Consider a response-level `partial: true` (+ dropped count/ids) so callers can detect and react rather than treating a short list as authoritative.

packages/wallet-service/src/db/index.ts
```diff
@@ -2615,14 +2678,16 @@ export const mapDbResultToDbTxOutput = (result: any): DbTxOutput => ({
   index: result.index as number,
   tokenId: result.token_id as string,
   address: result.address as string,
-  value: BigInt(result.value),
+  value: BigInt(result.value ?? 0),
```
tuliomir: suggestion(non-blocking): `value: BigInt(result.value ?? 0)` masks a NULL money column into a spendable-looking 0

Previously `BigInt(result.value as string)` threw on a NULL value — a loud integrity signal. `?? 0` now silently coerces any NULL into `0n`. Safety now rests entirely on every caller's query excluding unrecovered (NULL-value) rows: `getUtxos`/`filterTxOutputs`/`getExpiredTimelocksUtxos` do, but `getTxOutput` does not — it's saved only by the JS `isUnrecoveredShielded` check at its single caller (txOutputs.ts:210).

The money field is the one you least want silently zeroed. Prefer keeping the mapper strict — throw with `txId:index` context when `value == null` — and let the recovered-only queries continue filtering in SQL; a NULL reaching the mapper is a real integrity violation and should stay loud. (Same pattern, lower stakes: `getWalletTxHistory`'s `shielded_balance_delta ?? 0` at db/index.ts:1612 — confirm the column is `NOT NULL DEFAULT 0`; if it can be genuinely NULL, "unknown delta" is conflated with "zero delta" and the row mis-classifies as `transparent`.)

packages/wallet-service/src/types.ts
```diff
@@ -359,13 +362,22 @@ export class Balance {
...
   lockExpires: number | null;
 
-  constructor(totalAmountSent = 0n, unlockedAmount = 0n, lockedAmount = 0n, lockExpires = null, unlockedAuthorities = null, lockedAuthorities = null) {
+  unlockedShieldedAmount: bigint;
+
+  lockedShieldedAmount: bigint;
+
+  totalShieldedReceived: bigint;
+
+  constructor(totalAmountSent = 0n, unlockedAmount = 0n, lockedAmount = 0n, lockExpires = null, unlockedAuthorities = null, lockedAuthorities = null, unlockedShieldedAmount = 0n, lockedShieldedAmount = 0n, totalShieldedReceived = 0n) {
```
tuliomir, line 373: suggestion(non-blocking): the 9-argument positional `Balance` constructor is a fragility trap

The constructor is now `(totalAmountSent, unlockedAmount, lockedAmount, lockExpires, unlockedAuthorities, lockedAuthorities, unlockedShieldedAmount, lockedShieldedAmount, totalShieldedReceived)` — three fresh trailing `bigint`s all defaulting to `0n`. `lockedShieldedAmount` and `totalShieldedReceived` are adjacent same-typed args: swapping them compiles clean and corrupts balances.

`TokenBalanceMap.fromStringMap` (types.ts:588) already constructs with only 6 args, so the three shielded fields silently default to `0n` with no compile error — fine today only because that path predates shielded data. Move to an options object (`new Balance({ unlockedShieldedAmount, ... })`) or named factories; it also makes the partial construction explicit rather than accidental.

packages/wallet-service/src/api/txOutputs.ts
```diff
@@ -266,10 +309,75 @@ export const mapTxOutputsWithPath = (walletAddresses: AddressInfo[], txOutputs:
...
+ *
+ * The recovered-only filtering upstream means every shielded row here is
+ * expected to have a satellite row and a claimed CTSpend address row, but a
+ * miss can still happen (e.g. a partially-written recovery). When that
+ * happens the offending entry is skipped and a MAJOR alert is raised instead
+ * of failing the whole request — one invisible UTXO is a far smaller blast
+ * radius than a 500 on the entire /utxos or /tx_outputs response, which
+ * would otherwise also hide the wallet's unrelated transparent funds.
+ */
+export const formatTxOutputEntries = async (
```
tuliomir, line 339: suggestion(non-blocking): `formatTxOutputEntries` returns `Record<string, unknown>[]`, so the wire contract is untyped

This is the actual `/utxos` + `/tx_outputs` response body, and it's a discriminated union in disguise: a `kind: 'transparent'` branch and a `kind: 'shielded'` branch whose extra fields split again on `mode` (tokenData for AmountShielded; assetCommitment+surjectionProof for FullyShielded). As `Record<string, unknown>`, a dropped `.toString('hex')` or a mis-spelled key is invisible to the compiler, and consumers have nothing to import — the contract lives only in this function body.

Declare `type TxOutputEntry = TransparentTxOutputEntry | ShieldedTxOutputEntry` discriminated on `kind` and return it. The annotation alone catches assembly bugs with no behavior change.

packages/wallet-service/src/types.ts
```diff
@@ -694,6 +743,8 @@ export interface DbTxOutput {
   txProposalId?: string;
   txProposalIndex?: number;
   voided?: boolean | null;
+  mode?: number;
```
tuliomir: suggestion(non-blocking): type `mode` as `ShieldedOutputMode`, not bare `number`

`ShieldedOutputMode` (common/src/shielded.ts) is already a zero-runtime-cost `0 | 1 | 2` union and is already imported into these files. Typing `DbTxOutput.mode` (and `ShieldedOutputToRecover.mode`, shielded.ts:124) as `number` throws that away: it permits `mode: 7` and turns the `mode === ShieldedOutputMode.AmountShielded` comparisons into untyped narrowings. Mechanical change, no runtime effect.

packages/wallet-service/tests/txOutputs.test.ts
tuliomir, line 481: suggestion(non-blocking): the missing-**ownership** and both-missing degrade branches are untested

The existing degrade test seeds a recovered shielded UTXO with no `shielded_tx_output_data` row, driving the `!data` half of `if (!data || !addressInfo)`. Because `!data` short-circuits, the `!addressInfo` half and the `addressInfo.ctAddress`/`shieldedIndex` dereferences (txOutputs.ts:364) are never exercised.

Add a case with a present satellite but an unresolvable ownership row, asserting the entry is skipped, transparent siblings still 200, and the alert fires with `missing: ['ownership']`; a both-missing variant pins `['satellite', 'ownership']`. This guards against a refactor that collapses the guard to `if (!data)` turning a designed single-UTXO degrade into a full 500 — the suite would currently stay green. While there, assert the alert's `missing` array exact value (the current test only checks `txId`/`index` via `objectContaining`), since `missing` is the actionable payload of a MAJOR alert.

packages/wallet-service/src/db/index.ts
```diff
@@ -1426,6 +1456,7 @@ export const getNewAddresses = async (mysql: ServerlessMysql, wallet: Wallet): P
      WHERE \`wallet_id\` = ?
        AND \`transactions\` = 0
        AND \`index\` > ?
+       AND (\`bip32_account\` IS NULL OR \`bip32_account\` <> 2)
```
tuliomir: suggestion(non-blocking): `getNewAddresses` gap-scan uses a bare `2` with no comment, unlike its siblings

Its siblings `getWalletAddresses` and `getAddressAtIndex` both got a NULL-tolerance comment and use `Bip32Account.CTSpend`; this one uses a raw `2` unexplained, and the pre-existing "empty and index bigger than last used" comment above it no longer describes the account exclusion.

Use the named constant and add a one-liner (why CTSpend rows must not consume the transparent gap). This also matches the PR's own stated follow-up about parameterizing the Task-7 SQL guards.

packages/wallet-service/src/db/utils.ts
tuliomir: nitpick(non-blocking): `computeUnifiedStatus` docstring claims an invariant the catch-all doesn't enforce

The final `return WalletStatus.READY` is an unconditional fall-through, but the docstring says "both ready → ready". Exhaustive for today's three-value enum (verified), so this is rot-risk, not a live bug: if a new `CtStatus` is ever added, the function reports READY while the doc claims that only happens when both are ready.

Reword to "otherwise (neither errored nor creating) → ready", and consider making the fall-through conservative (only READY when `ctStatus` is explicitly ready).

packages/wallet-service/src/api/balances.ts
```diff
@@ -90,7 +93,11 @@ export const get: APIGatewayProxyHandler = middy(walletIdProxyHandler(async (wal
 
   return {
     statusCode: 200,
-    body: bigIntUtils.JSONBigInt.stringify({ success: true, balances }),
+    body: bigIntUtils.JSONBigInt.stringify({
+      success: true,
+      balances: value.split ? balances.map((balance) => balance.toSplitJSON()) : balances,
+      status: computeUnifiedStatus(status.status, status.ctStatus ?? 'none'),
```
tuliomir, line 99: nitpick(non-blocking): This is redundant: `getWallet` (db/utils.ts:89) already coalesces `ct_status` to `'none'`. Harmless; drop for clarity.

packages/wallet-service/src/db/index.ts
```diff
@@ -979,6 +981,11 @@ export const getAuthorityUtxo = async (
 /**
  * Get the requested UTXOs.
  *
+ * @remarks
+ * Unrecovered shielded rows are excluded: their `value`/`token_id` are NULL until
+ * recovery, and the mapper's `?? 0` fallback would otherwise mask that into a
```
tuliomir: nitpick(non-blocking): the `getUtxos` `@remarks` says the `?? 0` fallback masks "`value`/`token_id`", but the fallback (mapDbResultToDbTxOutput:2681) is `value`-only; `token_id` has no such fallback. Drop `token_id` from the sentence.

packages/wallet-service/src/api/addresses.ts
```diff
@@ -108,6 +113,18 @@ export const checkMine: APIGatewayProxyHandler = middy(walletIdProxyHandler(asyn
 })).use(cors())
   .use(errorHandler());
 
+// The default (legacy) response shape must stay byte-identical to what the
```
tuliomir: nitpick(non-blocking): The `toAddressResponse` comment claims "byte-identical" to the pre-shielded response; the code guarantees the same field *set*, not byte/key-ordering identity. Soften to "same response shape (fields)".

tuliomir (dismissed):

packages/wallet-service/src/db/index.ts
r4mmer: Added a legacy argument so we can return the new addresses as well.

packages/wallet-service/src/db/index.ts
r4mmer: Changed

packages/wallet-service/src/db/index.ts
r4mmer: Done

packages/wallet-service/src/db/index.ts
r4mmer: The query uses argument but the argument itself is hardcoded, which means we are filtering for transparent utxos OR shielded utxos which are recovered (i.e. value is present).

packages/wallet-service/src/db/index.ts
r4mmer: Just checked the method, it is actually dead code. the getExpiredTimelocksUtxos which is used is on the daemon

packages/wallet-service/src/db/index.ts
r4mmer: Removed

packages/wallet-service/src/db/utils.ts
r4mmer: No — an authority-only tx has `shielded_balance_delta = 0`, so `deriveTxKind` returns `'transparent'` via the `shieldedDelta === 0` branch. `'shielded'` requires a non-zero shielded delta.

packages/wallet-service/src/api/txOutputs.ts
r4mmer: Fixed in a9e4877 — when `totalAmount`/`maxAmount` is set without an explicit `kind`, selection now restricts to transparent (`mode = 0`); an explicit `kind=shielded` is still honored. Added a mixed-fixture test asserting only the transparent UTXO is selected.

packages/wallet-service/src/api/txOutputs.ts
r4mmer: Fixed in 042d6df — the degrade guard now treats a present mode-2 row with a NULL `asset_commitment`/`surjection_proof` as a data-integrity miss (skip + alert), so it can't reach `.toString()` on null. The alert's `missing` array names the absent field.

packages/wallet-service/src/api/txOutputs.ts
r4mmer: Good catch — deferring to a follow-up: wrap the degrade-path `addAlert` in its own try/catch → logger fallback so alert-delivery can't escalate a per-row skip into a 500.

packages/wallet-service/src/api/txOutputs.ts
r4mmer: Deferring to a follow-up — a response-level `partial: true` (+ dropped ids) is a wire-contract addition worth doing, but out of this PR's scope.

packages/wallet-service/src/db/index.ts
r4mmer: Agreed, done in a9e4877 — `mapDbResultToDbTxOutput` now throws with `txId:index` context on a NULL `value` instead of `?? 0`, and `getTxOutput` filters unrecovered rows in SQL so the single-output path never surfaces one. Also confirmed `wallet_tx_history.shielded_balance_delta` is `NOT NULL DEFAULT 0` (migration `…100006`), so the `?? 0` there can't conflate "unknown" with "zero".

packages/wallet-service/src/types.ts
r4mmer: Valid fragility concern; deferring the constructor refactor to a follow-up since it touches every `Balance` construction site.

packages/wallet-service/src/api/txOutputs.ts
r4mmer: Done in a9e4877 — `formatTxOutputEntries` now returns `TxOutputEntry = TransparentTxOutputEntry | ShieldedTxOutputEntry` (discriminated on `kind`), declared in `types.ts`.

packages/wallet-service/src/types.ts
r4mmer: Done in a9e4877 — `DbTxOutput.mode` and `ShieldedOutputToRecover.mode` are now typed `ShieldedOutputMode`.

packages/wallet-service/tests/txOutputs.test.ts
r4mmer: Added in 042d6df — missing-ownership, both-missing, and mode-2-NULL-asset cases, each asserting the exact `missing` array; also tightened the existing satellite test to assert `['satellite']`.

packages/wallet-service/src/db/index.ts
r4mmer: Already addressed in the new-addresses rework — the legacy/CTSpend selectors use `${Bip32Account.Legacy}` / `${Bip32Account.CTSpend}` with an explanatory comment; no bare literal remains.

packages/wallet-service/src/db/utils.ts
r4mmer: Reworded in 042d6df to "otherwise (neither errored nor creating) → ready". (Note: `computeUnifiedStatus` itself is from #475, now in master, not this PR.)

packages/wallet-service/src/api/balances.ts
r4mmer: Kept — `Wallet.ctStatus` is typed optional (`ctStatus?: CtStatus`), so this is required for `check-types` even though `getWallet` populates it at runtime. Dropping it fails typecheck.

packages/wallet-service/src/db/index.ts
r4mmer: Fixed in 042d6df — the remark now attributes the `?? 0` mask to `value` only.

packages/wallet-service/src/api/addresses.ts
r4mmer: Softened in 042d6df to "same field set".

packages/wallet-service/src/db/index.ts
pedroferreira1: Should we remove it?

r4mmer: I think so, but I was planning on a big cleanup PR to find all dead code like this one.
Maybe we should just defer the removal.

packages/wallet-service/src/db/index.ts
```diff
@@ -1440,25 +1475,47 @@ export const getWalletAddresses = async (mysql: ServerlessMysql, walletId: strin
...
+       ${accountFilter}
   ORDER BY \`index\`
        ASC
-  LIMIT ?`, [wallet.walletId, wallet.lastUsedAddressIndex, wallet.maxGap]);
+  LIMIT ?`, [wallet.walletId, frontier, gap]);
 
   for (const result of results) {
     const index = result.index as number;
     const address = {
       address: result.address as string,
```
tuliomir, line 1516: question(non-blocking): `/addresses/new?legacy=false` returns CTSpend spend addresses without their `ct_address`

The CTSpend entries return `{ address, index, addressPath }` where `address` is the base spend address; the row's `ct_address` is read by `SELECT *` but dropped by this mapping. A client asking for fresh shielded receive addresses presumably wants the CT-encoded address to hand out.

This mirrors the open thread on `addresses.ts` about whether `legacy: false` should surface the ct address — whatever is decided there should apply to this endpoint too, or be an explicit "client derives it" decision noted in both places.

packages/wallet-service/src/api/txOutputs.ts
```diff
@@ -266,10 +319,82 @@ export const mapTxOutputsWithPath = (walletAddresses: AddressInfo[], txOutputs:
...
+    // NULL required field would crash serialization, so it is a data-integrity
+    // miss too and must degrade (skip + alert) rather than 500 the response.
+    if (data && txOutput.mode === ShieldedOutputMode.FullyShielded) {
+      if (data.assetCommitment === null) missing.push('asset_commitment');
+      if (data.surjectionProof === null) missing.push('surjection_proof');
+    }
+
+    if (missing.length > 0) {
+      await addAlert(
+        'Shielded tx output missing satellite/ownership data',
```
tuliomir, line 369: nitpick(non-blocking): the degrade-alert title says "missing satellite/ownership data" for asset-field misses too

The guard now also fires on a *present* mode-2 row with a NULL `asset_commitment`/`surjection_proof`, but the title still names only satellite/ownership. The `missing` metadata array is exact, so on-call can tell — just retitle to something like "Shielded tx output missing required data" so the alert name doesn't mislead triage.

tuliomir (dismissed):

pedroferreira1 (approved):

tuliomir (approved):

